PR #394, opened 2026-03-23
test: id + file/ignore — monotonic ID ordering and ignore pattern coverage

anandgupta42:
### What does this PR do?

**1. `Identifier` — `src/id/id.ts`** (8 new tests)

This module generates all monotonic IDs used across the application for sessions, messages, parts, permissions, and tools. Zero tests existed previously. New coverage includes:
- Ascending IDs are strictly increasing (lexicographic order matches creation order)
- Descending IDs are strictly decreasing
- Correct prefix for all ID types (`ses_`, `msg_`, `prt_`, `per_`, `tool_`)
- Same-millisecond uniqueness via internal counter (100 rapid-fire IDs all unique)
- `given` parameter passthrough when prefix matches, throws on mismatch
- `timestamp()` relative ordering is preserved between IDs created at different times
- `timestamp()` round-trips exactly for small timestamps within 48-bit range

**Why it matters:** ID misordering would cause messages to display out of order in the TUI and API. Duplicate IDs would cause silent data corruption. The `timestamp()` function is used in `truncation.ts` for retention cleanup — incorrect relative ordering would cause premature or missed cleanup.

**Notable finding:** `timestamp()` does not return the actual wall-clock `Date.now()` for modern timestamps because `BigInt(ts) * BigInt(0x1000)` exceeds 48 bits (the storage width). However, relative comparisons between `timestamp()` values remain correct, which is how it's used in practice. The tests document this behavior.

**2. `FileIgnore.match` — `src/file/ignore.ts`** (6 new tests, extending 1 existing)

This function controls which files are excluded from watching, scanning, and tool access. The existing test only covered `node_modules` directory matching. New coverage includes:
- All registered directory patterns beyond `node_modules` (`dist`, `build`, `.git`, `__pycache__`, `.next`, `out`, `bin`, `desktop`)
- File glob patterns (`.swp`, `.DS_Store`, `.pyc`, `.log`, `tmp/**`)
- Negative cases — normal source files (`src/index.ts`, `README.md`, `package.json`) are NOT ignored
- Whitelist override — whitelisted patterns prevent directory matches
- Extra patterns — caller-supplied patterns extend the default ignore set
- Windows-style path separators (`\\`) are handled correctly

**Why it matters:** If whitelist overrides don't work, users can't customize which directories are watched. If extra patterns silently fail, config-driven ignore rules are broken. False positives would silently exclude source files from the agent's view.

### Type of change

### Issue for this PR
N/A — proactive test coverage from automated test discovery

```
bun test test/id/id.test.ts          # 8 pass
bun test test/file/ignore.test.ts    # 7 pass (1 existing + 6 new)
```

### Checklist
- [x] I have added tests that prove my fix is effective or that my feature works

https://claude.ai/code/session_012pJZ9ApRnh2fGKMS957frr

## Summary by CodeRabbit

* **Tests**
  * Increased test coverage for file pattern matching behavior, with improvements for advanced directory patterns, glob file matching, configuration options, and cross-platform path separator handling.
  * Implemented comprehensive tests validating identifier monotonic ID generation behavior, including type-specific ID prefixes, strict ID ordering (ascending and descending), uniqueness guarantees, and accurate timestamp handling.

### DIFF
--- a/packages/opencode/test/file/ignore.test.ts
+++ b/packages/opencode/test/file/ignore.test.ts
@@ -1,4 +1,4 @@
-import { test, expect } from "bun:test"
+import { describe, test, expect } from "bun:test"
 import { FileIgnore } from "../../src/file/ignore"
 
 test("match nested and non-nested", () => {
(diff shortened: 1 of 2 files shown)


anandgupta42:
Consolidated into #403

PR #655, opened 2026-04-05
fix: ${VAR} env-var interpolation for configs (closes #635)

anandgupta42:
### What does this PR do?

Fixes #635 by adding shell/dotenv-style `${VAR}` interpolation to config parsing. Users arriving from Claude Code, VS Code, dotenv, or docker-compose naturally write `${VAR}` and previously hit silent failures — the literal string passed through to MCP servers, shadowing the forwarded parent env.

**Syntax supported** (after full PR):

| Syntax | Mode | Example |
|---|---|---|
| `${VAR}` | String-safe (JSON-escaped) | `"API_KEY": "${MY_API_KEY}"` |
| `${VAR:-default}` | String-safe with fallback | `"MODE": "${APP_MODE:-production}"` |
| `{env:VAR}` | Raw text (unchanged for backward compat) | `"count": {env:NUM}` |
| `$${VAR}` | Escape hatch → literal `${VAR}` | `"template": "$${VAR}"` |

Bare `$VAR` (without braces) is intentionally NOT interpolated — too collision-prone with prices, JSON pointers, etc.

**Commits:**
1. Initial `${VAR}` support with POSIX-identifier regex
2. Consensus-review fixes: JSON-escape values (security), `$${VAR}` escape hatch, stale TUI tip
3. `${VAR:-default}` fallback syntax (POSIX / docker-compose convention)
4. `config_env_interpolation` telemetry event to measure adoption and footguns

### Type of change

- [x] Bug fix
- [x] New feature (`${VAR:-default}`, telemetry)
- [ ] Test coverage
- [x] Documentation
- [ ] Refactoring
- [ ] Infrastructure

### Issue for this PR

Closes #635

### Research

Surveyed 10 other MCP clients + docker-compose + dotenv. Full findings in PR comments. Key takeaways:
- `${VAR}` + `${VAR:-default}` is the de facto standard (POSIX shell, docker-compose, dotenv, GitHub Actions, Terraform, Claude Code, Cursor)
- Every surveyed tool does **raw** text substitution — none JSON-escape. **This PR is the first MCP client to implement string-safe substitution**, eliminating the injection risk documented in OpenCode issue #5299.
- Docker-compose convention for escape hatch is `$${VAR}` — matched here.

### Security

[...]

PR #666, opened 2026-04-09
fix: resolve env-var references in MCP server environment config (closes #656)

### LINKED ISSUE #656: ${VAR} env-var interpolation in MCP server config still not working on v0.5.19
Follow-up to #635 (closed as resolved in v0.5.19).

### Problem

Environment variable interpolation in MCP server `environment` config is still not working on v0.5.19. The MCP server receives the literal string instead of the resolved value.

### Config (`~/.config/altimate-code/altimate-code.json`)

```json
{
  "mcp": {
    "gitlab": {
      "type": "local",
      "command": ["node", ".../gitlab-mcp-server/dist/index.js"],
      "environment": {
        "GITLAB_PERSONAL_ACCESS_TOKEN": "{env:GITLAB_PERSONAL_ACCESS_TOKEN}",
        "GITLAB_API_URL": "https://gitlab.com/api/v4"
      }
    }
  }
}
```

### Reproduction steps

1. Set `GITLAB_PERSONAL_ACCESS_TOKEN` in the shell environment (confirmed present via `echo`)
2. Configure the MCP server in `altimate-code.json` using `{env:GITLAB_PERSONAL_ACCESS_TOKEN}` in the `environment` block
3. Call any GitLab MCP tool (e.g., `gitlab_get_current_user`)
4. **Result:** `MCP error -32603: This operation requires user authentication`

### Additional findings

- The `${GITLAB_PERSONAL_ACCESS_TOKEN}` syntax was also tested — same failure
- Passing the token directly via the `userCredentials` parameter **works**, confirming the MCP server itself is functional
- `altimate-code --version` confirms **v0.5.19**
- The fix in v0.5.19 (`fix: ${VAR} env-var interpolation for configs (closes #635) (#655)`) does not appear to resolve this case

---

### Metadata

| Field | Value |
|-------|-------|
| CLI Version | 0.5.19 |
| Platform | darwin |
| Architecture | arm64 |
| OS Release | 25.3.0 |
| Category | bug |
| Working Directory | data-warehouse-etl-dbt |
| Session ID | ses_2aba8ca90ffeLGmZEZ4sNP90Ng |

VJ-yadav:
### What does this PR do?

Fixes #656 — `${VAR}` and `{env:VAR}` patterns in MCP server `environment` config blocks were passed as literal strings to child processes instead of being resolved to actual environment variable values. This caused auth failures for MCP servers expecting tokens (e.g. `gitlab-mcp-server` receiving literal `"${GITLAB_PERSONAL_ACCESS_TOKEN}"` instead of the token).

### Root Cause

PR #655 added env-var interpolation to the config parsing pipeline (`ConfigPaths.parseText → substitute()`), but two code paths bypassed it:

1. **MCP launch site** (`mcp/index.ts:512`) — `mcp.environment` values were spread directly into the child process env without resolving any remaining `${VAR}` patterns. If interpolation failed upstream (config updates via `updateGlobal()`, timing issues), the literal string **overwrote** the correct value already present in `process.env`.

2. **External MCP discovery** (`discover.ts:readJsonSafe()`) — configs from Claude Code (`.claude.json`), Cursor (`.cursor/mcp.json`), VS Code (`.vscode/mcp.json`), Copilot, and Gemini were parsed via `parseJsonc()` directly, completely skipping the `substitute()` interpolation pipeline.

### Changes

| File | Change |
|------|--------|
| `packages/opencode/src/mcp/index.ts` | Added `resolveEnvVars()` safety net that resolves `${VAR}`, `${VAR:-default}`, `{env:VAR}`, and `$${VAR}` (escape) patterns in `mcp.environment` values before spawning the child process |
| `packages/opencode/src/mcp/discover.ts` | Changed `readJsonSafe()` to use `ConfigPaths.parseText()` which runs `substitute()` on raw text before JSONC parsing, with graceful fallback to direct parse on failure |
| `packages/opencode/test/mcp/env-var-interpolation.test.ts` | 14 new tests |

### Supported syntaxes (all three work in MCP environment blocks)

| Syntax | Behavior | Example |
|--------|----------|---------|
| `${VAR}` | Resolves to env value, empty string if unset | `"TOKEN": "${GITLAB_TOKEN}"` |
| `${VAR:-default}` | Resolves to env value, fallback if unset | `"MODE": "${APP_MODE:-production}"` |
| `{env:VAR}` | Raw text injection (backward compat) | `"KEY": "{env:API_KEY}"` |
| `$${VAR}` | Escape — preserves literal `${VAR}` | `"TPL": "$${VAR}"` → `"${VAR}"` |

### Test plan

- [x] 11 unit tests for `resolveEnvVars`: `${VAR}`, `{env:VAR}`, defaults, escapes, unset vars, plain passthrough, multiple vars in one value, mixed entries, bare `$VAR` not matched, empty object
- [x] 3 integration tests for discovery: `.vscode/mcp.json` with `${VAR}`, `.cursor/mcp.json` with `{env:VAR}`, fallback defaults
- [x] 18 existing `discover.test.ts` tests still pass
- [x] 34 existing `paths-parsetext.test.ts` tests still pass
- [x] TypeScript typecheck clean (only pre-existing ClickHouse driver error)

### Checklist

- [x] No new dependencies
- [x] No `any` types in new code
- [x] Follows `altimate_change` marker convention
- [x] Regex pattern matches the one in `paths.ts:substitute()` for consistency
- [x] Graceful fallback: if `ConfigPaths.parseText()` fails in discovery, falls back to direct parse (no regression for malformed external configs)

Fixes #656

---
## Summary by cubic
Fixes env-var interpolation in MCP server environments so ${VAR}, ${VAR:-default}, and {env:VAR} resolve before spawn. Also interpolates external MCP configs (Cursor, VS Code, Claude Code, Copilot, Gemini). Fixes #656.

- **Bug Fixes**
  - Resolve `${VAR}`, `${VAR:-default}`, `{env:VAR}`; support `$${VAR}` escape.
  - Run resolver at launch so `mcp.environment` resolves before child spawn.
  - Interpolate external configs via `ConfigPaths.parseText()` with fallback.
  - Export `resolveEnvVars`; tests cover syntax, defaults, escapes, and external discovery.

- **Refactors**
  - Tests import the production `resolveEnvVars` and use a shared `tmpdir` fixture for cleanup.

<sup>Written for commit 5e26db6219686b89e53d755e70cf9c1a5e25b8f9. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * MCP config files and local MCP environment values now support environment-variable interpolation using ${VAR}, ${VAR:-default}, and {env:VAR}; escaped $$ sequences are preserved. External MCP JSON configs are attempted to be interpolated first, with a fallback to the previous JSONC parsing behavior if interpolation fails.

* **Tests**
  * Added unit and integration tests validating interpolation behavior and discovery of external MCP configs.

### DIFF
--- a/packages/opencode/src/mcp/discover.ts
+++ b/packages/opencode/src/mcp/discover.ts
@@ -117,6 +117,18 @@ async function readJsonSafe(filePath: string): Promise<any | undefined> {
   } catch {
     return undefined
   }
+  // altimate_change start — apply env-var interpolation to external MCP configs
+  // External configs (Claude Code, Cursor, Copilot, Gemini) may contain ${VAR}
+  // or {env:VAR} references that need resolving before JSON parsing.
+  // Uses ConfigPaths.parseText which runs substitute() then parses JSONC.
+  try {
+    const { ConfigPaths } = await import("../config/paths")
+    return await ConfigPaths.parseText(text, filePath, "empty")
+  } catch {
+    // Substitution or parse failure — fall back to direct parse without interpolation
+    log.debug("env-var interpolation failed for external MCP config, falling back to direct parse", { file: filePath })
+  }
+  // altimate_change end
   const errors: any[] = []
   const result = parseJsonc(text, errors, { allowTrailingComma: true })
   if (errors.length > 0) {
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -31,7 +31,7 @@ import { Telemetry } from "@/telemetry"
 const ENV_VAR_PATTERN =
   /\$\$(\{[A-Za-z_][A-Za-z0-9_]*(?::-[^}]*)?\})|(?<!\$)\$\{([A-Za-z_][A-Za-z0-9_]*)(?::-([^}]*))?\}|\{env:([^}]+)\}/g
 
-function resolveEnvVars(environment: Record<string, string>): Record<string, string> {
+export function resolveEnvVars(environment: Record<string, string>): Record<string, string> {
   const resolved: Record<string, string> = {}
   for (const [key, value] of Object.entries(environment)) {
     resolved[key] = value.replace(ENV_VAR_PATTERN, (match, escaped, dollarVar, dollarDefault, braceVar) => {
--- a/packages/opencode/src/mcp/index.ts
+++ b/packages/opencode/src/mcp/index.ts
@@ -25,6 +25,29 @@ import { TuiEvent } from "@/cli/cmd/tui/event"
 import open from "open"
 import { Telemetry } from "@/telemetry"
 
+// altimate_change start — resolve env-var references in MCP environment values
+// Handles ${VAR}, ${VAR:-default}, and {env:VAR} patterns that may have survived
+// config parsing (e.g. discovered external configs, config updates via updateGlobal).
+const ENV_VAR_PATTERN =
+  /\$\$(\{[A-Za-z_][A-Za-z0-9_]*(?::-[^}]*)?\})|(?<!\$)\$\{([A-Za-z_][A-Za-z0-9_]*)(?::-([^}]*))?\}|\{env:([^}]+)\}/g
+
+function resolveEnvVars(environment: Record<string, string>): Record<string, string> {
+  const resolved: Record<string, string> = {}
+  for (const [key, value] of Object.entries(environment)) {
+    resolved[key] = value.replace(ENV_VAR_PATTERN, (match, escaped, dollarVar, dollarDefault, braceVar) => {
+      if (escaped !== undefined) return "$" + escaped
+      if (dollarVar !== undefined) {
+        const envValue = process.env[dollarVar]
+        return envValue !== undefined && envValue !== "" ? envValue : (dollarDefault ?? "")
+      }
+      if (braceVar !== undefined) return process.env[braceVar] || ""
+      return match
+    })
+  }
+  return resolved
+}
+// altimate_change end
+
 export namespace MCP {
   const log = Log.create({ service: "mcp" })
   const DEFAULT_TIMEOUT = 30_000
@@ -509,7 +532,9 @@ export namespace MCP {
         env: {
           ...process.env,
           ...(cmd === "altimate" || cmd === "altimate-code" ? { BUN_BE_BUN: "1" } : {}),
-          ...mcp.environment,
+          // altimate_change start — resolve ${VAR} / {env:VAR} patterns that survived config parsing
+          ...(mcp.environment ? resolveEnvVars(mcp.environment) : {}),
+          // altimate_change end
         },
       })
       transport.stderr?.on("data", (chunk: Buffer) => {
--- a/packages/opencode/test/mcp/env-var-interpolation.test.ts
+++ b/packages/opencode/test/mcp/env-var-interpolation.test.ts
@@ -1,0 +1,229 @@
+// altimate_change start — tests for MCP env-var interpolation (closes #656)
+import { describe, test, expect, beforeEach, afterEach } from "bun:test"
+import { mkdtemp, rm, mkdir, writeFile } from "fs/promises"
+import { tmpdir } from "os"
+import path from "path"
+
+// -------------------------------------------------------------------------
+// resolveEnvVars — safety-net resolver at MCP launch site
+// -------------------------------------------------------------------------
+
+// Import the module to access resolveEnvVars indirectly via the env spread.
+// Since resolveEnvVars is a module-level function (not inside the MCP namespace),
+// we test it directly by importing the file and extracting the function.
+// For now, inline the same logic here for unit testing.
+
+const ENV_VAR_PATTERN =
+  /\$\$(\{[A-Za-z_][A-Za-z0-9_]*(?::-[^}]*)?\})|(?<!\$)\$\{([A-Za-z_][A-Za-z0-9_]*)(?::-([^}]*))?\}|\{env:([^}]+)\}/g
+
+function resolveEnvVars(environment: Record<string, string>): Record<string, string> {
+  const resolved: Record<string, string> = {}
+  for (const [key, value] of Object.entries(environment)) {
+    resolved[key] = value.replace(ENV_VAR_PATTERN, (match, escaped, dollarVar, dollarDefault, braceVar) => {
+      if (escaped !== undefined) return "$" + escaped
+      if (dollarVar !== undefined) {
+        const envValue = process.env[dollarVar]
+        return envValue !== undefined && envValue !== "" ? envValue : (dollarDefault ?? "")
+      }
+      if (braceVar !== undefined) return process.env[braceVar] || ""
+      return match
+    })
+  }
+  return resolved
+}
+
+describe("resolveEnvVars", () => {
+  const ORIGINAL_ENV = { ...process.env }
+
+  beforeEach(() => {
+    process.env["TEST_TOKEN"] = "secret-123"
+    process.env["TEST_HOST"] = "gitlab.example.com"
+  })
+
+  afterEach(() => {
+    process.env = { ...ORIGINAL_ENV }
+  })
+
+  test("resolves ${VAR} syntax", () => {
+    const result = resolveEnvVars({
+      API_TOKEN: "${TEST_TOKEN}",
+      HOST: "${TEST_HOST}",
+    })
+    expect(result.API_TOKEN).toBe("secret-123")
+    expect(result.HOST).toBe("gitlab.example.com")
+  })
+
+  test("resolves {env:VAR} syntax", () => {
+    const result = resolveEnvVars({
+      API_TOKEN: "{env:TEST_TOKEN}",
+    })
+    expect(result.API_TOKEN).toBe("secret-123")
+  })
+
+  test("resolves ${VAR:-default} with fallback when unset", () => {
+    const result = resolveEnvVars({
+      MODE: "${UNSET_VAR:-production}",
+    })
+    expect(result.MODE).toBe("production")
+  })
+
+  test("resolves ${VAR:-default} to env value when set", () => {
+    const result = resolveEnvVars({
+      TOKEN: "${TEST_TOKEN:-fallback}",
+    })
+    expect(result.TOKEN).toBe("secret-123")
+  })
+
+  test("preserves $${VAR} as literal ${VAR}", () => {
+    const result = resolveEnvVars({
+      TEMPLATE: "$${TEST_TOKEN}",
+    })
+    expect(result.TEMPLATE).toBe("${TEST_TOKEN}")
+  })
+
+  test("resolves unset variable to empty string", () => {
+    const result = resolveEnvVars({
+      MISSING: "${COMPLETELY_UNSET_VAR_XYZ}",
+    })
+    expect(result.MISSING).toBe("")
+  })
+
+  test("passes through plain values without modification", () => {
+    const result = resolveEnvVars({
+      PLAIN: "just-a-string",
+      URL: "https://gitlab.com/api/v4",
+    })
+    expect(result.PLAIN).toBe("just-a-string")
+    expect(result.URL).toBe("https://gitlab.com/api/v4")
+  })
+
+  test("resolves multiple variables in a single value", () => {
+    const result = resolveEnvVars({
+      URL: "https://${TEST_HOST}/api?token=${TEST_TOKEN}",
+    })
+    expect(result.URL).toBe("https://gitlab.example.com/api?token=secret-123")
+  })
+
+  test("handles mixed resolved and plain entries", () => {
+    const result = resolveEnvVars({
+      TOKEN: "${TEST_TOKEN}",
+      STATIC_URL: "https://gitlab.com/api/v4",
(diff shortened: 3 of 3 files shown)


VJ-yadav:
@anandgupta42 Rebased onto latest main (v0.5.20) and CI re-ran — same 7 failures, all pre-existing infra flakes unrelated to this PR:

- `worker.test.ts:67` — TUI worker type assertion
- `dbt-first-execution.test.ts` — dispatcher call failures (4 tests)
- `tracing-adversarial-2.test.ts:588` — fetch error
- `oauth-callback.ts:188` — OAuth browser test
- `ripgrep.ts:238` — file listing

None of our changed files (`mcp/index.ts`, `mcp/discover.ts`, `test/mcp/env-var-interpolation.test.ts`) are involved. Same failures showing on PR #649 and other open PRs in the batch. 7020 tests pass, only these 7 flaky ones fail.

anandgupta42:
THanks @VJ-yadav

anandgupta42:
## Multi-Model Code Review — REQUEST CHANGES

Thanks for tackling #656 — the underlying bug is real and the two-layer approach is directionally right. Reviewed by a 9-participant panel (Claude + 8 external models: GPT 5.4, Gemini 3.1 Pro, Kimi K2.5, MiniMax M2.7, GLM-5, Qwen 3.6 Plus, DeepSeek V3, MiMo-V2-Pro). Consensus verdict: **REQUEST CHANGES** — 1 Critical, 4 Major, 5 Minor, 3 NIT.

The big one is a double-interpolation regression that breaks `$${VAR}` escaping end-to-end and opens a minor variable-chain injection vector. Details below.

---

## Critical

### 1. Double interpolation — `$${VAR}` escape broken + variable-chain injection
**Location**: `packages/opencode/src/mcp/discover.ts:117-130` + `packages/opencode/src/mcp/index.ts:28-50` (call site around line 534)

`discoverExternalMcp` now runs configs through `ConfigPaths.parseText → substitute()` (Layer 1), and then at spawn time `resolveEnvVars()` runs again on the already-substituted `mcp.environment` (Layer 2). Two independent failures:

**(a) `$${VAR}` escape no longer survives end-to-end.** Layer 1 correctly turns `$${VAR}` into the literal `${VAR}`. Layer 2 then sees that literal `${VAR}` (negative lookbehind passes — there's no preceding `$` anymore) and resolves it to the live env value. After this PR there is no syntax a user can write to pass a literal `${SOMETHING}` string to an MCP server. Regression from pre-PR behavior for `opencode.json` configs.

**(b) Variable-chain injection.** If *any* env var in the shell has a value that looks like `${OTHER}`, a config referencing that first var silently expands to the second one:

```
EVIL_VAR="${SECRET}"   # in shell (common in Bash PS1, GH Actions leftovers, etc.)
config: { environment: { X: "${EVIL_VAR}" } }
Layer 1: X = "${SECRET}"        (literal, as intended)
Layer 2: X = <actual SECRET value>   ← wrong
```

Gemini independently confirmed the bug with a running POC test.

**Fix — pick one layer:**
1. **Preferred**: scope Layer 1 to `env` blocks only (apply substitution inside `addServersFromFile`/`transform`, not in `readJsonSafe` over whole JSON text) and **drop Layer 2 entirely**. This also fixes Major #5 below.
2. **Alternative**: track provenance on `Config.Mcp` entries and run `resolveEnvVars` only on those that bypassed `ConfigPaths.parseText()` (e.g., added via `updateGlobal` or at runtime).

Please gate the fix on a regression test that asserts `$${VAR}` survives to the child process as a literal `${VAR}`, plus a test for the chain-injection vector.

---

## Major

### 2. Regex/resolution logic duplicated from `ConfigPaths.substitute()` with silent divergence
**Location**: `packages/opencode/src/mcp/index.ts:29-48` vs `packages/opencode/src/config/paths.ts:104-131`

`ENV_VAR_PATTERN` is a character-for-character copy of the regex inside `substitute()`. The bodies are near-identical but diverge in three ways:

| Behavior | `substitute()` | `resolveEnvVars()` |
|---|---|---|
| Output encoding | JSON-escaped (raw JSON text) | raw (parsed strings) |
| Telemetry | emits `config_env_interpolation` counters | none |
| Unresolved tracking | counts `dollarUnresolved` etc. | silent |

The raw-vs-JSON-escape split is justified by the call sites, but there's no shared source of truth. The next regex fix will be applied to only one copy.

**Fix**: Extract a shared core in `config/paths.ts`:
```ts
export function resolveEnvVarPatterns(
  input: string,
  opts: { mode: "json-escaped" | "raw"; onUnresolved?: (name: string) => void }
): string
```
Both `substitute()` and `resolveEnvVars()` delegate to it — single regex, single resolution path.

### 3. Silent empty-string substitution on unresolved `${VAR}`
**Location**: `packages/opencode/src/mcp/index.ts:38-40`

```ts
const envValue = process.env[dollarVar]
return envValue !== undefined && envValue !== "" ? envValue : (dollarDefault ?? "")
```

An unset `${GITLAB_TOKEN}` with no default becomes `""`. The MCP server then launches with `GITLAB_TOKEN=""` and fails with an opaque auth error. No log, no warning, no telemetry. Makes diagnosing "why isn't my MCP server connecting?" needlessly painful.

**Fix**: Accept a `serverName` parameter and emit per unresolved reference:
```ts
log.warn("MCP env var unresolved, substituting empty string", { server: serverName, var: dollarVar })
```

### 4. Fallback `catch {}` in `readJsonSafe` swallows the error
**Location**: `packages/opencode/src/mcp/discover.ts:124-127`

```ts
} catch {
  log.debug("env-var interpolation failed for external MCP config, falling back to direct parse", { file: filePath })
}
```

`ConfigPaths.parseText()` can throw `JsonError`, `InvalidError`, or fail during substitution. All swallowed without the error message, and `log.debug` is invisible in normal operation.

**Fix**:
```ts
} catch (error) {
  log.warn("env-var interpolation failed for external MCP config, falling back to raw parse", {
    file: filePath,
    error: error instanceof Error ? error.message : String(error),
  })
}
```

### 5. Layer 1 substitution operates on whole JSON text, not just `env` fields
**Location**: `packages/opencode/src/mcp/discover.ts:117-130`

`ConfigPaths.parseText` runs `substitute()` over the full external-config JSON text — every `${VAR}` anywhere in the file is resolved (command strings, URLs, args, server names), not just values inside `env` blocks. Foreign configs (Cursor, Copilot, Gemini) may use custom templating conventions that get clobbered.

**Fix**: Limit interpolation to `env` blocks inside `addServersFromFile`/`transform` rather than wholesale text substitution in `readJsonSafe`. Implementing this also cleanly eliminates Critical #1 without needing provenance tracking.

---

## Minor

### 6. `resolveEnvVars` exported at top level instead of under `MCP` namespace
`packages/opencode/src/mcp/index.ts:34` — every other symbol in the file is nested under `export namespace MCP { ... }`. Move inside the namespace and update the test import to `MCP.resolveEnvVars(...)`.

### 7. `mcp.headers` not resolved — `Authorization: Bearer ${TOKEN}` still leaks through as literal
PR fixes `mcp.environment` but not `mcp.headers`. Remote MCP configs commonly put tokens in `Authorization` headers; the discovery sources you target (`.vscode/mcp.json`, `.cursor/mcp.json`) all support a `headers` field. A config with `headers: { Authorization: "Bearer ${GITLAB_TOKEN}" }` still hits the original bug.

**Fix**: Apply the resolver (single implementation per #2) to `mcp.headers` as well. Add a test.

### 8. `{env:VAR}` silently does not support `:-default`
`packages/opencode/src/mcp/index.ts:31, 40-42` — the regex branch `\{env:([^}]+)\}` greedily captures `:-default` as part of the variable name. A user writing `{env:VAR:-default}` gets `process.env["VAR:-default"]`, which never matches. The inline comment is also misleading about the supported syntax.

**Fix**: Either add explicit support (`\{env:([A-Za-z_][A-Za-z0-9_]*)(?::-([^}]*))?\}`) or add a test asserting the current non-support and correct the comment.

### 9. Missing integration tests for `.copilot`, `.gemini`, `~/.claude.json` discovery sources
The integration tests only exercise `.vscode/mcp.json` and `.cursor/mcp.json`. If discovery for the other sources differs in how env blocks are located, the fix may silently not apply there.

**Fix**: Add tests for `.copilot/mcp.json`, `.gemini/settings.json`, and both the project-scoped and global `~/.claude.json` MCP paths.

### 10. No test for `readJsonSafe` fallback path
No test covers the case where `ConfigPaths.parseText()` throws and execution falls through to raw `parseJsonc`. This is load-bearing for error resilience and the site of #4. Worth adding a test that triggers `InvalidError` inside `parseText` (e.g., a `{file:missing}` reference) and asserts the fallback still returns a parsed server entry.

---

## NIT

### 11. Dynamic `import("../config/paths")` inside `readJsonSafe` runs per file
`packages/opencode/src/mcp/discover.ts:121` — presumably avoiding a circular dep. If none exists, switch to a top-of-file static import. Otherwise hoist to module scope so it resolves once per process.

### 12. Test teardown `process.env = { ...ORIGINAL_ENV }` reassigns the object reference
`packages/opencode/test/mcp/env-var-interpolation.test.ts:20, 97` — replacing the reference can desync from the native OS env backing store and leak keys into later tests. Prefer `delete process.env[key]` for each key set in `beforeEach`.

### 13. `ENV_VAR_PATTERN` at module scope with `/g` flag
`packages/opencode/src/mcp/index.ts:31` — mutable `lastIndex` state, safe only with `.replace()`. Current single call-site is fine, but a future `exec()`/`.test()` caller would silently break. Document the contract or recreate the regex inside the function.

---

## Positive observations

- **Two-layer defense concept is architecturally sound.** The bug report correctly identified multiple code paths; the instinct to cover all of them is right. The issue is the layers aren't isolated by provenance.
- **`$${VAR}` literal escape is implemented and tested in isolation** — many implementations forget the escape.
- **Negative lookbehind `(?<!\$)` correctly prevents `$$${VAR}` overmatching** at the regex level.
- **Fall-through behavior on interpolation failure is conservative** — the config still parses, even if without env resolution.
- **`altimate_change` markers are correctly placed** on all three modified files, satisfying Marker Guard CI.
- **`mcp.environment ? resolveEnvVars(...) : {}` handles undefined correctly** — the previous bare spread would have thrown on null.
- **Test suite organization** (unit direct + integration via `discoverExternalMcp`) is good structure.
- **Tests import from the production module** rather than re-implementing the helper — avoids stale-copy drift.

---

## Missing tests to add

1. **End-to-end `$${VAR}` preservation** — load `opencode.json` with `$${VAR}` under `mcp.environment`, assert child process env receives the literal `${VAR}`. Regression gate for Critical #1.
2. **Variable-chain injection** — `EVIL_VAR="${SECRET}"`, config references `${EVIL_VAR}`, assert result is the literal `${SECRET}`, not `SECRET`'s value.
3. **`.copilot/mcp.json`, `.gemini/settings.json`, `~/.claude.json` project + global paths** — see #9.
4. **`readJsonSafe` fallback path** — see #10.
5. **`mcp.headers` env resolution** — see #7.
6. **Unresolved `${VAR}` emits a warning** — once #3 is implemented.
7. **Env var value containing `$` (non-template)** — e.g., password `p$ssword`. Assert it survives unchanged.

---

## Rejected findings (for transparency)

- **"Dead code in `readJsonSafe` — fallback unreachable"** was flagged CRITICAL by 2 models initially, both accepted the rejection at convergence. The `return undefined` lives inside the `catch` of the outer `readText` try/catch. On successful read, execution falls through to the new `altimate_change` block, which is fully reachable. Both branches work as intended.
- **"Test file copy-pastes `ENV_VAR_PATTERN` and `resolveEnvVars`"** — flagged as a concern, confirmed not applicable to PR head. Tests import from the production module.

---

## Recommended minimum fix set before merge

1. Fix Critical #1 (pick one resolution layer)
2. Address Majors #2, #3, #4
3. Add the Critical #1 regression test and chain-injection test
4. **Preferred direction**: implement Major #5 (scope Layer 1 to `env` blocks), which cleanly eliminates the double-pass without requiring provenance tracking

Minors #6–#10 can be follow-ups if you prefer, but #7 (headers) is a genuine gap in the stated fix scope for #656 and probably should land with the same PR.

Happy to re-review once changes are in.

---

<details>
<summary>Review methodology</summary>

Reviewed by 9 participants: Claude Opus 4.6 + 8 external models via OpenRouter (GPT 5.4, Gemini 3.1 Pro, Kimi K2.5, MiniMax M2.7, GLM-5, Qwen 3.6 Plus, DeepSeek V3, MiMo-V2-Pro). Each model independently reviewed the PR with full codebase access before findings were synthesized and sent back for a convergence round. 5/8 external models responded in convergence (above quorum=2), all confirming CHANGES NEEDED with no accuracy objections.

| Severity | Count | Models that flagged (examples) |
|---|---|---|
| Critical #1 (double interpolation) | 7/8 consensus | GPT 5.4, Gemini (POC), GLM-5, MiniMax, Kimi, DeepSeek, Qwen |
| Major #2 (regex duplication) | 8/8 consensus | all |
| Major #3 (silent empty-string) | 5 | Claude, GLM-5, MiniMax, DeepSeek, Qwen |
| Major #4 (swallowed error) | 4 | Claude, GLM-5, DeepSeek, Gemini |
| Major #5 (whole-text substitution) | 1 unique | DeepSeek |
| Minor #7 (headers not resolved) | 1 unique | Qwen |

</details>

PR #692, opened 2026-04-10
fix: harden sql_explain and altimate_core_validate input handling

### LINKED ISSUE #691: sql_explain and altimate_core_validate: poor input validation and unhelpful errors cause retry loops
## Problem

Two tools produce high failure rates and trigger agent retry loops because their input validation and error messages do not match their parameter contracts.

### `altimate_core_validate`

**Contract mismatch.** The parameter schema declares `schema_path` and `schema_context` as `.optional()`, but the runtime hard-gates on their absence and returns a terse error before the call even reaches the core library:

```
"No schema provided. Provide schema_context or schema_path so table/column references can be resolved."
```

The Rust core (`altimate-core`) accepts an empty schema just fine (there is already a ``schemaOrEmpty`` helper for exactly this case), so the hard-gate is unnecessary. Because the schema *looks* optional to the caller, calls without schema are the common path and every one of them fails identically.

**Expected behavior:** the tool should run with an empty schema when neither `schema_path` nor `schema_context` is provided, surface a clear warning in the output that table/column references cannot be resolved, and still return semantic findings (syntax errors, dialect issues) that do not require a schema.

### `sql_explain`

**Missing input validation and verbatim DB errors.** Several issues stack on top of each other:

1. **No pre-execution validation.** `execute()` passes the `sql` string straight into ``\`${explainPrefix} ${params.sql}\``` without checking if it is empty, whitespace-only, or a bare placeholder character. An LLM-generated parameterized query like ``SELECT * FROM users WHERE id = ?`` sent through `sql_explain` ends up as ``EXPLAIN ?``, which produces:
   ```
   OperationFailedError: SQL compilation error: syntax error line 1 at position 8 unexpected ?.
   ```

2. **No warehouse name validation.** If the `warehouse` parameter is an empty string or a placeholder, the Registry lookup returns:
   ```
   Error: Connection ? not found. Available: (none)
   ```
   The error does not list what warehouses *are* available, so the caller has no way to recover.

3. **Error messages are not actionable.** Both of the above errors are surfaced verbatim from Snowflake / the Registry with no translation. A calling agent has no signal that the problem was a placeholder character versus a real SQL syntax issue, so it retries the same broken call.

4. **`EXPLAIN` is not dialect-aware.** The handler hard-codes ``EXPLAIN`` and ``EXPLAIN ANALYZE`` prefixes. Snowflake does not support ``EXPLAIN ANALYZE`` (it uses ``EXPLAIN USING TEXT`` / ``USING JSON``), and BigQuery requires a dry-run API call rather than an ``EXPLAIN`` statement at all.

**Expected behavior:** reject empty/placeholder inputs with an actionable error before hitting the warehouse, translate common DB errors into guidance that explains *how to fix the call*, and use dialect-specific EXPLAIN syntax.

## Scope

Both fixes live in ``packages/opencode/src/altimate``. The Rust core (`altimate-core-internal`) does not need changes — its ``validate()`` function already accepts empty schemas, and ``sql_explain`` does not call into Rust at all.

Files affected:
- `packages/opencode/src/altimate/tools/sql-explain.ts` (input validation)
- `packages/opencode/src/altimate/native/connections/register.ts` (dialect-aware EXPLAIN, error translation)
- `packages/opencode/src/altimate/tools/altimate-core-validate.ts` (remove hard-gate, add warning path)

## Acceptance Criteria

- `sql_explain` rejects empty SQL, whitespace-only SQL, and bare placeholder characters (`?`, `:var`, `$1`) with a clear actionable error message before executing any warehouse call
- `sql_explain` translates "Connection X not found" errors into messages that include the list of configured warehouses (or a suggestion to run `warehouse_add`)
- `sql_explain` translates SQL compilation errors related to placeholder characters into guidance that explicitly says "sql_explain does not support parameterized queries"
- `sql_explain` uses the correct EXPLAIN prefix for Snowflake (`EXPLAIN USING TEXT`), PostgreSQL (`EXPLAIN (ANALYZE, BUFFERS)`), MySQL, and other supported warehouses
- `altimate_core_validate` runs with an empty schema when neither `schema_path` nor `schema_context` is provided, returning semantic findings and a clear warning that schema-dependent checks were skipped
- `altimate_core_validate` exposes `has_schema` in metadata so callers can distinguish full validation from schema-less runs
- Unit tests cover the above paths for both tools

anandgupta42:
## Summary

Two tools produced high failure rates because their runtime behaviour did not match the parameter contracts they advertised. Both fixes live entirely in TypeScript under packages/opencode/src/altimate; the Rust altimate-core engine does not need changes (its validate already accepts empty schemas, and sql_explain never calls into Rust).

### What changed

**sql_explain**

- Pre-flight input validation. Reject empty, whitespace-only, and placeholder-only SQL (bare `?`, :var, $1, @name) before the call reaches the warehouse. Also reject empty or placeholder warehouse names with a pointer to warehouse_list. These now return a clear input_validation error envelope instead of surfacing verbatim driver errors.
- Dialect-aware EXPLAIN via a new buildExplainPlan helper that returns prefix plus actuallyAnalyzed so the handler picks warehouse-correct syntax AND reports the true plan mode. Warehouse coverage: Snowflake uses EXPLAIN USING TEXT (no ANALYZE), Postgres uses plain or ANALYZE-BUFFERS, Redshift uses plain EXPLAIN only — it does NOT support ANALYZE, MySQL / MariaDB / DuckDB support EXPLAIN ANALYZE, Databricks / Spark use EXPLAIN FORMATTED, ClickHouse uses plain EXPLAIN. BigQuery, Oracle, and SQL Server return a clear "not supported via statement prefix" error instead of issuing a broken statement.
- Fixes a subtle bug where Snowflake users who requested analyze:true saw analyzed:true in metadata even though the handler silently sent plain EXPLAIN. The result envelope now reports the real mode.
- New translateExplainError helper rewrites common driver errors into actionable guidance: connection-not-found lists available warehouses, unsubstituted-bind compile errors tell the caller to inline the literal value, permission denials call out role grants, generic compile errors ask for a SQL fix.
- Intentional non-change: the tool does not scan mid-query for stray question marks because PostgreSQL JSONB uses them as legitimate key-exists operators. Only the bare-placeholder check runs, which has no false positives.

**altimate_core_validate**

- Remove the hard-gate that returned "No schema provided..." before the engine ran. The parameter schema already declared schema_path and schema_context as optional, and schemaOrEmpty in the dispatcher has always been able to build a minimal empty schema for exactly this case.
- When schema is absent, the engine still runs so syntax and dialect findings come through. The tool surfaces a clear warning in the output and exposes has_schema:false in metadata so callers can distinguish full validation from schema-less runs.
- Dispatcher-thrown errors are now classified as engine_failure to keep them distinct from user-input problems.
- Extracted the no-schema warning into a module-level constant and fixed a formatting bug where the note's leading blank lines combined with a newline join produced three consecutive newlines in the rendered output.

### Type of change

- [x] Bug fix
- [x] Refactor (no functional change)
- [x] Tests (adds missing tests)

### Issue for this PR

Closes #691

New and updated unit tests under packages/opencode/test/altimate/tools cover input validation, warehouse-name validation, the full buildExplainPlan warehouse matrix including Snowflake, Redshift, BigQuery, Oracle, and SQL Server regressions, translateExplainError paths, SqlExplainTool.execute integration with a mocked dispatcher, classifyValidationError precedence, extractValidationErrors, formatValidate (including regression guards that assert no triple-newline in the rendered output), and AltimateCoreValidateTool.execute paths for no schema, inline schema_context, schema_path, empty schema_context, invalid findings classification, and dispatcher throws. The existing tool-error-propagation regression test was updated to reflect the new contract.

Local verification: marker check passes, typecheck passes with zero errors, full altimate suite reports 2745 passing and 0 failing (up from 2739 on main).

### Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have added tests that prove my fix is effective
- [x] The branch is rebased on the latest main
- [x] Marker Guard check passes locally
- [x] TypeScript typecheck passes locally

### DIFF
--- a/packages/opencode/src/altimate/native/connections/register.ts
+++ b/packages/opencode/src/altimate/native/connections/register.ts
@@ -182,6 +182,147 @@ function getWarehouseType(warehouseName?: string): string {
   return Registry.getConfig(warehouseName)?.type ?? "unknown"
 }
 
+/**
+ * Dialect-aware EXPLAIN plan describing how a warehouse should be asked for
+ * a query plan.
+ *
+ *   - `prefix`: the statement prefix to prepend to the SQL ("" means this
+ *     warehouse cannot be EXPLAIN'd via a simple statement prefix — the
+ *     handler will return a clear error instead of issuing a broken query)
+ *   - `actuallyAnalyzed`: whether the prefix actually runs the query and
+ *     reports runtime stats. The caller requested `analyze=true`, but some
+ *     warehouses silently downgrade to plan-only (e.g. Snowflake), so we
+ *     reflect the true mode back to the user in the result envelope.
+ */
+export interface ExplainPlan {
+  prefix: string
+  actuallyAnalyzed: boolean
+}
+
+/**
+ * Build a dialect-aware EXPLAIN plan for the given warehouse type.
+ *
+ * Warehouse-specific notes:
+ *   - Snowflake: `EXPLAIN USING TEXT`. No ANALYZE variant — silently downgraded.
+ *   - PostgreSQL: `EXPLAIN` or `EXPLAIN (ANALYZE, BUFFERS)` for runtime stats.
+ *   - Redshift: plain `EXPLAIN` only. Redshift does NOT support ANALYZE.
+ *   - MySQL / MariaDB: `EXPLAIN` or `EXPLAIN ANALYZE` (MySQL 8+).
+ *   - DuckDB: `EXPLAIN` or `EXPLAIN ANALYZE`.
+ *   - Databricks / Spark: `EXPLAIN` or `EXPLAIN FORMATTED`.
+ *   - ClickHouse: `EXPLAIN` (no ANALYZE form accepted via statement prefix).
+ *   - BigQuery: uses a dry-run API instead of any EXPLAIN statement. Not
+ *     supported via this code path — return empty prefix.
+ *   - Oracle: `EXPLAIN PLAN FOR` stores the plan in PLAN_TABLE rather than
+ *     returning it, so a bare prefix does not produce output. Not supported.
+ *   - SQL Server: requires `SET SHOWPLAN_TEXT ON` as a session setting.
+ *     Not supported via statement prefix.
+ */
+export function buildExplainPlan(warehouseType: string | undefined, analyze: boolean): ExplainPlan {
+  const type = (warehouseType ?? "").toLowerCase()
+  switch (type) {
+    case "snowflake":
+      // Snowflake: no ANALYZE; USING TEXT returns a readable plan.
+      return { prefix: "EXPLAIN USING TEXT", actuallyAnalyzed: false }
+    case "postgres":
+    case "postgresql":
+      return analyze
+        ? { prefix: "EXPLAIN (ANALYZE, BUFFERS)", actuallyAnalyzed: true }
+        : { prefix: "EXPLAIN", actuallyAnalyzed: false }
+    case "redshift":
+      // Redshift supports only plain EXPLAIN — no ANALYZE/BUFFERS.
+      return { prefix: "EXPLAIN", actuallyAnalyzed: false }
+    case "mysql":
+    case "mariadb":
+      // MySQL 8+ supports `EXPLAIN ANALYZE`. Older versions will reject it
+      // at the warehouse and the error will be surfaced to the caller.
+      return analyze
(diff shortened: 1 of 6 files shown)


anandgupta42:
Superseded by #693 (same branch, squashed commits + shorter description). Closing this one.

PR #757, opened 2026-04-25
feat: bridge upstream v1.4.0 across history rewrite + 3 backports + adversarial test suite

anandgupta42:
### What does this PR do?

Bridges upstream `anomalyco/opencode` `v1.4.0` into our fork across an upstream history rewrite (`v1.3.17` → `v1.4.0`, 2026-04-04). Our standard `script/upstream/merge.ts` cannot handle this — there is zero common ancestor between our `main` and any upstream tag from `v1.3.16` onward.

The PR has 3 layers:

1. **Bridge merge** (commit `b84264255a`+ cycles 1–7): one-off `script/upstream/bridge-merge.ts` that overlays `v1.4.0` while preserving altimate-owned code, plus 7 audit cycles fixing typecheck drift, runtime bugs, branding leaks, async-callsite drift, sync-transaction lock, service-identifier collisions, and 56 lock-in regression tests.
2. **3 dropped-feature backports** (commits `a1eafbcbfa`, `2a47d5efa8`, `51dafa5e82`):
   - `--dangerously-skip-permissions` flag (upstream PR #21266) — aliased to our existing `--yolo` (which respects deny rules; strictly safer than upstream).
   - `variant_list` TUI keybind (upstream PR #21185) — opens DialogVariant.
   - `maskString` hardening — strips unquoted `sk-`, `sk-ant-`, `Bearer …` API-key/token patterns from telemetry to close a data-leakage path.
3. **Defensive instrumentation + flake fixes** (commits `ee73863e9d`, `64906babc2`, `38463876bc`, `5299f29670`, `6a0d6fa848`):
   - `agent_outcome` event now populates `final_tool`/`error_class`/`reason` (was empty for ~30% of non-completed builder runs).
   - `Plugin.trigger` wraps hook invocation in try/catch so a buggy plugin can't crash the chat call.
   - `Recap.flushSync` race against async `snapshot()` closed via a `crashed` flag (deterministic).
   - `Recap.flush()` helper added so tests don't race against async writes (replaces flaky `setTimeout(50)`).
   - `classifyError` ERROR_PATTERNS extended to cover bare 5xx + rate-limit text.

### Bridge strategy

| Category | Action | Count |
|---|---|---|
| In v1.4.0, NOT in keepOurs/skipFiles | Take v1.4.0 | 730 |
| In v1.4.0, in `keepOurs` | Leave main's version | 65 |
| In v1.4.0, in `skipFiles` | Exclude | 3676 |
| Has `altimate_change` markers in main | Keep main's version entirely | 61 |
| In main, not in v1.4.0, no rule matched | Keep + flag for review | 243 |

The marker re-application strategy from earlier attempts produced duplicated code (markers don't carry replace-vs-insert semantics). Skipping it entirely and keeping main's version of marker files is the safer compromise — preserves all altimate code at the cost of not picking up upstream's improvements to those 61 files; followups handle that incrementally.

### Type of change
- [x] This change requires a documentation update
- [x] Tooling
- [x] Upstream merge

### Issue for this PR
N/A — this is an upstream-merge `feat:` PR. The `pr-standards` workflow exempts `docs|refactor|feat` titles from the issue-link requirement (see `.github/workflows/pr-standards.yml:114-119`). Tracked internally via the bridge-merge audit cycles documented in `.bridge-merge-report.md`.

**Build / type / lint**
- `bun turbo typecheck`: 5/5 packages clean (TypeScript CI: pass, 4m45s)
- `bun run script/build.ts --skip-install --single`: 1.3s, 132M arm64 binary, runs `--version` + `run --help` correctly with backports visible
- `bunx prettier --check` on all 297 PR-changed `.ts`/`.tsx` files: clean
- Marker check: 471 `altimate_change` pairs, all properly closed; 0 branding leaks (`opencode.ai` / `anomalyco` / `OpenCode`)

**Tests**
- Local full suite: 7949 pass / 503 skip / 0 functional fail (2 pre-existing flaky tests under local resource contention pass on CI and pass when isolated)
- v1.4.0-specific adversarial regression suites added in this PR: **87 tests, 3593 assertions, 0 fail**
  - `test/upstream/v140-merge-adversarial.test.ts` — 30 static invariants (R1)
  - `test/upstream/v140-merge-fuzz.test.ts` — 17 property-based fuzz tests (R2, 3463 assertions)
  - `test/upstream/v140-merge-chaos.test.ts` — 24 chaos/concurrency/migration/mutation tests (R3)
- Mutation-testing kill rate (R3): 6 of 7 deliberate code regressions caught by the suite

**Adversarial review**
- 3 distinct adversarial methodologies (subsystem-by-subsystem, fuzz/runtime, chaos/concurrency/security) plus a multi-model consensus pre-merge review (Claude + GPT 5.4, Gemini 3.1 Pro, Kimi K2.5, MiniMax M2.7, GLM-5, DeepSeek V3.2). 5 of 6 substantive reviewers explicitly APPROVE. The 1 dissent's findings were either verified false on the actual code or are intentional design choices already documented.

### Checklist
- [x] All 11 required CI checks pass on the latest commit (`6a0d6fa848`): TypeScript, dbt-tools, Marker Guard, Detect Changes, anti-slop, check-compliance, check-standards, cubic, CodeRabbit, plus skipped-as-intended Driver E2E / Sanity / dbt-tools E2E
- [x] Persistent regression test suite added (`v140-merge-{adversarial,fuzz,chaos}.test.ts`) — runs on every future bridge merge
- [x] Bridge tooling committed for reproducibility (`script/upstream/bridge-merge.ts`)
- [x] Per-file report saved as `.bridge-merge-report.md`
- [x] User-facing documentation updated for the 3 backports (`docs/docs/usage/cli.md`, `docs/docs/configure/keybinds.md`, `docs/docs/reference/telemetry.md`)
- [x] Branding scan clean, 0 leaks
- [x] No unrelated changes; everything in this PR is the v1.4.0 bridge merge or its audit/fix cycles

### Followup work (NOT in this PR — non-blocking)

1. Manual upstream merge for the 61 marker files — incrementally integrate upstream's improvements while keeping `altimate_change` blocks intact.
2. Triage the 243 review-list files — for each, either add to `keepOurs` (altimate code that isn't covered by globs) or add to `skipFiles` and delete (truly stale upstream code).
3. `OPENCODE_STRICT_PLUGINS` dev flag (Kimi K2.5 suggestion) — re-throw plugin hook errors in dev mode so a developer notices the silent-error path instead of just seeing the log line.
4. Runtime functional test for `Plugin.trigger` (currently regex-only source-text assertion).

### Discoveries surfaced

- Upstream rewrote git history at `v1.3.16`+ (only 10–59 commits in their lineage from `v1.4.0` back). The standard merge tooling cannot bridge this; the new `bridge-merge.ts` is required.
- `git ls-tree -r v1.4.0` returns 669 files; `git ls-tree -r --full-tree v1.4.0` returns 4532 — `--full-tree` is required when invoking from a subdirectory or for trees with unusual entries.
- `script/upstream/utils/config.ts` `keepOurs` / `skipFiles` checks were missing `{ dot: true }` for minimatch — files inside dot-prefixed directories weren't matching globs. Fixed in `bridge-merge.ts`.

🤖 Bridge merge tooling and adversarial test suite developed iteratively with [Claude Code](https://claude.com/claude-code) over 3 audit rounds + pre-merge ritual. Human review and design decisions throughout.


dev-punia-altimate:

- baseline [0.49ms]
- baseline [0.34ms]
- baseline [0.06ms]
- baseline [0.21ms]
- connection_refused [0.34ms]
- timeout [0.08ms]
- permission_denied [0.06ms]
- parse_error [0.05ms]
- oom [0.06ms]
- network_error [0.06ms]
- auth_failure [0.05ms]
- rate_limit [0.06ms]
- internal_error [0.05ms]
- empty_error [0.05ms]
- connection_refused [0.06ms]

### Next Step

cc @anandgupta42

kulvirgit:
# Code Review — `upstream/merge-v1.4.0` vs main merge-base (`941c8cae2`)

**Verdict**: REQUEST CHANGES
**Critical**: 6 · **Major**: 7 · **Minor**: 5

Lens: code at face value, looking for concrete user-facing breakage in the altimate-code TUI / CLI. Every line citation below is grep-verified against the current branch.

---

## Critical Issues

### 1. Basic-auth username defaults to `"opencode"` in 3 client paths; server expects `"altimate"` → 401 on every password-protected connection

- **Locations**:
  - `packages/opencode/src/cli/cmd/run.ts:873` — `process.env.OPENCODE_SERVER_USERNAME ?? "opencode"`
  - `packages/opencode/src/cli/cmd/tui/worker.ts:331` — `Flag.OPENCODE_SERVER_USERNAME ?? "opencode"`
  - `packages/opencode/src/plugin/index.ts:51` — `Flag.OPENCODE_SERVER_USERNAME ?? "opencode"`
- **User scenario**: User sets `OPENCODE_SERVER_PASSWORD` but not `OPENCODE_SERVER_USERNAME`, then either:
  - runs `altimate run --attach http://localhost:4096 -p <pw>` → 401
  - starts the TUI against a password-protected server → worker can't fetch session/permission state, UI is blank or errors loop
  - any plugin that self-connects to the local server → auth fails, plugin features dead
- **Description**: The branch contains an explicit `altimate_change` block at `cli/cmd/tui/attach.ts:69-72` that hardcodes the username to `"altimate"` and explains the *why*: "Upstream's was 'opencode'; restoring it would break the attach handshake against altimate-code servers." Server side at `server/server.ts:84` matches: `Flag.OPENCODE_SERVER_USERNAME ?? "altimate"`. But three sibling client call sites still default to `"opencode"`. The fix was applied to one file; the same pattern in three others was missed.
- **Fix**: Change all three defaults to `"altimate"`.

### 2. Fatal-error TUI screen builds a GitHub issue URL with `"opentui: fatal:"` title and `"opencode-version"` query param; the rebranded fix is dead code

- **Locations**:
  - `packages/opencode/src/cli/cmd/tui/app.tsx:216` — uses inline `ErrorComponent` (defined at line 936)
  - `packages/opencode/src/cli/cmd/tui/app.tsx:971` — `issueURL.searchParams.set("title", \`opentui: fatal: ${...}\`)`
  - `packages/opencode/src/cli/cmd/tui/app.tsx:981` — `issueURL.searchParams.set("opencode-version", ...)`
  - `packages/opencode/src/cli/cmd/tui/component/error-component.tsx` (95-line NEW file with **correct** branding) — **never imported** (`grep -rn "error-component"` returns 0 importers)
- **User scenario**: TUI crashes. Error screen offers a "report issue" URL. Title is `"opentui: fatal: <msg>"`; query string has `?opencode-version=...`. User clicks → lands on a GitHub issue form with stale field names; if the bug-tracker dispatch routes by params, points away from `AltimateAI/altimate-code`.
- **Fix**: Either import the new `<ErrorComponent>` from `./component/error-component` and delete the inline duplicate, or apply the same branding fix to the inline implementation in `app.tsx`.

### 3. `cli/cmd/uninstall.ts:185-191` package-name map uses upstream names — `altimate uninstall` silently fails for npm/pnpm/bun/yarn/brew users

- **Location**: `packages/opencode/src/cli/cmd/uninstall.ts:185-191`
- **User scenario**: User installed via `npm install -g @altimateai/altimate-code` (or `brew install altimate-code`). Runs `altimate uninstall`. Command resolves to `npm uninstall -g opencode-ai` or `brew uninstall opencode` — package not found, error or no-op, real install left in place.
- **Description**: Package map literally is:
  ```ts
  const cmds: Record<string, string[]> = {
    npm: ["npm", "uninstall", "-g", "opencode-ai"],
    pnpm: ["pnpm", "uninstall", "-g", "opencode-ai"],
    bun: ["bun", "remove", "-g", "opencode-ai"],
    yarn: ["yarn", "global", "remove", "opencode-ai"],
    brew: ["brew", "uninstall", "opencode"],
    choco: ["choco", "uninstall", "opencode"],
    scoop: ["scoop", "uninstall", "opencode"],
  }
  ```
- **Fix**: Update npm/pnpm/bun/yarn entries to `"@altimateai/altimate-code"` and brew to `"altimate-code"`. Choco/scoop may be intentional (deliberate decision about Windows distribution channels).

### 4. OTLP `serviceName` hardcoded to `"opencode"` → telemetry mislabels altimate-code instances in user observability backends

- **Location**: `packages/opencode/src/effect/oltp.ts:16` — `serviceName: "opencode"`
- **User scenario**: User configures `OTEL_EXPORTER_OTLP_ENDPOINT` to ship traces to Honeycomb / Datadog / signoz. Their backend shows the service as `"opencode"`; runbooks and dashboards keyed on `service.name=altimate-code` find nothing.
- **Fix**: `serviceName: "altimate-code"`.

### 5. `config/tui.ts:57` filters out `.altimate-code` directories — TUI customizations in `.altimate-code/tui.json` are silently ignored

- **Location**: `packages/opencode/src/config/tui.ts:57`
  ```ts
  if (!dir.endsWith(".opencode") && dir !== Flag.OPENCODE_CONFIG_DIR) continue
  ```
- **User scenario**: User puts a custom theme, keybinding, or TUI tweak in `.altimate-code/tui.json` (the branded path). Loader iterates dirs, checks `endsWith(".opencode")`, skips. Customization silently dropped — TUI launches with defaults.
- **Compounded by**: Even if the file were in `.opencode/tui.json`, `TuiInfo` schema at `config/tui-schema.ts:27` does not declare `plugin` / `plugin_enabled` keys, so the TUI plugin install path is broken end-to-end (see #6 too).
- **Fix**:
  ```ts
  if (!dir.endsWith(".opencode") && !dir.endsWith(".altimate-code") && dir !== Flag.OPENCODE_CONFIG_DIR) continue
  ```
  And extend `TuiInfo` to include the `plugin` / `plugin_enabled` fields the installer writes.

### 6. Plugin config schema is `string[]` but the installer can write tuple entries `[name, opts]` → next startup either rejects `altimate-code.json` or crashes the plugin loader

- **Locations**:
  - `packages/opencode/src/plugin/install.ts:407` — `const item = target.opts ? ([spec, target.opts] as const) : spec` (writes a tuple when the plugin package exports default options)
  - `packages/opencode/src/config/config.ts:1112` — `plugin: z.string().array().optional()` (schema accepts only string entries)
  - `packages/opencode/src/plugin/index.ts:83-90` — runtime: `for (let plugin of plugins) { ... if (!plugin.startsWith("file://")) ... }` (assumes string)
- **User scenario**: User installs any plugin whose package.json declares default options. The installer patches `altimate-code.json` with a tuple entry. On next launch:
  1. Zod validates the new config against `plugin: string[]` → rejection, all custom config is dropped.
  2. Even if validation is lax, the runtime calls `plugin.startsWith("file://")` on a tuple → TypeError.
- **Fix**: Update the schema to accept tuples (e.g. `z.array(z.union([z.string(), z.tuple([z.string(), z.record(z.any())])])).optional()`), preserve tuple values through `dedupe`/`resolve`, and update `plugin/index.ts` runtime to handle both shapes.

---

## Major Issues

### 7. `cli/cmd/plug.ts` (235 LOC, NEW) exports `PluginCommand` but is never registered in the dispatcher → `altimate plug …` returns "Unknown command"

- **Locations**:
  - `packages/opencode/src/cli/cmd/plug.ts:180` — `export const PluginCommand = cmd({ ... })`
  - `packages/opencode/src/index.ts` — does not import or `.command(PluginCommand)`
- **User scenario**: User reads release notes / `--help` and types `altimate plug install …` → yargs reports unknown command. Feature ships dark.
- **Fix**: Add `import { PluginCommand } from "./cli/cmd/plug"` and `.command(PluginCommand)` in `packages/opencode/src/index.ts`. (Or delete `plug.ts` if unintended.)

### 8. Four NEW prompt files (`codex.txt`, `default.txt`, `gpt.txt`, `kimi.txt`) shipped to `session/prompt/` but never imported

- **Locations**:
  - `packages/opencode/src/session/prompt/{codex,default,gpt,kimi}.txt` (NEW)
  - `packages/opencode/src/session/system.ts:5-11` — imports only `anthropic.txt`, `qwen.txt`, `beast.txt`, `gemini.txt`, `codex_header.txt`, `trinity.txt`. None of the new files are imported.
- **User scenario**: User selects a Kimi/GPT/Codex/default model expecting the new model-tuned prompt. Runtime falls back to `PROMPT_CODEX` (from the existing `codex_header.txt`) for `gpt-5*`, or `PROMPT_ANTHROPIC_WITHOUT_TODO` (from `qwen.txt`) for everything else. Behavior unchanged from before.
- **Fix**: Wire new prompts into `system.ts` selection logic, or delete the orphan files.

### 9. Provider `X-Title: "opencode"` header leaks upstream brand to OpenRouter / Vercel / Zenmux / Kilo dashboards

- **Locations**: `packages/opencode/src/provider/provider.ts:458, 568, 722` (verified by `rg -n`)
- **User scenario**: User sets up OpenRouter/Vercel/Zenmux/Kilo. Their provider dashboard shows "opencode" as the calling app — confusing for shared accounts; harmful for customers wondering why "opencode" is hitting their key.
- **Fix**: Replace `"X-Title": "opencode"` with `"X-Title": "altimate-code"` at all three lines.

### 10. Plugin install error message: `"Plugin requires opencode ${range} but running ${opencodeVersion}"`

- **Location**: `packages/opencode/src/plugin/shared.ts:205`
- **User scenario**: User installs a plugin whose version range doesn't match. Sees an error mentioning "opencode" in their own CLI's output.
- **Fix**: Reword to "Plugin requires altimate-code ${range}…" or generic "Plugin requires version ${range}…".

### 11. `config/config.ts:49-56` `systemManagedConfigDir` hardcodes upstream paths — enterprise managed config in `/etc/altimate-code` ignored

- **Location**: `packages/opencode/src/config/config.ts:49-56`
  ```ts
  case "darwin":  return "/Library/Application Support/opencode"
  case "win32":   return path.join(process.env.ProgramData || "C:\\ProgramData", "opencode")
  default:        return "/etc/opencode"
  ```
- **User scenario**: Enterprise admin deploys `altimate-code.json` to `/etc/altimate-code/` (or the macOS/Windows equivalents). The CLI silently reads from `/etc/opencode/` instead and finds nothing. Managed defaults never reach the user.
- **Fix**: Switch the three paths to `altimate-code` (and either drop the old paths or add fallback reads).

### 12. `config/config.ts:429,468` `loadCommand`/`loadAgent` patterns only check `.opencode/` — nested commands/agents in `.altimate-code/` lose their namespace

- **Locations**:
  - `packages/opencode/src/config/config.ts:429` — `["/.opencode/command/", "/.opencode/commands/", "/command/", "/commands/"]`
  - `packages/opencode/src/config/config.ts:468` — `["/.opencode/agent/", "/.opencode/agents/", "/agent/", "/agents/"]`
- **User scenario**: User puts a nested command at `.altimate-code/commands/utils/format.md`. Pattern match fails → falls back to `path.basename(item)` → silently drops the `utils/` segment → namespace collapses → `/utils/format` becomes `/format`, potentially colliding with another command.
- **Fix**: Add `.altimate-code` variants to both pattern arrays.

### 13. `dialog-skill.tsx:97, 279` — user-visible success message tells the user files were created at `.opencode/skills/${name}/`

- **Locations**:
  - `packages/opencode/src/cli/cmd/tui/component/dialog-skill.tsx:97` — `Created skill + tool at .opencode/skills/${name}/`
  - `packages/opencode/src/cli/cmd/tui/component/dialog-skill.tsx:279` — `✓ Created "${name}"\n\nSkill + CLI tool at .opencode/skills/${name}/\nType /${name} in the prompt to use it.`
- **User scenario**: User creates a custom skill via the TUI dialog. Success toast directs them to look in `.opencode/skills/`. User goes hunting in the wrong directory; either confused or skill files actually written elsewhere.
- **Fix**: Update message strings to use `.altimate-code/skills/${name}/` (or the fork's actual write path — make it match wherever the skill is in fact saved).

---

## Minor Issues

### 14. `cli/cmd/run.ts:316` help text exposes upstream env-var name

- `--password  basic auth password (defaults to OPENCODE_SERVER_PASSWORD)` — visible whenever a user runs `altimate run --help`.
- **Fix**: Either rename the env var to `ALTIMATE_SERVER_PASSWORD` (with `OPENCODE_SERVER_PASSWORD` fallback for compat — recommended), or drop the brand from the help text.

### 15. mDNS default hostname is `"opencode.local"` — visible in users' Bonjour/Avahi network browsers

- **Locations**:
  - `packages/opencode/src/server/mdns.ts:15` — fallback `"opencode.local"`
  - `packages/opencode/src/config/config.ts:1021` — `mdnsDomain` describe text says "default: opencode.local"
- **Fix**: Default to `"altimate-code.local"`.

### 16. `app.tsx:676` registers internal slash-command value as `"opencode.status"` (rather than `"altimate.status"`)

- The user-typed slash `/status` still works (driven by `slash.name: "status"`), but the internal `value` field leaks the upstream brand and is the wrong handle if any code keys off the value.
- **Fix**: `value: "altimate.status"`.

### 17. `clipboard.ts:48` writes clipboard image to temp path `opencode-clipboard.png`

- User copies an image in the TUI; temp file written to `$TMPDIR/opencode-clipboard.png`. Brand leak in filesystem, visible to anyone who lists `/tmp` on a shared machine.
- **Fix**: Rename to `altimate-code-clipboard.png`.

### 18. `x-opencode-*` HTTP headers (`x-opencode-proxy-url`, `x-opencode-directory`, `x-opencode-workspace`)

- **Locations**: `server/proxy.ts:24-25,60,120`, `server/server.ts:195-196`, `server/router.ts:33,45,96`
- Internally consistent today (client and server both use the same upstream-named headers), so no immediate break. Latent mismatch if a future SDK version is rebranded — flagged for tracking.

---

## Findings flagged but DROPPED on verification

- **`pr.ts:89` regex `opncd.ai/s/`**: not present in current source (`grep -nE "opncd\.ai" packages/opencode/src/cli/cmd/pr.ts` returns no match). False alarm.
- **`OAUTH_DUMMY_KEY` barrel-export crash**: `auth/service.ts` still exists and is imported by `provider/auth-service.ts:7`. The named export was relocated to `auth/index.ts` but no caller imports `OAUTH_DUMMY_KEY` from `./service` — verified by grep. Dropped.

## Findings flagged but UNVERIFIED (track as follow-ups)

- **`auth/index.ts` Effect Schema vs Zod error shape**: `auth/index.ts:38` migrates `Auth.Info` to Effect Schema with a `.zod` adapter. Any UI catch-block expecting `err.errors` / `err.flatten()` from a Zod-style error may misbehave. Server validator at `server/server.ts:155` uses `Auth.Info.zod` correctly; broader caller audit not performed.
- **`auth/service.ts:49` (`@opencode/Auth` tag) vs `auth/index.ts:57` (`@opencode/Auth.cli` tag)**: Both runtimes coexist, similar to the `Permission` / `PermissionNext` split that audit commit `33ffd9c51` already addressed. Whether this causes a runtime crash depends on which runtime resolves the tag first. Worth a runtime smoke test before merging.

---

## Positive observations

- Logo brand colors restored (`tui/component/logo.tsx:79-82` uses `theme.primary` / `theme.accent`)
- Permission deadlock fix correctly applied: `server/routes/permission.ts` routes through `PermissionNext`, matching every runtime caller (`session/processor.ts`, `session/prompt.ts`, `session/llm.ts`, `tool/task.ts`, `agent/agent.ts`, `skill/skill.ts`, `server/routes/session.ts`)
- Terminal title shows "Altimate Code" (`app.tsx:393, 400`)
- Upgrade indicator shows `altimate upgrade` (`tui/component/upgrade-indicator.tsx`)
- `Auth.Info.zod` adapter preserved so `server/server.ts:155` validator keeps working
- `SessionStatus.set` async migration: every caller in `session/prompt.ts` and `session/processor.ts` properly awaited
- `bus/bus-event.ts` `define()` made idempotent so re-imports don't crash
- `cli/cmd/tui/attach.ts:71` correctly hardcodes username `"altimate"` with reason-of-record comment
- `server/server.ts:84` correctly defaults username `"altimate"`
- `tui-migrate.ts:16` config schema URL points to altimate.ai
- Variant keybind backport (PR #21185) wired
- `--dangerously-skip-permissions` flag backport (PR #21266)
- GitLab User-Agent shows `altimate-code` in `provider.ts:588`
- Dual config filename support (`altimate-code.json` and `opencode.json`) at the file-name level
- Plugin hook isolation — `plugin/index.ts` plugin hook failures caught and logged instead of propagating to LLM call

## Missing tests

- No test for `--attach` auth handshake with `OPENCODE_SERVER_PASSWORD` set but no username — would have caught issue #1
- No test exercising `altimate uninstall` against a fake package-manager simulation — would have caught #3
- No tree-wide invariant asserting every `cli/cmd/*.ts` is registered in the dispatcher — would have caught #7
- No tree-wide invariant asserting every `session/prompt/*.txt` is imported by `system.ts` — would have caught #8
- No test asserting the inline `ErrorComponent` in `app.tsx` produces a properly-branded issue URL — would have caught #2
- No test for end-to-end TUI plugin install→reload, with a plugin that exports default options — would have caught #5 + #6 together
- No test for nested commands/agents loaded from `.altimate-code/commands/<subdir>/foo.md` — would have caught #12
- No coverage for the new `PluginRoute`, `DialogMcp`, trace-viewer URL generation, config dedup, or variant cycle dialog flows

Issue #760, opened 2026-04-28
[feat] Collapse entity-per-table warehouses in schema_index (composite digest)


sahrizvi:
## Problem

When a database has many tables that share the same column structure — one ticker per table, one tenant per table, time-partitioned tables, per-region tables — the existing \`schema_index\` emits per-table schemas. N copies of the same schema fill the agent's context with redundancy. With \`schema_index\`'s truncation cap, the agent sometimes can't see all tables at all.

Concrete example we hit on a benchmark workload:

- stockmarket dataset has **2754 ticker tables**, every one with identical columns (\`Date\`, \`Open\`, \`High\`, \`Low\`, \`Close\`, \`Adj Close\`, \`CloseUSD\`).
- \`schema_index\` truncates to 200 → agent only sees ticker symbols A–Bxxx and has to discover the rest at runtime via SQL probes.
- Score on this dataset: 1/5 vs. yelp (per-table-is-different schemas) at 6/7. Direct attribution to redundant per-table schemas filling context.

## Proposed mechanism

Heuristic in \`schema_index\`'s scan logic, after enumerating tables for a connection:

1. **Group tables by column-structure fingerprint** — sorted tuple of \`(column_name, column_type)\`.
2. **Detect entity-per-table pattern** — if any single fingerprint group covers ≥50% of tables AND has ≥20 tables, mark that group as the \"entity\" group.
3. **Emit a different digest format** for the entity group:

\`\`\`
schema: <db>.<schema>
pattern: entity-per-table
table_count: 2754
composite_columns: [{name: \"Date\", type: \"VARCHAR\"}, {name: \"Open\", type: \"DOUBLE\"}, …]
sample_table: AAPL
table_names: [AAPL, AAA, AAL, AAN, …]   ← full list, searchable
\`\`\`

4. Tables NOT in the entity group still get per-table schemas. The two formats coexist.
5. \`schema_search\` should still surface table names from entity groups in lookups.

## What it enables

- ~10× context-saving on entity-per-table warehouses.
- Agent sees ALL tables (full list of names) instead of a truncated subset.
- Per-table queries still work — composite schema describes the columns; the agent uses table names directly.
- Backwards compatible: tables with unique structures still get per-table schemas as today.

## Beyond benchmark / general use cases

This is a real production pattern, common in:

- Multi-tenant SaaS (Stripe-style \"one table per customer for isolation\")
[...]

Issue #759, opened 2026-04-28
[feat] Headless mode: commit best-guess answer at max-turns instead of meta-prose


sahrizvi:
## Problem

In headless mode (`-p` / `--print`), when the agent hits `--max-turns`, the existing MAX_STEPS injection at \`packages/opencode/src/session/prompt.ts:881-888\` causes the model to emit multi-paragraph meta-prose instead of committing to an answer.

Concrete failure traces from a recent benchmark run (kimi-k2 and Claude Sonnet 4.6 both exhibit this):

> \"I have reached the maximum number of steps allowed for this task. Let me provide a summary of what has been accomplished and what remains to be done. Summary of Work Completed: 1. Successfully analyzed the question…\"

This happens even with explicit prompt instructions like \"always write SOMETHING to ANSWER before stopping\" — the MAX_STEPS injection overrides them.

In interactive mode this is fine — a human can react and ask follow-ups. In headless mode there's no human, so the program's final output is unparseable meta-prose. We measured ~5–10% of trials in a batch eval lost to this pattern.

## What it costs callers today

- Eval/benchmark frameworks (DAB, SWE-Bench, BIRD, internal evals): every team writes their own answer-extraction regex on top of altimate-code's prose, or counts those trials as fails.
- CI/CD pipelines: \`claude -p \"review this PR for security issues\"\` returns a status update instead of a partial review.
- Composable shell pipelines: \`git diff | claude -p \"summarize\" | tee summary.md\` is unreliable for long tasks.
- Predictable batch cost: trials that emit prose-instead-of-answer waste their full token budget.

## Proposed mechanism

A behaviour change scoped to **headless / `--print` mode only**:

1. Detect headless mode (CLI entry under \`packages/opencode/src/cli/\` or session config flag).
2. Branch the MAX_STEPS prompt:
   - Interactive mode → keep existing text unchanged.
   - Headless mode → use a different prompt that explicitly tells the model to commit a best-guess answer NOW.

Suggested headless prompt (refine to existing style):

\`\`\`
You have 1 turn left. Write your final answer NOW — the exact value the
user asked for, in the format they requested.

Do NOT summarize what you tried. Do NOT explain limitations. Do NOT write
meta-commentary about hitting the step limit.

Just emit the answer. If you're uncertain, emit your best guess anyway;
an uncertain answer is more useful than a meta-summary.
\`\`\`

[...]

PR #761, opened 2026-04-28
feat(sql): cross-DB join key inference via prefix/suffix overlap

### LINKED ISSUE #758: [feat] Cross-DB join key inference (prefix/suffix overlap)
## Problem

When altimate-code is connected to multiple warehouses (a real customer pattern: post-acquisition systems, multi-vendor SaaS analytics, ingest-from-many-sources), there's no built-in way to detect that columns across DBs are entity-equivalent despite naming/format mismatches. Today the agent has to discover this manually for each new pair of warehouses.

Concrete examples we hit on a benchmark workload:

- yelp dataset: `business.business_id` in MongoDB has values like `businessid_42`; `review.business_ref` in DuckDB has values like `businessref_42`. Same entity, different prefix. The agent loses many turns figuring this out.
- bookreview dataset: `books_info.book_id` (`bookid_42`) ↔ `review.purchase_id` (`purchaseid_42`).

This is the canonical \"ill-formatted join keys\" trap that the DataAgentBench paper highlights, and it shows up regularly in production warehouses (acquired-company ID prefix mismatches, dirty migrations, multi-vendor analytics).

`altimate_core.column_lineage` traces lineage *within* a query and `schema_inspect` describes *one* table — neither suggests cross-DB joins.

## Proposed mechanism

A new tool `altimate_core.detect_join_candidates(connections)` that:

1. For each pair of (db, table, string-column), pull ~50 sample values via existing \`schema_inspect\` plumbing.
2. Compute longest common prefix of L's sample values; same for R.
3. If both prefixes are non-empty AND end in \`_\` / \`-\` / \`:\` AND L's prefix ≠ R's prefix AND there's at least one matching suffix between L and R after stripping prefixes → emit a candidate.
4. Score by suffix overlap count + same-DB-type compatibility.
5. Return ranked: \`[{left_db, left_table, left_col, right_db, right_table, right_col, prefix_rule, suffix_overlap, confidence}]\`.

A working reference implementation in Python exists at \`dab-bench-internal/dab_bench/preindexer.py:_detect_join_candidates\` (~50 LOC) — straightforward to translate.

## What it enables

- Agent calls it once after a multi-warehouse connect; gets cross-DB join recommendations without writing analysis SQL.
- Foundation for `column_lineage` to become *cross-DB aware*.
- `schema_inspect` could surface a \"see also\" footer with cross-DB matches.

## Beyond benchmark / general use cases

- Acquired company's `customer_id` format ≠ parent's
- Salesforce `AccountId` ↔ Hubspot `company_id` ↔ internal warehouse `account_external_id`
- Dirty migrations leaving prefix mismatches (`v1_user_123` vs `user_123`)
- Multi-vendor analytics where each system has its own ID format

## Implementation pointers

- New file: `packages/opencode/src/altimate/tools/altimate-core-detect-join-candidates.ts`
- Existing patterns to follow: `packages/opencode/src/altimate/tools/schema-inspect.ts`, `altimate/native/altimate-core.ts`
- Tool registration: `packages/opencode/src/tool/registry.ts` (alongside other `altimate_core_*` entries)
- Estimated scope: ~200-300 LOC + tests

## Tests to add

- Prefix-overlap unit test
- Suffix-overlap unit test
- Integration test against in-memory SQLite fixture with 2 fake \"DBs\" exhibiting the `businessid_X` ↔ `businessref_X` pattern

## Evidence

Benchmark trace from \`dab-bench-internal\` showing the agent burning turns on this exact problem available on request.

sahrizvi:
Closes #758

## Motivation

When altimate-code is connected to multiple warehouses, columns that point at the same entity often have mismatched naming or formatting — e.g., one DB stores `cust_42` while another stores `42`, or one warehouse uses `businessid_X` and another uses `businessref_X`. There's currently no built-in way to surface candidate joins across DBs; the agent has to discover them per-pair, which costs tokens and is brittle.

Common situations where this lands:
- Post-acquisition system integration (acquired company's ID format ≠ parent's)
- Multi-vendor analytics (Salesforce `AccountId` ↔ Hubspot `company_id` ↔ internal `account_external_id`)
- Migrations leaving prefix differences (`v1_user_123` vs `user_123`)
- Per-source ingestion with system-specific ID schemes

## What's added

A new tool `altimate_core.detect_join_candidates` that:

1. Pulls ~50 sample values per string-typed column from each configured warehouse via the existing `connector.execute` plumbing.
2. Computes the longest common prefix per column, walking back to the last `_`/`-`/`:` separator (so partial-name matches don't count).
3. Compares prefix-stripped suffixes pairwise across DBs; emits a candidate when both columns have non-empty distinct prefixes AND share at least one suffix.
4. Returns ranked candidates: `{left_db, left_table, left_col, right_db, right_table, right_col, prefix_rule, suffix_overlap, confidence}` where `confidence = overlap / min(|left_suffixes|, |right_suffixes|)` (cheap and monotonic, deliberately not a probability).

Errors are scoped: per-table sample failures don't kill the scan; per-connection failures are returned in `connection_errors`. Identifier quoting is ANSI-safe (double quotes). The `STRING_TYPE_PATTERN` filter restricts sampling to text-like columns (varchar/text/char/string/uuid/json) — numeric/temporal columns are skipped.

## Tests

**21 new tests** in `test/altimate/tools/altimate-core-detect-join-candidates.test.ts`:

- `commonPrefix` LCP logic + separator handling (`_`/`-`/`:`) + no-separator rejection + non-string defensive paths
- `detectJoinCandidatesFromBags`: canonical pattern, same-DB rejection, identical-prefix rejection, zero-overlap rejection, no-separator rejection, ranking by overlap/confidence, full N*(N-1)/2 fan-out
- Integration: two real `bun:sqlite` `:memory:` DBs holding 10 `businessid_X` rows + 8 `businessref_X` rows (with intentional non-overlapping outliers), driven through the actual native handler with `Registry.get` stubbed
- Tool surface: title/output formatting, empty-list rendering, connection-error rendering, dispatcher-throws envelope

**Results:**
- This file: **21 pass / 0 fail**
- Sibling tests under `test/altimate/tools/`: **192 pass / 0 fail** (the 9 pre-existing failures in `sql-classify.test.ts` were verified unrelated by stashing this branch's changes — same 9 fail on `main`)
- Full opencode suite when merged with #759 + #760: **7421 pass / 0 fail / 494 skip** (~88s)
- `bun run typecheck`: clean (0 errors after `bun install` in the worktree)

The existing `altimate-core-native.test.ts` count assertion was updated (34 → 35) to reflect the new method, with `altimate_core.detect_join_candidates` added to the canonical `ALL_METHODS` list.

## Backwards compatibility

Pure addition. No existing behaviour changes; no existing tests modified beyond the count update.

## Files

- `packages/opencode/src/altimate/native/connections/detect-join-candidates.ts` (new — 317 LOC)
- `packages/opencode/src/altimate/native/connections/register.ts` (handler registration)
- `packages/opencode/src/altimate/native/types.ts` (params + result types)
- `packages/opencode/src/altimate/tools/altimate-core-detect-join-candidates.ts` (new tool wrapper)
- `packages/opencode/src/altimate/index.ts` (barrel export)
- `packages/opencode/src/tool/registry.ts` (tool registration)
- `packages/opencode/test/altimate/tools/altimate-core-detect-join-candidates.test.ts` (new — 385 LOC)
- `packages/opencode/test/altimate/altimate-core-native.test.ts` (count assertion update)

---
## Summary by cubic
Adds cross-DB join key inference that finds candidate joins by stripping distinct prefixes and matching suffixes across warehouses. Now uses dialect-aware sampling SQL, stricter permission/error handling, and safer input limits.

- **New Features**
  - Exposes `altimate_core.detect_join_candidates` and the `altimate_core_detect_join_candidates` tool; registered and exported.
  - Builds sampling SQL with dialect-aware quoting via `quoteIdentForDialect`; no inline LIMITs (caps passed to drivers).
  - Prefix logic trims to `_`/`-`/`:`; prefixes must differ; emits on suffix overlap; ranks by `suffix_overlap`, then `match_score = overlap / min(left, right)`.
  - Samples ~50 string-like values/column; scans up to 50 tables/connection; optional `schema_name`; cross-DB only; skips non-string columns.
  - Robustness: gated by `sql_execute_read` permission; parallel per-connection scans; returns `connection_errors` and bounded `partial_errors`; input caps (`connections` ≤ 16, `sample_size` ≤ 1000, `max_tables_per_connection` ≤ 500).

- **Bug Fixes**
  - Registration tests now load `connections/register` and assert against the advertised method list instead of a hard-coded count, ensuring `altimate_core.detect_join_candidates` is reliably registered.

<sup>Written for commit 65d7a9b83be74353c284d15e79524b43d913371c. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added an Altimate tool that detects and ranks cross-database join candidates, with sampling, dialect-aware SQL handling, and clear success/failure results.
* **Chores**
  * Made the detection capability available through the native bridge and tool registry for discovery and use; improved shared dialect quoting utility.
* **Tests**
  * Added extensive unit and integration tests covering detection logic, sampling, dialect quoting, permissions, and tool execution paths.


sahrizvi:
# Multi-Model Consensus Code Review — PR #761

**Title:** feat(sql): cross-DB join key inference via prefix/suffix overlap
**Branch:** `feat/cross-db-join-inference`
**Net change:** +875 / -2 across 8 files
**New code:** `detect-join-candidates.ts` (317 LOC) + `altimate-core-detect-join-candidates.ts` (125 LOC) + 385 LOC test file

---

## Verdict: **REQUEST CHANGES**

Three independent reviewers (Claude, GPT, Gemini) flagged the SQL portability problem as a blocker; GPT additionally flagged a missing permission-gating step. Kimi gave a "conditional yes" but its concerns overlap with the major issues raised elsewhere. The pure algorithm and test architecture are unanimously praised; the warehouse-facing implementation is unanimously called out for not aligning with the repo's dialect-aware patterns.

The fixes are small and bounded — likely ~30 LOC plus a couple of tests — and worth doing before merge so the tool doesn't ship as silently-broken on MySQL/SQL Server/Oracle/BigQuery.

---

## CRITICAL

### C1 — `LIMIT` + ANSI double-quote identifier quoting are not portable across the supported drivers
**Reviewers:** Claude, GPT, Gemini, Kimi (4/4 active reviewers — **Consensus**)
**Category:** Bug / Portability
**Location:** `packages/opencode/src/altimate/native/connections/detect-join-candidates.ts:175–177` (`quoteIdent`), lines 190–194 (`fetchColumnSamples`)
**Why it matters:** The handler builds:
```ts
const sql = `SELECT ${col} FROM ${target} WHERE ${col} IS NOT NULL LIMIT ${sampleSize}`
```
Both ingredients break on warehouses this repo ships drivers for:
- **`LIMIT N`** is invalid on **SQL Server** (uses `TOP`/`OFFSET ... FETCH NEXT`) and pre-12c **Oracle** (uses `ROWNUM`/`FETCH FIRST`).
- **Double-quoted identifiers** are parsed as **string literals** by **MySQL/MariaDB** in default mode (without `ANSI_QUOTES`); **Databricks/Spark** and **BigQuery** use backticks; **SQL Server** uses brackets.

Combined with the bare `catch {}` in `fetchColumnSamples` (line 201), the failure mode is silent: the user gets `success: true, candidates: [], bags_scanned: 0` and no signal that every sample query threw.

**Suggested fix:** Drop the inline `LIMIT` and let the existing driver contract apply row-capping (`Connector.execute(sql, sampleSize)` already does this — see `packages/drivers/src/sqlite.ts:25–47`). For identifier quoting, either delegate to a dialect-aware helper (the repo has dialect-aware quoting in `data-diff`) or branch on `Registry.getConfig(name)?.type`. GPT supplied a concrete patch:

```ts
const warehouseType = Registry.getConfig(name)?.type ?? "unknown"
const quotedTarget = [schema, table].filter(Boolean)
  .map(p => quoteIdentForDialect(p, warehouseTypeToDialect(warehouseType)))
  .join(".")
const quotedCol = quoteIdentForDialect(column, warehouseTypeToDialect(warehouseType))
const sql = `SELECT ${quotedCol} FROM ${quotedTarget} WHERE ${quotedCol} IS NOT NULL`
const result = await connector.execute(sql, sampleSize)
```

---

## MAJOR

### M1 — No permission gating on warehouse reads (policy bypass)
**Reviewers:** GPT (Unique, but high-impact)
**Category:** Security / Design
**Location:** `packages/opencode/src/altimate/tools/altimate-core-detect-join-candidates.ts:51`
**Why it matters:** This tool issues many `SELECT` reads across every string column of every scanned table on every configured connection — but it never goes through the `sql_execute_read` permission flow that `data_diff` uses for less-invasive operations. Even though raw values aren't echoed back to the user, they are pulled from the warehouse and held in memory. This is a policy-consistency gap that the security review will catch downstream if not now.
**Suggested fix:** Gate execution with `ctx.ask({ permission: "sql_execute_read", ... })` before the dispatcher call, and add a tool test asserting the permission request occurs.

### M2 — `safeListSchemas` silent fallback to `["public"]` masks real failures
**Reviewers:** Claude, GPT, Gemini (3/4 — **Consensus**)
**Category:** Logic Error / Observability
**Location:** `detect-join-candidates.ts:256–262`
**Why it matters:** If `listSchemas()` fails, the helper returns `["public"]`. That is wrong for Snowflake (`PUBLIC` uppercase, per-database scoping), wrong for BigQuery (no `public` concept), wrong for Oracle, ClickHouse, MySQL. The user gets an empty result with no entry in `connection_errors` — the silent-failure mode again.
**Suggested fix:** If `listSchemas()` fails and the caller didn't pass `schema_name`, record the connection-level error and skip the connection:
```ts
let schemas: string[]
try {
  schemas = params.schema_name ? [params.schema_name] : await connector.listSchemas()
} catch (e) {
  errors[name] = `Failed to list schemas: ${String(e)}`
  continue
}
```

### M3 — Per-table sample failures swallowed without any reporting
**Reviewers:** Claude, GPT (2/4 — **Consensus**, also implicit in Kimi's commentary)
**Category:** Design / Observability
**Location:** `detect-join-candidates.ts:201–203` (`fetchColumnSamples`), lines 256–286 (other `safe*` helpers)
**Why it matters:** The PR description claims "per-table failures don't kill the scan; per-connection failures are returned in `connection_errors`" — but there is no surface for per-table or per-column errors. A warehouse user who can list tables but lacks `SELECT` on one of them gets a silently-skipped table. Combined with C1+M2, the user has no signal that anything went wrong.
**Suggested fix:** Add a counter (or a small bounded list) of swept errors to the result envelope, even if just `swept_errors: number`.

### M4 — Tool wrapper mishandles non-throwing native failures
**Reviewers:** GPT (Unique)
**Category:** Bug
**Location:** `altimate-core-detect-join-candidates.ts:59–75`
**Why it matters:** If the dispatcher returns `{ success: false, error: ... }` (which happens when the connection-count guard fires at line 295–301 of the native handler), the tool wrapper still renders `Join candidates: 0 found` / `No cross-DB join candidates detected.` That presents an operational failure as a clean negative result. Existing tests cover the *thrown-error* path but not the `success: false` envelope.
**Suggested fix:**
```ts
if (!result.success || error) {
  return {
    title: "Join candidates: FAILED",
    metadata: { success: false, candidate_count: 0, ...(error && { error }) },
    output: `Failed to detect join candidates: ${error ?? "Unknown error"}`,
  }
}
```

### M5 — `confidence` is a misleading field name
**Reviewers:** Claude (Unique)
**Category:** Design / Documentation
**Location:** `detect-join-candidates.ts:88–93`, line 124
**Why it matters:** `confidence = overlap / min(|lsuf|, |rsuf|)`. With `sample_size: 50`, two columns whose 50 sampled values happen to share a suffix yield `confidence = 1.0` — even if the underlying tables are mostly disjoint. The doc-comment notes "not a probability" but the *field name* in the JSON output is `confidence`, which an LLM consumer will read as "very likely correct."
**Suggested fix:** Rename to `sample_overlap_ratio` (consistent with `suffix_overlap` naming) and propagate to the formatter.

### M6 — Sample collection is fully serial across connections, schemas, tables, columns
**Reviewers:** Gemini (Unique)
**Category:** Performance
**Location:** `detect-join-candidates.ts:220–251`
**Why it matters:** Every `await` in `collectSampleBags` is sequential. With defaults (50 tables × N string columns × M connections), total latency = sum of all per-column queries. For a real two-warehouse scan, this is many seconds to minutes.
**Suggested fix:** Parallelize at the connection level with `Promise.all`; consider chunked parallelism within a connection (e.g., `p-limit`-style) to avoid hammering a single warehouse.

### M7 — No bound on column count or sample-size, allowing memory/time blowups
**Reviewers:** Kimi (Unique, valid)
**Category:** Code Quality / Performance
**Location:** `altimate-core-detect-join-candidates.ts:38–49` (Zod schema)
**Why it matters:** `sample_size` and `max_tables_per_connection` accept any positive integer. `sample_size: 1000000` ⇒ allocate sets of 1M strings × N columns × M connections. The `Promise.all` parallelization (M6) makes this worse.
**Suggested fix:**
```ts
sample_size: z.number().int().positive().max(1000).optional()
max_tables_per_connection: z.number().int().positive().max(500).optional()
```

---

## MINOR

### MI1 — `register.ts:634` registers handler with implicit `any` params
**Reviewers:** Claude (Unique)
**Location:** `register.ts:634`
Every other `register()` call in the file types its params. Add the import + type annotation:
```ts
register("altimate_core.detect_join_candidates",
  async (params: AltimateCoreDetectJoinCandidatesParams) => detectJoinCandidates(params))
```

### MI2 — `STRING_TYPE_PATTERN` may miss/include real types
**Reviewers:** Claude, Kimi
**Location:** `detect-join-candidates.ts:163`
ClickHouse `FixedString(8)` is not anchored against `string` (the `^` anchor will fail). Snowflake `STRING` covered. Postgres `xml` not in pattern (good — it's not text-y in this sense). Worth a unit test for the type filter to pin behavior.

### MI3 — No exclusion of obviously-non-key string columns
**Reviewers:** Kimi (Unique)
**Location:** `detect-join-candidates.ts:234–244`
Wide tables expose `description`, `notes`, `email`, `address` etc. — all string-typed but not realistic join keys. Suggested skip pattern:
```ts
const SKIP_COLUMN_PATTERNS = /^(description|notes?|comments?|email|phone|address|created_at|updated_at|deleted_at)$/i
```

### MI4 — No deduplication of overlapping candidates within a connection pair
**Reviewers:** Kimi (Unique)
**Location:** `detect-join-candidates.ts:100–145`
If both DBs have `id1` and `id2` columns, you get N×M candidates with identical or near-identical suffix sets. Either dedupe by suffix-set hash or document the behavior.

### MI5 — `table` field encodes `schema.table` as a single string
**Reviewers:** Kimi (Unique)
**Location:** `detect-join-candidates.ts:244`
`bag.table = `${schema}.${t.name}`` forces consumers to parse. Add an explicit `schema` field to `ColumnSampleBag` and `JoinCandidate`. Low-impact but improves the public type contract.

### MI6 — `formatCandidates` may produce thousands of lines on real workloads
**Reviewers:** Gemini (Unique)
**Location:** `altimate-core-detect-join-candidates.ts:105–115`
Truncate human-readable output to top ~20–30 candidates; keep the full count in `metadata.candidate_count`.

### MI7 — Test stub references `ALTIMATE_TELEMETRY_DISABLED` but handler doesn't emit telemetry
**Reviewers:** Claude (Unique)
**Location:** test file lines 30–36
The `beforeEach` sets `ALTIMATE_TELEMETRY_DISABLED=true` but `detectJoinCandidates` doesn't call `Telemetry.track`. Either wire telemetry in (precedent: `register.ts:393–406` for `sql.execute`) or remove the dead env-var setup.

### MI8 — `tablesScanned` cap doesn't surface truncation
**Reviewers:** Claude (Unique)
**Location:** `detect-join-candidates.ts:226–247`
When `max_tables_per_connection` fires mid-scan, the response has no field telling the caller "we hit the cap." Add `truncated: boolean` or `tables_skipped: number` to `bags_scanned`-adjacent metadata.

---

## NIT

- **N1** (Claude, Kimi): The doc comment at `detect-join-candidates.ts:17` references "dab_bench's preindexer" — internal codename leaks into a public-looking source file. Either inline the algorithm summary or remove the breadcrumb.
- **N2** (Claude): Unicode `↔` in the formatter output may render badly in some CI logs. Consider `<->`.
- **N3** (Kimi): Double-wrapped error handling — both the native handler and the tool wrapper catch and stringify exceptions; can produce nested `Failed: ...` prefixes.

---

## Positive observations (Consensus)

- **Pure-vs-IO split is excellent.** `commonPrefix` / `stripPrefixSet` / `detectJoinCandidatesFromBags` are side-effect-free and exported. Praised by every reviewer.
- **Test architecture is genuinely good.** Three layers (pure unit, SQLite-backed integration, tool surface) plus dispatcher-throws path. ~21 tests for ~317 LOC of new logic.
- **`altimate_change` markers** correctly applied in every modified upstream-shared file.
- **Zod validation** at the tool boundary plus a defensive re-check in the native handler — belt-and-braces, appropriate for an LLM-callable surface.
- **Type contract** is registered in `BridgeMethods` and reuses the generic `AltimateCoreResult` envelope. No schema sprawl.
- **Graceful degradation** on per-connection failures (when they actually surface) is the right shape for a multi-warehouse tool.

---

## Missing tests / edge cases (Consensus)

1. **Non-ANSI dialect path.** No test issues SQL the way an MSSQL/MySQL/Oracle driver would receive it. Add at least one test that proves the SQL is portable, or accept the C1 fix and prove driver-side limit handling works.
2. **`max_tables_per_connection` cap actually applied.** The cap can be removed without any test failing.
3. **`schema_name` filter.** Branch at lines 223–225 is uncovered.
4. **Permission-gating** (if M1 is accepted).
5. **`success: false` envelope** at the tool layer (M4).
6. **String-type filter.** Insert an `INTEGER amount` column (the SQLite fixture already has one at line 239) and assert it's not in the bags.
7. **Empty-suffix / whitespace-only values** (Kimi).
8. **Case sensitivity** — e.g. `BusinessID_` vs `businessid_` (Kimi).

---

## Attribution

| Finding | Severity | Reviewers |
|---|---|---|
| C1 SQL portability (LIMIT + quoting) | CRITICAL | Claude, GPT, Gemini, Kimi |
| M1 No permission gating | MAJOR | GPT |
| M2 `["public"]` schema fallback | MAJOR | Claude, GPT, Gemini |
| M3 Silent per-table errors | MAJOR | Claude, GPT |
| M4 `success: false` mishandled | MAJOR | GPT |
| M5 `confidence` misleading | MAJOR | Claude |
| M6 Serial sample collection | MAJOR | Gemini |
| M7 No upper bounds on params | MAJOR | Kimi |
| MI1 Implicit `any` register | MINOR | Claude |
| MI2 String-type pattern coverage | MINOR | Claude, Kimi |
| MI3 No column exclusion list | MINOR | Kimi |
| MI4 No candidate dedup | MINOR | Kimi |
| MI5 schema/table compound key | MINOR | Kimi |
| MI6 Output truncation | MINOR | Gemini |
| MI7 Dead telemetry env-var | MINOR | Claude |
| MI8 Cap-truncation not signaled | MINOR | Claude |
| N1 dab_bench codename leak | NIT | Claude, Kimi |
| N2 Unicode separator | NIT | Claude |
| N3 Double-wrapped errors | NIT | Kimi |

---

## Disagreements

- **Kimi**'s overall verdict was "conditional yes — fix #1, #4, #8 before merge" while **Claude / GPT / Gemini** lean "request changes" (with C1 as blocker). The disagreement reduces to whether SQL portability is a release blocker vs. a follow-up: GPT and Claude argue it ships visibly broken on multiple supported warehouses (silent zero-result), so blocker. Kimi assumed dialect parity would be sorted post-merge.
- **Performance** framing differs: Gemini emphasizes the serial-I/O latency (M6); Kimi emphasizes algorithmic worst-case in `commonPrefix` (long values); Claude emphasizes the unbounded bag count (M7-related). All are valid; the bounded-fix is M7's Zod cap, the latency fix is M6.
- **`commonPrefix` long-string defensiveness** (Kimi suggested a 100-char early-exit) vs. **leave as-is** (Claude's view): the values are sample-bounded by `sample_size <= 50` by default, so the worst case is small. Apply only if M7 doesn't add the upper bound.

---

## Footer

Reviewed by 8 participants: Claude + GPT 5.4 Codex + Gemini 3.1 Pro + Kimi K2.5 + MiniMax M2.7 + GLM-5.1 + Qwen 3.6 + MiMo V2 Pro.

**Active responses: 4/8** — Claude, GPT (full), Gemini (partial — Gemini API quota exhaustion mid-run, but produced a coherent review), Kimi (full).

**Failed/timed-out: 4/8** — GLM-5.1, Qwen 3.6, MiniMax M2.7, MiMo V2 Pro all blocked by a kilo SQLite database lock (a separate consensus run on PR #762 held the DB lock; running these models sequentially still hit the contention plus per-process startup races). The synthesis above represents the consensus of the 4 active reviewers.

This synthesis is single-pass (no convergence rounds) per the user's instruction.

sahrizvi:
## Update: addressed-vs-pending from the [consensus review](https://github.com/AltimateAI/altimate-code/pull/761#issuecomment-4335155823)

Pushed `80758f5c2` on top of the original branch.

### Addressed (this commit)

|---|---|---|---|
| C1 | CRITICAL | SQL portability — `LIMIT` + ANSI quoting break MySQL/SQL Server/Oracle/Databricks/BigQuery | Reuse exported `quoteIdentForDialect` from `data-diff.ts` (backticks/brackets/ANSI per dialect); drop inlined `LIMIT N`; delegate row cap to `connector.execute(sql, sampleSize)` so each driver applies its native limit syntax. Extracted `buildSampleSql` as a pure helper. |
| M1 | MAJOR | Permission gating bypass | Tool wrapper now calls `ctx.ask({permission:"sql_execute_read", ...})` before any SELECT, mirroring `data_diff` / `sql_execute`. |
| M2 | MAJOR | `["public"]` schema fallback is unsafe | Removed `safeListSchemas → ["public"]` fallback. `listSchemas` failure now surfaces as a connection-level error and the connection is skipped. |
| M3 | MAJOR | Silent per-table errors | Per-table / per-column sampling failures collected in bounded `partial_errors: Record<string, string[]>` (cap 20/connection); surfaced in tool output and metadata. |
| M4 | MAJOR | `success: false` mishandling | Tool wrapper returns `title: "Join candidates: FAILED"` envelope when dispatcher returns `{success:false, error}`. |
| M5 | MAJOR | Misleading `confidence` field name | Renamed to `match_score`, JSDoc updated to call out it's a heuristic ranking signal, not a probability. Note this is a pre-merge rename — no aliased back-compat field. |
| M6 | MAJOR | Serial I/O across connections | `collectSampleBags` parallelizes at the connection level via `Promise.all`. Per-connection cap still bounds blast radius. |
| M7 | MAJOR | No Zod upper bounds | Added `.max()` caps: `connections ≤ 16`, `sample_size ≤ 1000`, `max_tables_per_connection ≤ 500`. |
| MI1 | MINOR | (drive-by) Untyped dispatcher registration | Typed with `AltimateCoreDetectJoinCandidatesParams` / `AltimateCoreResult`. |

**Tests:** 35 pass / 0 fail in the targeted file (was 21; +14 new for dialect SQL emission across MySQL/T-SQL/Fabric/Postgres/Snowflake/generic, no-`LIMIT` inlining, delimiter escaping, permission gating, `success: false` envelope, `partial_errors` collection, no-`public` fallback, Zod upper-bound rejection). Sibling regression: 2,886 pass. Typecheck clean.

### Deferred (per consensus tags)

All MINOR / NIT findings — MI2 (string-type pattern coverage), MI3 (column-name skip list), MI4 (candidate dedup), MI5 (compound `schema.table` field), MI6 (output truncation), MI7 (dead telemetry env-var), MI8 (cap-truncation signal), N1 (codename leak), N2 (unicode), N3 (double-wrap). Also deferred CodeRabbit's `index.ts` barrel cleanup since it wasn't in the consensus list.

Demoted as invalid: none from this PR (no Kimi-style misreads).

### Heads-up

`quoteIdentForDialect` is now exported from `data-diff.ts`. Inside the package it's only consumed by `detect-join-candidates.ts` today. Tool API changes from this round:

- `confidence` → `match_score` (rename, not aliased — pre-merge so no installed users yet)
- New `partial_errors` field on the result envelope
- First invocation triggers a `sql_execute_read` permission prompt; non-interactive harnesses must auto-approve

dev-punia-altimate:
## ✅ Tests — All Passed

### TypeScript — passed


cc @sahrizvi

dev-punia-altimate:

- connection_refused [2.58ms]
- timeout [2.50ms]
- permission_denied [2.93ms]
- parse_error [2.29ms]
- oom [2.46ms]
- network_error [2.31ms]
- auth_failure [2.50ms]
- rate_limit [2.59ms]
- internal_error [2.98ms]
- empty_error [0.25ms]
- connection_refused [0.13ms]
- timeout [0.07ms]
- permission_denied [0.07ms]
- parse_error [0.07ms]
- oom [0.07ms]

### Next Step

cc @sahrizvi

PR #762, opened 2026-04-28
feat(schema): collapse entity-per-table warehouses in schema_index

### LINKED ISSUE #760: [feat] Collapse entity-per-table warehouses in schema_index (composite digest)

sahrizvi:
Closes #760

## Motivation

When a database has many tables that share the same column structure — one ticker per table, one tenant per table, time-partitioned tables, per-region tables — `schema_index` today emits redundant per-table schemas. The agent's context fills with N copies of the same shape, and `schema_index`'s truncation cap means the agent often can't see all tables at all.

This is a real production pattern across:
- Multi-tenant SaaS (one table per customer for isolation)
- Time-partitioned analytics (`events_2024_01_01`, `events_2024_01_02`, …)
- Per-region tables (`orders_us`, `orders_eu`, `orders_apac`)
- Per-environment schemas (dev/staging/prod with same shape)
- Per-source ingestion (`events_segment`, `events_amplitude`, …)

For warehouses that exhibit it, current behaviour wastes context on duplicate schemas and silently hides tail tables beyond the truncation cap.

## What's added

A heuristic in `schema_index`'s scan logic that detects the entity-per-table pattern at scan time and emits a composite digest — one shared schema description plus the full list of table names — instead of N redundant schemas. Tables that don't fit the dominant fingerprint still get per-table schemas as before. `schema_search` is updated to surface entity-group matches.

**Detection rule:** group tables by sorted `(column_name, column_type)` fingerprint. If any one fingerprint covers ≥50% of tables AND ≥20 tables, treat it as the entity group and emit a composite digest. Tables with empty column lists are ignored. Both thresholds are overridable per-call via `indexWarehouse` options.

**Composite digest format:**

```
schema: public
pattern: entity-per-table
table_count: 30
composite_columns: [{name: "Close", type: "DOUBLE"}, {name: "Date", type: "VARCHAR"}, ...]
sample_table: TICKER_0
table_names: [TICKER_0, TICKER_1, TICKER_10, TICKER_11, ...]
```

`schema_search` queries entity-group rows in addition to per-table rows; a query for a specific table name returns just that row from the group, not the entire member list. The tool-level formatter renders matches under a `## Entity-per-table Groups` section.

`cacheStatus().total_tables` and `total_columns` were updated to fold collapsed counts in, so existing callers see real warehouse counts regardless of digest format. This also fixed a pre-existing scenario where a 1000-table stress test under-reported counts when groups collapsed.

## Tests

**19 new unit tests** in `test/altimate/entity-groups.test.ts`:
- 100 same-shape tables → marked as entity
- 5 mixed-shape tables → not marked
- 50 same-shape + 50 different-shape → not marked (under 50% threshold)
- Empty-column tables ignored for fingerprinting
- Threshold and min-count overrides honoured per-call

**7 new integration tests** in `test/altimate/schema-cache.test.ts`:
- In-memory SQLite fixture with 30 identical-schema tables produces composite digest
- Pre-existing 5-mixed-table fixture produces identical results to previous behaviour (backwards compat)
- 1000-table stress test passes (regression fix)

**Results:**
- Affected files (`entity-groups.test.ts`, `schema-cache.test.ts`, `schema-finops-dbt.test.ts`, `schema-resolver.test.ts`): **139 pass / 0 fail**
- Telemetry suite: 46/0 pass; `real-tool-simulation`: 40/0 pass — no regressions
- Full opencode suite when merged with #758 + #759: **7421 pass / 0 fail / 494 skip** (~88s)
- `bun run typecheck`: clean (0 errors after `bun install` in the worktree)

## Backwards compatibility

- When no fingerprint group qualifies (the common case), `entity_groups` is `undefined` and per-table rows are written exactly as before. Verified by integration test against the existing 5-table mixed fixture.
- The new `entity_groups` field on `SchemaIndexResult` and `SchemaSearchResult` is optional; older consumers see no breaking change.

## Files

- `packages/opencode/src/altimate/native/schema/cache.ts` (DDL + scan + search + status)
- `packages/opencode/src/altimate/native/schema/entity-groups.ts` (new — pure detector, 174 LOC)
- `packages/opencode/src/altimate/native/types.ts` (new optional fields + interfaces)
- `packages/opencode/src/altimate/tools/schema-index.ts` (output formatter)
- `packages/opencode/src/altimate/tools/schema-search.ts` (output formatter)
- `packages/opencode/test/altimate/entity-groups.test.ts` (new — 269 LOC)
- `packages/opencode/test/altimate/schema-cache.test.ts` (added integration tests)

---
## Summary by cubic
Collapses “entity-per-table” warehouses in `schema_index` by emitting a single composite digest per schema, cutting redundant context and keeping search accurate. Adds safe search/output and column reconstruction so downstream tools still see all tables; addresses #760.

- **New Features**
  - Detects entity-per-table via column-shape fingerprinting (≥50% and ≥20 tables by default); overrides via `schema_index` params `entityRatioThreshold` and `entityMinTables`.
  - Stores qualifying groups once with shared columns and full `table_names`; non-group tables remain per-table. At most one group is collapsed per schema.
  - `schema_search` shows entity-group matches and returns only the queried table(s) from a group. Tool output lists groups and truncates huge `table_names` (head + tail).

- **Bug Fixes**
  - `listColumns` reconstructs columns for collapsed groups so PII and SQL validation see complete schemas; `cacheStatus` now includes collapsed groups in `total_tables` and `total_columns`.
  - Reduced false positives and fingerprint edge cases with bounded group `search_text`, table-name matching via post-filter, and safe fingerprinting for `:`/`|`/missing types.

<sup>Written for commit dc2af3c0aa80534b7742b0f0524f511c83a09cb0. Summary will update on new commits. <a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/762?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **New Features**
  * Schema indexing now detects and collapses groups of tables with identical column structures and surfaces them in index/search results.
  * Search and reporting include entity-group matches, counts, and summarized composite column info.
  * Detection thresholds (ratio and minimum tables) are configurable.

* **Behavior Changes**
  * Aggregate table/column counts and column listings now include contributions from collapsed entity groups.

* **Tests**
  * New unit and integration tests covering detection, indexing, search, listing, and threshold validation.

### DIFF
--- a/packages/opencode/src/altimate/native/schema/cache.ts
+++ b/packages/opencode/src/altimate/native/schema/cache.ts
@@ -18,7 +18,15 @@ import type {
   SchemaCacheWarehouseStatus,
   SchemaSearchTableResult,
   SchemaSearchColumnResult,
+  SchemaEntityGroupSummary,
+  SchemaSearchEntityGroupResult,
 } from "../types"
+import {
+  detectEntityGroup,
+  type TableShape,
+  DEFAULT_ENTITY_RATIO_THRESHOLD,
+  DEFAULT_ENTITY_MIN_TABLES,
+} from "./entity-groups"
 
 // ---------------------------------------------------------------------------
 // DDL
@@ -67,6 +75,24 @@ CREATE INDEX IF NOT EXISTS idx_columns_search ON columns_cache(search_text);
 CREATE INDEX IF NOT EXISTS idx_tables_warehouse ON tables_cache(warehouse);
 CREATE INDEX IF NOT EXISTS idx_columns_warehouse ON columns_cache(warehouse);
 CREATE INDEX IF NOT EXISTS idx_columns_table ON columns_cache(warehouse, schema_name, table_name, column_name);
+
+CREATE TABLE IF NOT EXISTS entity_groups_cache (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    warehouse TEXT NOT NULL,
+    database_name TEXT,
+    schema_name TEXT NOT NULL,
+    pattern TEXT NOT NULL DEFAULT 'entity-per-table',
+    fingerprint TEXT NOT NULL,
+    table_count INTEGER NOT NULL,
+    sample_table TEXT,
+    composite_columns_json TEXT NOT NULL,
+    table_names_json TEXT NOT NULL,
+    search_text TEXT NOT NULL,
+    UNIQUE(warehouse, database_name, schema_name, fingerprint)
+);
+
+CREATE INDEX IF NOT EXISTS idx_entity_groups_search ON entity_groups_cache(search_text);
+CREATE INDEX IF NOT EXISTS idx_entity_groups_warehouse ON entity_groups_cache(warehouse);
 `
 
 // ---------------------------------------------------------------------------
@@ -141,22 +167,36 @@ export class SchemaCache {
 
   /**
    * Crawl a warehouse and index all schemas/tables/columns.
+   *
+   * For each schema, runs entity-per-table detection: if ≥50% of tables share
+   * the same column structure and the group has at least 20 tables, emits a
+   * single composite digest row in `entity_groups_cache` instead of N
+   * near-identical per-table rows. Tables outside the group still get their
+   * normal per-table entries.
    */
   async indexWarehouse(
     warehouseName: string,
     warehouseType: string,
     connector: Connector,
+    options: {
+      entityRatioThreshold?: number
+      entityMinTables?: number
+    } = {},
   ): Promise<SchemaIndexResult> {
     const now = new Date().toISOString()
+    const ratioThreshold = options.entityRatioThreshold ?? DEFAULT_ENTITY_RATIO_THRESHOLD
+    const minTables = options.entityMinTables ?? DEFAULT_ENTITY_MIN_TABLES
 
     // Clear existing data
     this.db.prepare("DELETE FROM columns_cache WHERE warehouse = ?").run(warehouseName)
     this.db.prepare("DELETE FROM tables_cache WHERE warehouse = ?").run(warehouseName)
+    this.db.prepare("DELETE FROM entity_groups_cache WHERE warehouse = ?").run(warehouseName)
 
     let totalSchemas = 0
     let totalTables = 0
     let totalColumns = 0
     const databaseName: string | null = null
+    const entityGroupsEmitted: SchemaEntityGroupSummary[] = []
 
     let schemas: string[] = []
     try {
@@ -175,6 +215,13 @@ export class SchemaCache {
       `INSERT OR REPLACE INTO columns_cache
        (warehouse, database_name, schema_name, table_name, column_name, data_type, nullable, search_text)
        VALUES (?, ?, ?, ?, ?, ?, ?, ?)`,
+    )
+
+    const insertEntityGroup = this.db.prepare(
+      `INSERT OR REPLACE INTO entity_groups_cache
+       (warehouse, database_name, schema_name, pattern, fingerprint, table_count,
+        sample_table, composite_columns_json, table_names_json, search_text)
+       VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
     )
 
     // Batch inserts per-table inside a transaction to avoid per-statement disk fsyncs.
@@ -200,35 +247,100 @@ export class SchemaCache {
         continue
       }
 
+      // Collect every table's columns first so the entity-group detector
+      // can see the whole schema at once. We also retain the table_type and
+      // raw column rows for downstream insert.
+      type TableSnapshot = {
+        name: string
+        type: string
+        columns: Array<{ name: string; data_type: string; nullable: boolean }>
+      }
+      const snapshots: TableSnapshot[] = []
       for (const tableInfo of tables) {
-        totalTables++
-        const searchText = makeSearchText(databaseName, schemaName, tableInfo.name, tableInfo.type)
-
         let columns: Array<{ name: string; data_type: string; nullable: boolean }> = []
         try {
           columns = await connector.describeTable(schemaName, tableInfo.name)
         } catch {
           // continue with empty columns
         }
+        snapshots.push({ name: tableInfo.name, type: tableInfo.type, columns })
+      }
 
-        // Build column insert args
+      const shapes: TableShape[] = snapshots.map((s) => ({
+        table_name: s.name,
+        columns: s.columns.map((c) => ({ name: c.name, data_type: c.data_type })),
+      }))
+      const detection = detectEntityGroup(shapes, { ratioThreshold, minTables })
+
+      const inGroup = new Set<string>(
+        detection.entity_group ? detection.entity_group.table_names : [],
+      )
+
+      for (const snap of snapshots) {
+        totalTables++
+
+        if (inGroup.has(snap.name)) {
+          // Tables inside the entity group are NOT emitted per-table.
+          // Their columns are still counted toward totalColumns so the index
+          // result accurately reflects the warehouse, but they live inside
+          // the composite digest (and the entity_groups_cache row below).
+          totalColumns += snap.columns.length
+          continue
+        }
+
+        const searchText = makeSearchText(databaseName, schemaName, snap.name, snap.type)
         const columnArgsBatch: any[][] = []
-        for (const col of columns) {
+        for (const col of snap.columns) {
           totalColumns++
           const colSearch = makeSearchText(
-            databaseName, schemaName, tableInfo.name, col.name, col.data_type,
+            databaseName, schemaName, snap.name, col.name, col.data_type,
           )
           columnArgsBatch.push([
-            warehouseName, databaseName, schemaName, tableInfo.name,
+            warehouseName, databaseName, schemaName, snap.name,
             col.name, col.data_type, col.nullable ? 1 : 0, colSearch,
           ])
         }
 
-        // Insert table + all its columns in a single transaction
         insertTableBatch(
-          [warehouseName, databaseName, schemaName, tableInfo.name, tableInfo.type, searchText],
+          [warehouseName, databaseName, schemaName, snap.name, snap.type, searchText],
           columnArgsBatch,
         )
+      }
+
+      // Persist the entity group for this schema, if one was detected.
+      if (detection.entity_group) {
+        const eg = detection.entity_group
+        const compositeColumns = eg.composite_columns
+        const groupSearchText = makeSearchText(
+          databaseName,
+          schemaName,
+          eg.sample_table,
+          ...eg.table_names,
+          ...compositeColumns.map((c) => c.name),
+          ...compositeColumns.map((c) => c.data_type),
+        )
+        insertEntityGroup.run(
+          warehouseName,
+          databaseName,
+          schemaName,
+          "entity-per-table",
+          eg.fingerprint,
+          eg.table_names.length,
+          eg.sample_table,
+          JSON.stringify(compositeColumns),
+          JSON.stringify(eg.table_names),
+          groupSearchText,
+        )
+        entityGroupsEmitted.push({
+          warehouse: warehouseName,
+          database: databaseName ?? undefined,
+          schema_name: schemaName,
+          pattern: "entity-per-table",
+          table_count: eg.table_names.length,
+          composite_columns: compositeColumns,
+          sample_table: eg.sample_table,
+          table_names: eg.table_names,
+        })
       }
     }
 
@@ -249,6 +361,7 @@ export class SchemaCache {
       tables_indexed: totalTables,
       columns_indexed: totalColumns,
       timestamp: now,
+      entity_groups: entityGroupsEmitted.length > 0 ? entityGroupsEmitted : undefined,
     }
   }
 
@@ -317,11 +430,54 @@ export class SchemaCache {
       }
     })
 
+    // Search entity groups: a group matches if any of its tokens appears in
+    // its search_text (which includes the full table_names list, composite
+    // column names/types, schema, and sample table). This means the agent
+    // can still find a specific collapsed table by name.
+    const groupRows = this.db.prepare(
+      `SELECT warehouse, database_name, schema_name, pattern, table_count,
+              sample_table, composite_columns_json, table_names_json
+       FROM entity_groups_cache
+       WHERE ${searchCondition} ${whFilter}
+       ORDER BY schema_name, sample_table
+       LIMIT ?`,
+    ).all(...searchParams, ...whParams, limit) as any[]
+
+    const entityGroups: SchemaSearchEntityGroupResult[] = groupRows.map((row) => {
+      const compositeColumns = JSON.parse(row.composite_columns_json) as {
+        name: string
+        data_type: string
+      }[]
+      const tableNames = JSON.parse(row.table_names_json) as string[]
+      // Of the group's table names, return the subset that actually matches
+      // any query token. This makes "find me table AAPL" return just AAPL,
+      // not all 2754 tickers.
+      const lowerTokens = tokens.map((t) => t.toLowerCase())
+      const matching = tableNames.filter((name) => {
+        const lowerName = name.toLowerCase()
+        return lowerTokens.some((tok) => lowerName.includes(tok))
+      })
+      return {
+        warehouse: row.warehouse,
+        database: row.database_name ?? undefined,
+        schema_name: row.schema_name,
+        pattern: row.pattern,
+        table_count: row.table_count,
+        composite_columns: compositeColumns,
+        sample_table: row.sample_table,
+        // If query matched on schema or composite columns rather than a
+        // specific table, return an empty matching_tables list so callers
+        // know it was a structural match.
+        matching_tables: matching,
+      }
+    })
+
     return {
       tables,
       columns,
       query,
-      match_count: tables.length + columns.length,
+      match_count: tables.length + columns.length + entityGroups.length,
+      entity_groups: entityGroups.length > 0 ? entityGroups : undefined,
     }
   }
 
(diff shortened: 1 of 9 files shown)


sahrizvi:
# Multi-model consensus review — PR #762

> 8 reviewers ran independently against the worktree. 6 produced full reviews; Gemini was blocked by workspace restrictions and Qwen timed out mid-investigation. The remaining 6 reviewers reach a strong consensus that the detector itself is well-built but the integration has at least one production-impacting bug that must land before merge.

## Verdict: **REQUEST CHANGES**

The detector module (`entity-groups.ts`) is a clean, well-tested pure function and represents the strongest part of this PR. The integration layer in `cache.ts` and the tool surface have several real defects. The most serious is a silent regression in PII detection that will produce false-clean reports on warehouses with collapsed tables.

---

## Critical

### C1. `listColumns()` skips columns from collapsed tables — silently breaks PII detection (and SQL pre-validation)
- **Category:** Bug / Data Loss
- **Location:** `packages/opencode/src/altimate/native/schema/cache.ts:548-573`; downstream consumers `pii-detector.ts:50-92`, `tools/sql-execute.ts:171-176`
- **Description:** Tables in an entity group are intentionally **not** inserted into `columns_cache` (line 287 `continue`s). `listColumns()` only queries `columns_cache`. Every consumer that enumerates columns therefore goes blind on collapsed warehouses:
  - `schema_detect_pii` returns 0 findings on a fully-collapsed warehouse — silent false-clean.
  - `sql_execute` pre-validation falls back to `empty_cache` and skips schema validation entirely.
  - Any column-level downstream tool inherits the same blind spot.
  
  This is a behavior change for existing safety features that the PR did not flag.

- **Suggestion:** Have `listColumns()` union in synthetic column rows reconstructed from `entity_groups_cache.composite_columns_json` × `table_names_json`. Sample suggestion:

```ts
listColumns(warehouse: string, limit: number = 10000) {
  const perTable = ... /* existing query */
  const groupRows = this.db.prepare(
    `SELECT schema_name, database_name, composite_columns_json, table_names_json
     FROM entity_groups_cache WHERE warehouse = ?`
  ).all(warehouse) as any[]
  for (const gr of groupRows) {
    const cols   = JSON.parse(gr.composite_columns_json) as {name:string; data_type:string}[]
    const tables = JSON.parse(gr.table_names_json) as string[]
    for (const t of tables) {
      for (const c of cols) {
        if (perTable.length >= limit) break
        perTable.push({ /* reconstructed row */ })
      }
    }
  }
  return perTable
}
```

  Add an integration test that exercises `pii-detector` against a fully-collapsed warehouse.

---

## Major

### M1. Threshold overrides are dead code at the public API
- **Category:** Bug / Design
- **Location:** `types.ts:372-389` (`SchemaIndexParams`), `native/schema/register.ts:29-37`, `tools/schema-index.ts:12-19`
- **Description:** `SchemaCache.indexWarehouse` accepts `{ entityRatioThreshold, entityMinTables }`, but the dispatcher handler (`schema.index`) does not destructure or forward them, `SchemaIndexParams` does not declare them, and the Zod schema in `schema_index` doesn't accept them. The PR description claims thresholds are "overridable per call" — this is true only for the internal class API.
- **Suggestion:** Add the fields to `SchemaIndexParams`, the Zod schema, and the dispatcher handler; pass them into `indexWarehouse`. Add a tool-level integration test that passes a custom `entityMinTables` and observes the result.

### M2. Entity-group `search_text` includes every table name → unbounded blob + false-positive matches
- **Category:** Performance / Logic Error
- **Location:** `cache.ts:314-321`
- **Description:** `groupSearchText` is built from `...eg.table_names` plus column names plus column types. For a 5,000-table per-tenant warehouse this becomes a multi-tens-of-KB string per row. Two consequences:
  1. Any `LIKE '%token%'` on `entity_groups_cache.search_text` scans these big blobs (the index can't help leading-wildcard LIKEs).
  2. Any common substring (e.g. `2024`, `prod`, `varchar`) anywhere in the giant concatenated text matches the group, producing false-positive structural matches.
- **Suggestion:** Drop individual table names from `search_text` (keep them only in `table_names_json` for the post-filter step that already exists at `cache.ts:455-459`). Search by `sample_table`, schema, and composite column names/types only.

### M3. `schema_index` tool emits the full `table_names` list — recreates the context blow-up the feature is supposed to prevent
- **Category:** Performance / UX
- **Location:** `tools/schema-index.ts:78`, `tools/schema-search.ts:91`
- **Description:** The whole point of collapsing entity-per-table groups is to save context. The new formatter then emits `table_names: [name1, name2, ...]` on a single line — for the motivating cases (per-tenant, per-ticker, time-partitioned) this is thousands of names per group, megabytes in the worst case.
- **Suggestion:** Truncate the displayed list (e.g. first 10 + `… and N more`) while keeping the full set in the structured response, OR move full membership behind an explicit `schema_inspect_group` tool / debug flag.

### M4. Fingerprint format breaks on column names containing `|` or `:` (and on `null`/`undefined` types)
- **Category:** Bug
- **Location:** `entity-groups.ts:80-83`
- **Description:** Two related correctness gaps:
  1. The fingerprint string `${name}:${data_type}|${name}:${data_type}` collides if `name` contains `:` or `|`. SQL identifiers can carry these via quoted identifiers in PG/Snowflake/BigQuery. Unlikely in practice for the per-tenant cases but breaks the invariant.
  2. `(c.data_type || "").toLowerCase()` collapses `null`, `undefined`, and an actual empty string to the same fingerprint slot. Drivers can return null `data_type` for some types (computed columns, etc.); two semantically-different tables can collide.
- **Suggestion:** Use `JSON.stringify([name, data_type])` per column, joined by `\x1F`, or reject columns with missing types (skip the table from grouping). Add fingerprint tests for special characters and null types.

### M5. Detector returns at most ONE group per schema; secondary patterns silently degrade to per-table
- **Category:** Design / Documentation
- **Location:** `entity-groups.ts:130-138`
- **Description:** A schema with two distinct entity-per-table patterns (e.g., 100 per-ticker tables + 80 per-region tables) gets one group plus 80 per-table rows. This is intentional but undocumented in the public types/JSDoc. The `entity_groups_cache` UNIQUE on `(warehouse, database_name, schema_name, fingerprint)` would happily store multiple groups per schema if the detector returned them.
- **Suggestion:** Either lift the single-group restriction (return all qualifying buckets above threshold), or document it explicitly in `EntityGroupDetection`. Add a test for the "two patterns in one schema" case.

### M6. Threshold parameters accept garbage without validation
- **Category:** Logic Error
- **Location:** `entity-groups.ts:102-103`, `cache.ts:187-188`
- **Description:** `ratioThreshold: -1`, `1.5`, `NaN`, or `minTables: 0`/`-5`/`3.7` produce silently wrong behavior (ratio never passes; integer comparisons get reversed; negative min triggers spuriously).
- **Suggestion:**
```ts
if (!(ratioThreshold > 0 && ratioThreshold <= 1)) throw new Error(...)
if (!Number.isInteger(minTables) || minTables < 1) throw new Error(...)
```

---

## Minor

### N1. `match_count` mixes per-row and per-group entries
- **Location:** `cache.ts:479`. `match_count` adds `entityGroups.length` directly, so one entity-group hit (which represents N collapsed tables) counts as 1, but pre-PR `match_count` would have had N rows. Comparators across versions will see misleading numbers. Consider exposing `entity_group_count` separately and keeping `match_count = tables.length + columns.length`. *(MiMo)*

### N2. Search `LIMIT` applies independently per category
- **Location:** `cache.ts:395, 417, 444`. A `limit=20` call returns up to 60 rows total. Pre-existing pattern but newly tripled by entity_groups. Document in the schema_search description. *(GLM5)*

### N3. `data_type` token pollution in entity-group search
- **Location:** `cache.ts:319-320`. `groupSearchText` includes type strings. A query like `varchar` will match every group containing any varchar column. Either drop types from the search blob or weight them down in scoring. *(Claude N2)*

### N4. Non-numeric alphabetical sort yields counter-intuitive `sample_table`
- **Location:** `entity-groups.ts:81, 155, 157`. `TICKER_10 < TICKER_2` lexicographically. Surfaced order may surprise users. Document or natural-sort. *(Claude N3, Kimi)*

### N5. `composite_columns` casing differs from fingerprint (lowercased)
- **Location:** `entity-groups.ts:153-155`. Fingerprint is lowercased, the digest preserves source casing. Correct but worth a comment to prevent later confusion. *(Claude N4, GLM5)*

### N6. `cacheStatus` aggregates `entity_groups_cache` globally, not per-warehouse
- **Location:** `cache.ts:506-510, 520-522`. Pre-existing pattern for `tables_cache`/`columns_cache` too, so behavior is internally consistent — but if cacheStatus is ever extended to a per-warehouse mode, this needs filtering. Document the semantics. *(Minimax C1, MiMo)*

### N7. Error metadata in `schema_index` tool drops `entity_groups: 0`
- **Location:** `tools/schema-index.ts:47`. Trivial inconsistency with the success path. *(Minimax, MiMo)*

### N8. `total_columns` consistency across `cacheStatus` and per-warehouse `columns_count` is untested
- **Location:** `cache.ts:354 vs 503-534`. The arithmetic happens to align today (collapsed columns added once during indexing into `warehouses.columns_count`, reconstructed once in `cacheStatus`). But there is no test asserting `total_columns == sum(warehouses[].columns_count)` for a mixed warehouse. Add one. *(Claude C1, Kimi)*

### N9. `INSERT OR REPLACE` on entity_groups_cache + UNIQUE on `(warehouse, db, schema, fingerprint)` with `db = NULL`
- **Location:** `cache.ts:91, 220-225`. SQLite treats NULLs as distinct in UNIQUE constraints. Today this is harmless because the warehouse-scoped DELETE at line 193 happens first. Worth a one-line comment to lock in the invariant for future maintainers. *(Claude)*

---

## Nit

- Comment typo `entity-groups.ts:6` — "Naively cached per-table this wastes" needs a comma. *(GLM5)*
- `databaseName: string | null = null` is dead/placeholder; either implement multi-database support or drop the variable. *(Kimi)*
- Tests use `toBeGreaterThan(0)` where exact counts are knowable. *(Kimi)*
- Re-cast in `cacheStatus`: `JSON.parse(...) as unknown[]` should be typed. *(Minimax)*

---

## Disagreements / not adopted
- **Kimi — "CRITICAL SQL injection in `search()`":** Not reproducible. Both `bun:sqlite` `.prepare(...).all(...)` and the `LIKE` pattern construction use bound parameters; the dynamic part is only the count of `?` placeholders, not user-controlled. Kimi's suggested mitigation (validate `limit` and `warehouse`) is reasonable hardening but the underlying claim of SQL injection is incorrect.
- **Minimax — "CRITICAL: `cacheStatus` aggregates across all warehouses without filter":** This is the same pre-existing pattern that `tables_cache` aggregation has used; not a regression introduced by this PR. Demoted to N6 with a docs note.
- **Kimi — "Integer overflow in `cols.length * r.table_count`":** Practically impossible — would require >2^53 collapsed columns.
- **MiMo — "Add warehouseName to entity-group `search_text`":** The warehouse filter is a separate column (`AND warehouse = ?`), not part of the search_text token list, so this is consistent with table/column rows. (Their critique reads `tables_cache` differently than the code does.)

---

## Positive observations
- Pure-function detector with no I/O is the right factoring; easy to reason about and unit-test (consensus).
- Re-index correctly clears `entity_groups_cache` (`cache.ts:193`); covered by integration test on line 854.
- `matching_tables` post-filter on search returns just the names that actually match the query, not all 2,754 — neat UX (consensus).
- DDL is `CREATE TABLE IF NOT EXISTS`, so legacy bun:sqlite caches keep working (legacy upgrade test on line 78 confirms).
- 19 unit + 7 integration tests is a reasonable baseline given the surface area.
- `TableSnapshot` local type inside `indexWarehouse` is a small but meaningful clarity win.

## Missing tests / edge cases
1. PII detection over a fully-collapsed warehouse (covers C1).
2. Tool-level threshold override (covers M1).
3. Two distinct entity-per-table patterns in one schema (covers M5).
4. Special-character column names + null `data_type` (covers M4).
5. Exactly the boundary `ratio = 0.5` and `tables = 20` cases.
6. `sql_execute` pre-validation against a collapsed warehouse (downstream of C1).
7. `total_columns == sum(warehouses[].columns_count)` invariant (N8).
8. `concurrent indexWarehouse` calls on the same warehouse (no lock today).

---

## Attribution

| # | Finding | Claude | GPT | Kimi | Minimax | GLM-5 | MiMo |
|---|---|:-:|:-:|:-:|:-:|:-:|:-:|
| C1 | `listColumns` skips collapsed columns → PII broken | M | C | M | — | C | C |
| M1 | Threshold knobs not exposed via dispatcher / tool | M | M | — | — | — | C |
| M2 | Entity-group `search_text` unbounded + false positives | — | — | — | — | — | M |
| M3 | `schema_index` emits full table_names blob | N | M | — | M | — | M |
| M4 | Fingerprint collisions (special chars + null types) | M | — | — | M | M | — |
| M5 | One-group-per-schema limitation undocumented | — | — | — | M | — | M |
| M6 | No validation on threshold params | — | — | M | M | M | — |
| N1 | `match_count` mixes rows + groups | — | — | — | — | — | M |
| N2 | LIMIT applies per-category (3× total) | — | — | — | — | M | — |
| N3 | Data-type tokens pollute entity-group search | N | — | — | — | — | — |
| N4 | Lex sort surprises (`TICKER_10 < TICKER_2`) | N | — | N | — | — | — |
| N5 | Casing differs between fingerprint and digest | N | — | — | — | N | — |
| N6 | `cacheStatus` global vs per-warehouse semantics | — | — | — | C* | — | M |
| N7 | Error metadata missing entity_groups:0 | — | — | — | N | — | N |
| N8 | `total_columns` consistency untested | C | — | M | — | — | — |
| N9 | UNIQUE w/ NULL db handled by DELETE-first | N | — | — | — | — | — |

C = critical, M = major, N = minor/nit, — = not raised. C* = Minimax raised it as critical; demoted by majority to nit/doc per "Disagreements" above.

---

Reviewed by 8 participants: Claude + GPT 5.4 Codex + Gemini 3.1 Pro + Kimi K2.5 + MiniMax M2.7 + GLM-5.1 + Qwen 3.6 + MiMo V2 Pro. Gemini was blocked by sandbox workspace restrictions and produced no review. Qwen exceeded its 15-minute budget mid-investigation and produced no review. The remaining 6 (including Claude) converged on the findings above.

sahrizvi:
## Update: addressed-vs-pending from the [consensus review](https://github.com/AltimateAI/altimate-code/pull/762#issuecomment-4335048612)

Pushed `dc2af3c0a` on top of the original branch.

### Addressed (this commit)

|---|---|---|---|
| C1 | CRITICAL | `listColumns` skips collapsed columns → silent PII false-cleans | `listColumns` now reconstructs synthetic per-`(member_table, composite_col)` rows from `entity_groups_cache` for collapsed warehouses; the `limit` is applied to the combined result. Verified both readers — `pii-detector.ts:50` (PII scan loop) and `sql-execute.ts:171` (SQL pre-validation) — now see complete column visibility. |
| M1 | MAJOR | Threshold knobs dead at the dispatcher / tool layer | `entityRatioThreshold` / `entityMinTables` added to `SchemaIndexParams`, threaded through the `schema.index` handler in `register.ts`, and exposed in the `schema_index` Zod tool. Per-call overrides are now wired end-to-end. |
| M2 | MAJOR | Unbounded `search_text` blob with false-positive matches | Entity-group `search_text` excludes member table names and column types. `search()` now matches table-name queries via `OR table_names_json LIKE ?` so single-table lookups still hit the right group. |
| M3 | MAJOR | Formatter emits the full N-thousand `table_names` list | `schema_index` formatter truncates to head 50 + tail 5 + `+N more (use schema_search to enumerate)` hint. The structured response keeps the full list intact for programmatic callers. Also tightens N7: error metadata now includes `entity_groups: 0` for parity with the success path. |
| M4 | MAJOR | Fingerprint collisions on `:` / `\|` separators and null types | `fingerprintColumns` now uses `JSON.stringify([name, type])` joined by `\x1F` (delimiter-safe). Null/undefined `data_type` values map to a sentinel distinct from the empty string. |
| M5 | MAJOR | One-group-per-schema undocumented | JSDoc on `EntityGroupDetection` and the `schema_index` tool description both call out the one-dominant-group scope; secondary patterns fall through to `remaining_tables`. |
| M6 | MAJOR | No input validation on threshold params | `detectEntityGroup` rejects NaN, `≤0`, `>1`, non-integer `minTables`, and `minTables < 2`. Matching Zod constraints (`.gt(0).lte(1)`, `.int().min(2)`) on the tool surface so callers see the failure early. |

**Tests:** 87 pass / 0 fail across `entity-groups.test.ts` (M4, M6), `schema-cache.test.ts` (C1 reconstruction across fully-collapsed / mixed / limit-bounded / ghost-warehouse plus a PII-flow integration), and a new `schema-index-tool.test.ts` (M1 dispatcher forwarding, M3 truncation snapshot for 2,754-table case, M6 Zod rejection). Wider regression on `test/altimate/` is 2,897 pass; the 4 pre-existing flaky `tool-lookup` timeouts are reproducible on clean `b7584c5c3` so they're not from this change. Typecheck clean across all 7 packages.

### Deferred

- All MINOR / NIT items (N1–N9 except the trivial N7 metadata-consistency bundled with M3)
- N8 `total_columns == sum(columns_count)` invariant test — out of scope for this round

### Demoted / invalid

- Kimi's "SQL injection" claim — invalid (`bun:sqlite` uses bound parameters)
- MiniMax's "`cacheStatus` aggregates across all warehouses" — pre-existing pattern, not a regression introduced by this PR

### Heads-up

Reconstructed rows default `nullable: true` since the entity fingerprint doesn't carry per-column nullability today. This is an intentional minor info loss matching `columns_cache`'s default — flagged inline at `cache.ts` so future work can extend the fingerprint if PII rules ever need it.

dev-punia-altimate:
## ✅ Tests — All Passed

### TypeScript — passed


cc @sahrizvi

PR #763, opened 2026-04-28
feat(session): commit best-guess answer in headless mode at max-turns

### LINKED ISSUE #759: [feat] Headless mode: commit best-guess answer at max-turns instead of meta-prose

sahrizvi:
Closes #759

## Motivation

In headless mode (`-p` / `--print`), when the agent hits `--max-turns`, the existing MAX_STEPS injection produces a multi-paragraph "I have reached the maximum number of steps" summary instead of a committed answer. In interactive mode this is fine — a human can react. In headless / scripted use, there is no human; the program's final stdout is unparseable meta-prose.

Common situations where this matters:
- CI pipelines: `claude -p "review this PR for security issues"` returns a status update instead of a partial review
- Eval harnesses (any non-interactive batch grader)
- Composable shell pipelines: `git diff | claude -p "summarize"` — works only if the agent always emits the requested shape
- Scheduled / cron jobs: produce parseable output regardless of run length

## What's changed

This PR keeps interactive behaviour completely unchanged and adds a separate prompt path for headless mode.

**New prompt files:**
- `prompt/max-steps-headless.txt` — final-step prompt explicitly instructing the model to emit the answer (not a summary). The wording makes the rationale explicit: an uncertain best-guess is more useful than meta-prose, because the caller cannot ask for retries.
- `prompt/max-steps-headless-prewarn.txt` — soft pre-warning that fires one step before the limit when headless and `Number.isFinite(maxSteps)`. The `isFinite` guard prevents misfiring when no max-turns is set.

**Wiring:**
- `session/prompt.ts`: exported `selectMaxStepsPrompt({ step, maxSteps, headless })` — pure selector returning the right prompt for the current state, plumbed into the existing prompt loop.
- `cli/cmd/run.ts`: passes `headless: true` from the `-p` / `--print` CLI entry into the session SDK call.
- `session/message-v2.ts`: extended the `User` zod schema with optional `headless: boolean`, persisted on the user message so resumed sessions retain the flag.
- SDK gen cascade (`sdk.gen.ts` + `types.gen.ts`): updated by the existing `@hey-api/openapi-ts` step.

## Tests

**7 new tests** in `test/session/max-steps-prompt.test.ts`:
1. Mid-loop returns interactive prompt
2. Interactive last step returns interactive prompt
3. Headless last step returns headless prompt
4. Headless one-step-before-last returns prewarning
5. Interactive one-step-before-last returns nothing
6. Over-limit returns headless prompt (continues to fire if invoked late)
7. Infinite-budget (no maxSteps) returns nothing — `Number.isFinite` guard

**Results:**
- This file: **7 pass / 0 fail**
- Full session test directory: **543 pass / 4 skip / 0 fail** across 24 files (~8.6s) — no regressions
- Full opencode suite when merged with #758 + #760: **7421 pass / 0 fail / 494 skip** (~88s)
- `bun run typecheck` (opencode + sdk/js): clean

## Backwards compatibility

- Interactive prompt text completely unchanged — only headless invocations see new behaviour.
- New `headless?: boolean` field is optional everywhere; older SDK consumers see no breaking change.
- A session run with `-p` and then resumed (e.g., session reload via the API) still applies headless behaviour, which matches the original intent — the flag persists on the user message.

## Notes

- The headless prompt's full text is in `max-steps-headless.txt`; it explicitly tells the model not to emit a meta-summary, with the reasoning (no human to ask follow-ups, an uncertain answer is more useful than no answer).
- The optional pre-warning is wired in and tested. It only fires in headless mode and only when `maxSteps` is finite; interactive mode never sees it.

## Files

- `packages/opencode/src/cli/cmd/run.ts`
- `packages/opencode/src/session/message-v2.ts`
- `packages/opencode/src/session/prompt.ts`
- `packages/opencode/src/session/prompt/max-steps-headless.txt` (new)
- `packages/opencode/src/session/prompt/max-steps-headless-prewarn.txt` (new)
- `packages/opencode/test/session/max-steps-prompt.test.ts` (new)
- `packages/sdk/js/src/v2/gen/types.gen.ts`
- `packages/sdk/js/src/v2/gen/sdk.gen.ts`

---
## Summary by cubic
Headless runs now commit a best-guess answer at `--max-turns` and actually disable tools on the final step. This makes CI/evals reliably parseable and fully addresses #759.

- **Bug Fixes**
  - Propagate `headless: true` through all non-interactive paths: `run --command`, GitHub, and GitLab commands.
  - Preserve `headless` on synthetic user messages (post-task summary, compaction replay/continue) so long runs don’t flip back to interactive.
  - Enforce tools-disabled on the headless final step (strip tools, set toolChoice to none; exempt `json_schema`).
  - Add `headless` to `User` and to `session.prompt`/`session.prompt_async`/`session.command` in `openapi.json` and the JS SDK types.
  - Make the prewarn copy consistent via `{TURNS_REMAINING}` substitution; add focused tests for selection, tool-disabling, schemas, and edge cases.

<sup>Written for commit 78acb0fc42f2264e525d359971d1992b68515373. Summary will update on new commits. <a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/763?utm_source=github">Review in cubic</a></sup>

## Summary by CodeRabbit

* **New Features**
  * Headless mode for non-interactive runs now produces a final best-guess answer (rather than a meta-summary) and disables tool use on the final headless step.
  * CLI/CI executions default to headless behavior; API now accepts an optional headless flag.

* **Tests**
  * Added tests covering max-steps prompt selection, headless pre-warn/final behavior, and headless flag propagation.

### DIFF
--- a/packages/opencode/src/cli/cmd/github.ts
+++ b/packages/opencode/src/cli/cmd/github.ts
@@ -941,6 +941,9 @@ export const GithubRunCommand = cmd({
             providerID,
             modelID,
           },
+          // altimate_change - non-interactive CI runner; force headless so max-steps
+          // emits a best-guess answer rather than a meta-summary.
+          headless: true,
           // agent is omitted - server will use default_agent from config or fall back to "build"
           parts: [
             {
@@ -996,6 +999,9 @@ export const GithubRunCommand = cmd({
             modelID,
           },
           tools: { "*": false }, // Disable all tools to force text response
+          // altimate_change - non-interactive CI runner; force headless so max-steps
+          // emits a best-guess answer rather than a meta-summary.
+          headless: true,
           parts: [
             {
               id: PartID.ascending(),
--- a/packages/opencode/src/cli/cmd/gitlab.ts
+++ b/packages/opencode/src/cli/cmd/gitlab.ts
@@ -453,6 +453,9 @@ async function runReview(
     variant,
     model: { providerID, modelID },
     tools: { "*": false }, // No tools needed for review — text-only
+    // altimate_change - non-interactive CI runner; force headless so max-steps
+    // emits a best-guess answer rather than a meta-summary.
+    headless: true,
     parts: [
       {
         id: PartID.ascending(),
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -815,6 +815,10 @@ You are speaking to a non-technical business executive. Follow these rules stric
           variant: args.variant,
           parts: [...files, { type: "text", text: message }],
           ...(audienceSystem ? { system: audienceSystem } : {}),
+          // altimate_change - the `run` command is the headless / non-interactive
+          // entry point. Tell the session loop so the max-steps prompt commits a
+          // best-guess answer instead of writing a meta-summary.
+          headless: true,
         })
       }
 
--- a/packages/opencode/src/cli/cmd/run.ts
+++ b/packages/opencode/src/cli/cmd/run.ts
@@ -805,6 +805,10 @@ You are speaking to a non-technical business executive. Follow these rules stric
           command: args.command,
           arguments: message,
           variant: args.variant,
+          // altimate_change - the `run --command` path is also non-interactive
+          // (no human in the loop). Propagate headless so the max-steps prompt
+          // commits a best-guess answer instead of writing a meta-summary.
+          headless: true,
         })
       } else {
         const model = args.model ? Provider.parseModel(args.model) : undefined
--- a/packages/opencode/src/session/compaction.ts
+++ b/packages/opencode/src/session/compaction.ts
@@ -342,6 +342,9 @@ When constructing the summary, try to stick to this template:
           tools: original.tools,
           system: original.system,
           variant: original.variant,
+          // altimate_change - retain headless flag on replayed user msg so the
+          // max-steps prompt selection stays consistent after compaction.
+          headless: original.headless,
         })
         for (const part of replay.parts) {
           if (part.type === "compaction") continue
@@ -364,6 +367,9 @@ When constructing the summary, try to stick to this template:
           time: { created: Date.now() },
           agent: userMessage.agent,
           model: userMessage.model,
+          // altimate_change - retain headless flag on continuation user msg so
+          // the max-steps prompt selection stays consistent after compaction.
+          headless: userMessage.headless,
         })
         const text =
           (input.overflow
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -372,6 +372,15 @@ export namespace MessageV2 {
     // altimate_change - true when the prompt was sent from a non-interactive
     // caller (e.g. the `run` CLI command). Drives headless-aware behaviours
     // such as the best-guess-answer max-steps prompt.
+    //
+    // Layer rationale: stored on the User message (not Session.Info) so that
+    // (a) resumed sessions whose last user message was sent headlessly retain
+    // the behaviour, and (b) sessions reused by both interactive and headless
+    // callers (e.g. a TUI session continued from a `run -p` invocation) reflect
+    // the *most recent* invocation mode rather than the original session origin.
+    // Every synthetic-user-message constructor (task-summary, compaction
+    // replay/continue) MUST propagate this flag from the prior `lastUser`
+    // explicitly — a missing copy silently flips the mode back to interactive.
     headless: z.boolean().optional(),
   }).meta({
     ref: "UserMessage",
--- a/packages/opencode/src/session/message-v2.ts
+++ b/packages/opencode/src/session/message-v2.ts
@@ -369,6 +369,10 @@ export namespace MessageV2 {
     system: z.string().optional(),
     tools: z.record(z.string(), z.boolean()).optional(),
     variant: z.string().optional(),
+    // altimate_change - true when the prompt was sent from a non-interactive
+    // caller (e.g. the `run` CLI command). Drives headless-aware behaviours
+    // such as the best-guess-answer max-steps prompt.
+    headless: z.boolean().optional(),
   }).meta({
     ref: "UserMessage",
   })
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -24,6 +24,10 @@ import { Plugin } from "../plugin"
 import PROMPT_PLAN from "../session/prompt/plan.txt"
 import BUILD_SWITCH from "../session/prompt/build-switch.txt"
 import MAX_STEPS from "../session/prompt/max-steps.txt"
+// altimate_change start - headless-mode max-steps prompt commits a best-guess answer
+import MAX_STEPS_HEADLESS from "../session/prompt/max-steps-headless.txt"
+import MAX_STEPS_HEADLESS_PREWARN from "../session/prompt/max-steps-headless-prewarn.txt"
+// altimate_change end
 import { defer } from "../util/defer"
 import { ToolRegistry } from "../tool/registry"
 import { MCP } from "../mcp"
@@ -99,6 +103,25 @@ export namespace SessionPrompt {
     if (match) throw new Session.BusyError(sessionID)
   }
 
+  // altimate_change start - exposed for unit testing the prompt-selection logic.
+  // Picks which max-steps prompt (if any) to inject at the start of a step.
+  // - Final step in headless mode: ask for a best-guess answer NOW, since there
+  //   is no human to follow up.
+  // - Final step in interactive mode: existing summarize-what-you-tried wording.
+  // - One step before the final step in headless mode: a softer pre-warning that
+  //   nudges the model to start writing its answer.
+  export function selectMaxStepsPrompt(input: {
+    step: number
+    maxSteps: number
+    headless: boolean
+  }): string | undefined {
+    const { step, maxSteps, headless } = input
+    if (step >= maxSteps) return headless ? MAX_STEPS_HEADLESS : MAX_STEPS
+    if (headless && Number.isFinite(maxSteps) && step === maxSteps - 1) return MAX_STEPS_HEADLESS_PREWARN
+    return undefined
+  }
+  // altimate_change end
+
   export const PromptInput = z.object({
     sessionID: SessionID.zod,
     messageID: MessageID.zod.optional(),
@@ -119,6 +142,12 @@ export namespace SessionPrompt {
     format: MessageV2.Format.optional(),
     system: z.string().optional(),
     variant: z.string().optional(),
+    // altimate_change start - mark sessions invoked from non-interactive callers
+    // (e.g. the `run` CLI command). When true, the max-steps prompt switches to a
+    // best-guess-answer wording instead of the interactive "summarize what you
+    // tried" wording, since there is no human to follow up.
+    headless: z.boolean().optional(),
+    // altimate_change end
     parts: z.array(
       z.discriminatedUnion("type", [
         MessageV2.TextPart.omit({
@@ -624,7 +653,11 @@ export namespace SessionPrompt {
       // normal processing
       const agent = await Agent.get(lastUser.agent)
       const maxSteps = agent.steps ?? Infinity
-      const isLastStep = step >= maxSteps
+      // altimate_change start - select max-steps injection text (if any). In
+      // headless mode the final-step prompt asks for a best-guess answer; one
+      // step earlier, a softer pre-warning nudges the model to start writing.
+      const maxStepsInjection = selectMaxStepsPrompt({ step, maxSteps, headless: !!lastUser.headless })
+      // altimate_change end
       msgs = await insertReminders({
         messages: msgs,
         agent,
@@ -878,11 +911,12 @@ export namespace SessionPrompt {
         system,
         messages: [
           ...MessageV2.toModelMessages(msgs, model),
-          ...(isLastStep
+          // altimate_change - inject the max-steps text chosen above (if any)
+          ...(maxStepsInjection
             ? [
                 {
                   role: "assistant" as const,
-                  content: MAX_STEPS,
+                  content: maxStepsInjection,
                 },
               ]
             : []),
@@ -1330,6 +1364,8 @@ export namespace SessionPrompt {
       system: input.system,
       format: input.format,
       variant,
+      // altimate_change - persist headless flag for downstream prompt selection
+      headless: input.headless,
     }
     using _ = defer(() => InstructionPrompt.clear(info.id))
 
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -110,6 +110,9 @@ export namespace SessionPrompt {
   // - Final step in interactive mode: existing summarize-what-you-tried wording.
   // - One step before the final step in headless mode: a softer pre-warning that
   //   nudges the model to start writing its answer.
+  // The pre-warning template substitutes `{TURNS_REMAINING}` with the count of
+  // tool-using turns left (computed as `maxSteps - step`) so the wording is
+  // always internally consistent with the actual budget.
   export function selectMaxStepsPrompt(input: {
     step: number
     maxSteps: number
@@ -117,8 +120,26 @@ export namespace SessionPrompt {
   }): string | undefined {
     const { step, maxSteps, headless } = input
     if (step >= maxSteps) return headless ? MAX_STEPS_HEADLESS : MAX_STEPS
-    if (headless && Number.isFinite(maxSteps) && step === maxSteps - 1) return MAX_STEPS_HEADLESS_PREWARN
+    if (headless && Number.isFinite(maxSteps) && step === maxSteps - 1) {
+      const remaining = maxSteps - step
+      return MAX_STEPS_HEADLESS_PREWARN.replaceAll("{TURNS_REMAINING}", String(remaining))
+    }
     return undefined
+  }
+
+  // Extracted so the loop's tool-stripping behaviour is unit-testable.
+  // At the headless final step, the injected prompt promises "Tools are
+  // disabled" — to make that promise honest we strip the active tool set from
+  // the upstream request. JSON-schema mode is exempt because StructuredOutput
+  // is the only way to capture the structured response.
+  export function shouldDisableToolsForHeadlessFinalStep(input: {
+    step: number
+    maxSteps: number
+    headless: boolean
+    formatType: string
+  }): boolean {
+    const { step, maxSteps, headless, formatType } = input
+    return headless && Number.isFinite(maxSteps) && step >= maxSteps && formatType !== "json_schema"
   }
   // altimate_change end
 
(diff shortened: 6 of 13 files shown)


sahrizvi:
## Multi-Model Consensus Review — PR #763

**Verdict: REQUEST CHANGES** — solid design and well-tested selector, but the headless flag is silently dropped by at least two real codepaths and the SDK auto-gen is desynchronized from `openapi.json`. All issues are tractable; once the propagation gaps are fixed this looks ready to ship.

---

### Critical / Major

**C1. `run --command` path does not propagate `headless: true`** *(Bug / Logic Error)*
**Location**: `packages/opencode/src/cli/cmd/run.ts:800-808`, plus `prompt.ts:CommandInput` (~2120) and `command()` (~2280)

The `run` CLI has two branches: `sdk.session.prompt(...)` (which got `headless: true`) and `sdk.session.command(...)` (which did not). `SessionPrompt.command()` calls `prompt({...})` internally without forwarding any `headless`, so `altimate-code -p --command foo "..."` keeps the old meta-summary at max-turns — exactly the case the PR is supposed to fix. Fix: add `headless` to `CommandInput` zod schema, forward it inside `command()`, regenerate the SDK, and pass `headless: true` from `run.ts` on the command branch as well.

**C2. Synthetic user message after `task.command` drops `headless`** *(Bug / Logic Error)*
**Location**: `packages/opencode/src/session/prompt.ts:572-580` (`summaryUserMsg`); same shape concern at the shell-command synthetic user (~1874) and compaction replay/continue paths (`compaction.ts:~334, 360`)

After a subtask command, the loop inserts a synthetic `role: "user"` message ("Summarize the task tool output above and continue with your task."). The next iteration's `lastUser` is this new message — constructed without `headless`. So `lastUser.headless` becomes `undefined`, `selectMaxStepsPrompt` falls back to interactive, and the prewarn never fires. Same risk applies to compaction-created user messages. Fix: propagate `headless: lastUser.headless` (or the original session's value) in every synthetic user-message constructor. Long-running headless runs that hit subtasks or compaction are exactly the cases this PR is targeting, so this is a real regression of intent.

**C3. Auto-gen SDK diverges from `packages/sdk/openapi.json`** *(Bug)*
**Location**: `packages/sdk/openapi.json` (unchanged) vs `packages/sdk/js/src/v2/gen/{types,sdk}.gen.ts` (changed)

The PR description calls these an "auto-gen cascade" via `@hey-api/openapi-ts`, but the source `openapi.json` does not contain `headless` in `UserMessage` or in the `SessionPromptData` / `SessionPromptAsyncData` request bodies. The next codegen run will silently wipe `headless` from the SDK. Fix: regenerate `openapi.json`, or add `headless` to it manually so subsequent codegen is stable. Verify any sibling SDKs (Python/Kotlin) need parallel updates.

**M1. Storing `headless` on the user message is the wrong layer** *(Design)*
**Location**: `packages/opencode/src/session/message-v2.ts:372`, `prompt.ts:1367`, `prompt.ts:659`

`headless` is a property of *how this session was invoked*, not of an individual user message. Putting it on `User` forces every synthetic user-message constructor to remember to copy it (see C2). It also breaks the PR's stated promise that "resumed sessions retain the behavior" — that only holds when the resumed session's `lastUser` happens to be a flagged one. A resume that adds a fresh interactive user message would silently flip the mode. Recommend storing on `Session.Info` (or computing from session origin metadata) instead. If keeping it on User is intentional, document the resume semantics carefully and propagate to every user-message creation site.

**M2. Other non-interactive entry points also lack `headless: true`** *(Design)*
**Location**: `packages/opencode/src/cli/cmd/gitlab.ts:~450`, `packages/opencode/src/cli/cmd/github.ts:~936`, `packages/opencode/src/acp/agent.ts:~1420`

GitHub/GitLab CI entry points and the ACP agent all call `sdk.session.prompt()` without `headless: true`. If any of these run with agent step limits, they will see the old summarize behavior despite being headless in practice. Either fix as part of this PR or file a tracked follow-up; do not silently leave them.

**M3. Final-step prompt promises "Tools are disabled" but tools are NOT actually disabled** *(Bug / Logic Error)*
**Location**: `packages/opencode/src/session/prompt/max-steps-headless.txt`, `prompt.ts:~880-893`

Both the new headless prompt and the existing interactive `max-steps.txt` claim "Tools are disabled," but the loop still passes `tools` to `processor.process({...})`. The injection is policy-only nudging. A model that ignores the instruction can still call tools, which (a) burns more steps, (b) potentially re-enters the loop, (c) makes the prompt's claim a lie. Either actually drop `tools` from the request when injecting the final-step prompt, or soften the wording from "disabled" to "you must not call tools." Real disablement is the safer fix and applies to interactive mode too.

**M4. Prewarn copy is inaccurate / mixed framing** *(Bug / Documentation)*
**Location**: `packages/opencode/src/session/prompt/max-steps-headless-prewarn.txt:3,7`

The prewarn says "You have 2 turns left before tools are disabled" but the next sentence ("After this turn you will have only one more chance to respond... tools will be disabled") implies only one tool-using turn remains. Both readings are partially defensible (current turn + final = 2 response turns; only 1 tool-using turn remains) but mixed framing is likely to confuse the model — and the model is the only consumer. Recommend a single consistent phrasing, e.g., "You have 1 more tool-using turn before tools are disabled. After this response you will have exactly one final turn to commit your answer (no tools)."

---

### Minor

**m1. Schema lacks explicit default for `headless`** *(Code Quality)*
`packages/opencode/src/session/message-v2.ts:373` and `prompt.ts:148` declare `headless: z.boolean().optional()` with no `.default(false)`. Runtime is fine because of `!!lastUser.headless`, but an explicit default makes the intent clearer at the type level and protects against future readers who don't double-bang.

**m2. `!!lastUser.headless` masks `false` vs `undefined`** *(Code Quality)*
`prompt.ts:659`. Today it's harmless; with `.default(false)` (m1) it becomes a no-op. Consider `lastUser.headless ?? false` for readability. Defensive `?.` chaining is unnecessary because line 378 already guarantees `lastUser` exists.

**m3. Missing edge-case tests** *(Testing)*
`test/session/max-steps-prompt.test.ts` covers `maxSteps = 10` and `Infinity` only. Add:
- `maxSteps = 1` (final fires immediately, prewarn skipped because `step === 0` never holds with `step` starting at 1)
- `maxSteps = 2` (prewarn at step 1, final at step 2)
- explicit assertion that prewarn does NOT fire when `step > maxSteps` in headless mode
- explicit assertion that headless final-step prompt does NOT contain `INTERACTIVE_SUMMARY_MARKER` (already covered) AND that interactive over-limit does NOT contain `HEADLESS_MARKER`.

**m4. No integration test for the wiring** *(Testing)*
The unit test exercises the selector but not the full path: `run -p` → SDK → `prompt({headless: true})` → persisted on User → read back at max-turns. Adding even one end-to-end assertion would have caught C1.

**m5. No test for resumed-session retention** *(Testing)*
The PR claims resumed sessions retain headless behavior, but there's no test that loads a session, hits max-turns, and confirms the headless prompt fires. Even an integration-style test that round-trips `headless` through the User message JSON would close the gap.

**m6. `INTERACTIVE_MARKER` is a substring of `HEADLESS_MARKER`** *(Testing)*
`test/session/max-steps-prompt.test.ts:11`. `"MAXIMUM STEPS REACHED"` substrings inside `"MAXIMUM STEPS REACHED (HEADLESS MODE)"`, so a copy-edit could silently flip prompt selection without the assertions catching it. Use a unique-to-interactive marker like `"Recommendations for what should be done next"`.

---

### Nit

- **n1.** Mixed `// altimate_change` comment styles (`start`/`end` blocks vs single-line `// altimate_change - ...`). Pick one. *(Code Quality)*
- **n2.** PR description references `-p / --print`; the `run` CLI uses `-p` for `--password` and the `headless: true` is set unconditionally on every `run` invocation (intentional — `run` IS the headless entry). Update the PR description to avoid confusion. *(Documentation)*
- **n3.** Test imports namespace then uses `SessionPrompt.selectMaxStepsPrompt` everywhere — fine, matches house style. *(Code Quality)*
- **n4.** `selectMaxStepsPrompt` returns `string | undefined`. A discriminated union (`{type: "none"} | {type: "final", text} | {type: "prewarn", text}`) would be more self-documenting and future-proof, but is over-engineering for current needs. *(Design)*

---

### Positive Observations

- Extracting `selectMaxStepsPrompt` is exactly right — pure, testable, no need to mock the loop.
- `Number.isFinite(maxSteps)` guard on the prewarn is the kind of subtle bug that gets missed; nicely caught.
- New headless prompt is well-crafted: explicit anti-summary instructions, explicit "uncertain answer is more useful than meta-summary" framing, format-preservation directive.
- Backward compatibility is genuinely additive: interactive mode untouched, optional flag, infinite-budget handled.
- Tests use markers in both directions (must-contain and must-not-contain), which is the right defensive style.
- `step >= maxSteps` (not `===`) correctly handles overshoot — preserved from upstream.
- Single-commit, well-scoped diff with clear `altimate_change` markers, easing future upstream rebases.

---

### Missing Tests / Edge Cases

| Edge case | Coverage | Risk |
|---|---|---|
| `run --command` headless propagation | None | **High** (C1) |
| Synthetic user msg after `task.command` retains headless | None | **High** (C2) |
| Compaction replay/continue retains headless | None | High |
| Session resume retains headless | None | Medium |
| End-to-end: `run -p` produces non-meta output at max-turns | None | Medium |
| `maxSteps = 1` (final fires at first step, no prewarn) | None | Low |
| `maxSteps = 2` (prewarn step 1, final step 2) | None | Low |
| GitHub / GitLab / ACP non-interactive callers | None (out of scope but related) | Medium |

---

### Attribution

| Finding | Models flagging |
|---|---|
| C1 `run --command` drops `headless` | Claude, GPT, MiniMax, Qwen, MiMo (5/8) — **strong consensus** |
| C2 Synthetic user (task.command) drops `headless` | Claude, GPT (2/8) — unique but verified |
| C2b Compaction-path drops `headless` | GPT (1/8) — unique, plausible |
| C3 SDK gen vs `openapi.json` drift | Claude (1/8) — unique, verified by reading `openapi.json` |
| M1 Wrong layer for `headless` | Claude (1/8) |
| M2 GitHub/GitLab/ACP also need flag | MiMo, Kimi (2/8) |
| M3 "Tools disabled" lie (also affects interactive) | Claude (1/8) — unique |
| M4 Prewarn "2 turns" wording | Claude, Kimi, MiniMax, GLM-5, MiMo (5/8) — **strong consensus** |
| m1 No `.default(false)` | Qwen (1/8) |
| m2 `!!lastUser.headless` style | Claude, Kimi, Qwen (3/8) |
| m3 `maxSteps=1`, `maxSteps=2` test gaps | Claude, Kimi, MiniMax, GLM-5, Qwen, MiMo (6/8) — **strong consensus** |
| m4 No end-to-end wiring test | Claude, GPT (2/8) |
| m5 No resume-retention test | Claude, Kimi, MiMo (3/8) |
| m6 Substring marker brittleness | Claude (1/8) |
| n1 Mixed `altimate_change` style | Kimi (1/8) |
| n2 PR description `-p`/`--print` mismatch | Kimi (1/8) |
| n4 Discriminated-union return | Kimi (1/8) |

Disagreements: GLM-5 marked the PR "READY TO MERGE" with only minor test gaps, missing C1/C2/C3 entirely. Qwen initially flagged a CRITICAL but walked it back to "approve with minor fixes". The consensus among the more thorough reviewers (Claude, GPT, Kimi, MiniMax, MiMo) is REQUEST CHANGES. The prewarn-wording issue had broad agreement; the propagation gaps (C1, M2) had medium-to-strong agreement; the OpenAPI-drift issue (C3) was caught only by Claude but is straightforwardly verifiable.

---

*Reviewed by 8 participants: Claude + GPT 5.4 Codex + Gemini 3.1 Pro + Kimi K2.5 + MiniMax M2.7 + GLM-5.1 + Qwen 3.6 + MiMo V2 Pro. Gemini failed to launch (CLI rejected the prompt because it contained `-p`/`--print` strings interpreted as flags).*

sahrizvi:
## Update: addressed-vs-pending from the [consensus review](https://github.com/AltimateAI/altimate-code/pull/763#issuecomment-4335071060)

Pushed `78acb0fc4` on top of the original branch.

### Addressed (this commit)

|---|---|---|---|
| C1 | CRITICAL | `run --command` flow drops `headless` | `CommandInput` zod schema accepts `headless`; `command()` forwards it into the internal `prompt({...})` call; `cli/cmd/run.ts` `--command` branch sets `headless: true`. |
| C2 | CRITICAL | Synthetic user message paths drop `headless` | `summaryUserMsg` (task summary), compaction `replay`, and compaction `continue` user-message constructors all now propagate `headless` from `lastUser` / `userMessage`. JSDoc on `MessageV2.User.headless` documents the contract that every synthetic-user-message constructor must propagate the flag. |
| C3 | CRITICAL | SDK gen drift from `openapi.json` | `packages/sdk/openapi.json` patched in the 4 schemas the PR cares about (`UserMessage`, `session.prompt`, `session.prompt_async`, `session.command` request bodies). Regenerated `types.gen.ts` / `sdk.gen.ts` match the patched schemas. Verified by re-running the generator and confirming `"headless"` appears 4× in the expected places. |
| M1 | MAJOR | Wrong layer for the flag (Session vs. User) | Kept on `User` message rather than `Session.Info`. Rationale documented in JSDoc: resumed sessions whose last user message was sent headlessly retain the behaviour, and sessions reused by both interactive and headless callers reflect the most recent invocation mode. |
| M2 | MAJOR | Other non-interactive callers also lack the flag | Audited and patched: `cli/cmd/github.ts` (2 CI-runner `SessionPrompt.prompt` calls), `cli/cmd/gitlab.ts` (`runReview`). ACP (`acp/agent.ts`) and TUI shell (`session/prompt.ts::shell()`) intentionally NOT changed — both serve interactive editor / human-driven flows. |
| M3 | MAJOR | "Tools disabled" was policy-only | At the headless final step the loop now strips the active tool set (`tools: {}`, `toolChoice: "none"`) before calling `processor.process`. JSON-schema mode is exempt (StructuredOutput tool must remain). Logic extracted to exported `shouldDisableToolsForHeadlessFinalStep` for unit-testability. |
| M4 | MAJOR | Prewarn copy is mixed | Rewrote `max-steps-headless-prewarn.txt` with `{TURNS_REMAINING}` token, substituted at runtime from `maxSteps - step`. Old "2 turns" / "1 tool-using turn" contradiction is gone; copy is internally consistent. |

**Tests:** 7,394 pass / 0 fail full opencode suite (`bun test`); typecheck clean across 5 packages. Targeted: `max-steps-prompt.test.ts` 19/0, `prompt.test.ts` 9/0, full `test/session/` 560/0, full `test/cli/` 492/0. Coverage spans edge cases (`maxSteps=1` one-shot, `maxSteps=2` prewarn-then-final, overshoot, no-budget Infinity), `CommandInput` / `PromptInput` zod schemas, headless round-trip through `MessageV2` storage, and prewarn-copy consistency.

### Deferred (per consensus tags)

- Minor m1 (`.default(false)`), m2 (`?? false` style), m4 (full e2e CLI subprocess test), m5 (resume-retention test)
- Nits n1-n4 (comment style, PR description copy, discriminated-union return type)
- m6 partially addressed (replaced substring-vulnerable "MAXIMUM STEPS REACHED" marker with "Recommendations for what should be done next")

### Heads-up

The committed `packages/sdk/openapi.json` has substantial pre-existing drift from a clean regen (different `info.title`, single-line vs multi-line `required` arrays, missing `/config/tui` route, formatting style). This PR's manual patch only touches the 4 schemas relevant to `headless`, leaving the broader drift unchanged. If a maintainer later runs the full SDK regen via `bun src/index.ts generate > ../sdk/openapi.json`, that drift will surface — recommend addressing it in a separate cleanup PR.

dev-punia-altimate:

- connection_refused [2.67ms]
- timeout [2.66ms]
- permission_denied [2.69ms]
- parse_error [2.40ms]
- oom [2.65ms]
- network_error [2.42ms]
- auth_failure [2.64ms]
- rate_limit [2.77ms]
- internal_error [2.82ms]
- empty_error [0.24ms]
- connection_refused [0.14ms]
- timeout [0.08ms]
- permission_denied [0.07ms]
- parse_error [0.07ms]
- oom [0.07ms]

### Next Step

cc @sahrizvi

PR #767, opened 2026-05-01
fix(permission): route HTTP reply to `PermissionNext` to unblock tool ask deadlock

suryaiyer95:
## Summary

The v1.4.0 upstream merge introduced a new Effect-based `Permission` namespace (`packages/opencode/src/permission/index.ts`) and rewired the HTTP permission routes to it, but **left the tool ask path on the existing `PermissionNext` namespace** (`packages/opencode/src/permission/next.ts`). Result: ask and reply now talk to two separate in-memory pending Maps, and every permission round-trip silently no-ops.

This PR reverts the routing layer back to `PermissionNext`, matching `main`.

## Root cause (verified end-to-end)

1. `tool/bash.ts:155` and `altimate/tools/sql-execute.ts:32` call `ctx.ask(...)` → wired to `PermissionNext.ask` via `session/processor.ts:193,221` and `session/prompt.ts:517,1239`. The deferred `{resolve, reject}` is stored in **`PermissionNext`'s pending Map**, then `permission.asked` is published.
2. TUI `--yolo` auto-reply (`cli/cmd/tui/context/sync.tsx:180`) and manual `Allow once` / `Allow always` / `Reject` buttons all reply via `POST /permission/{id}/reply`.
3. Post-merge, `server/routes/permission.ts:39` invoked `Permission.reply` (`permission/index.ts:204`), which looked up the id in **`Permission`'s pending Map**, found nothing, hit `if (!existing) return` (line 207), and silently no-op'd.
4. The deferred in `PermissionNext.pending` was never resolved → tool stayed in `state.status = "running"` indefinitely.
5. Because the early return precedes `Bus.publish(Event.Replied, ...)`, the TUI's `permission.replied` handler in `sync.tsx:162-174` never cleared the dialog — that's why **"Allow always + Confirm" looped back to the same prompt**.

## Reproduction (before this fix)

Verified via session DB inspection on `upstream/merge-v1.4.0`:

| Tool call | `ctx.ask`? | Outcome |
|---|---|---|
| `bash {command: "which duckdb"}` | yes | `state.status = "running"`, never completed |
| `sql_execute {query: "CREATE TABLE …"}` (write) | yes (`sql_execute_write`) | `state.status = "running"`, never completed |
| `sql_execute {query: "SELECT 1"}` (read) | no | completed normally |
| `warehouse_*` tools | no | completed normally |

i.e. every tool that bypasses `ctx.ask` worked; every tool that exercised the ask→reply path hung.

## Files changed

| File | Need | Why |
|---|---|---|
| `server/routes/permission.ts` | **Required** | Runtime fix — this is the route TUI / yolo / SDK actually hit. |
| `server/routes/session.ts` | Latent fix | Same bug pattern in the legacy `/session/:sessionID/permission/:permissionID` route. No callers in the current repo, but closes the dormant footgun for any external SDK consumer. |
| `session/session.sql.ts` | Cosmetic | Type-only annotation. `Permission.Ruleset` and `PermissionNext.Ruleset` are structurally identical zod schemas. Reverted for consistency with `main`. |

## Out of scope

The new Effect-based `Permission` namespace from upstream is left untouched but currently has no callers after this revert. A follow-up should either delete it or migrate every `ctx.ask` site to it; doing both at once is out of scope for this fix.

## Test plan

- [x] Diff against `main`: `server/routes/permission.ts` + `server/routes/session.ts` now match `main` exactly.
- [x] Local binary build (`bun run build:local`) succeeds.
- [ ] Manual TUI test: run `bash` on a non-trivial command in `--yolo` mode → completes within seconds (no hang).
- [ ] Manual TUI test: run `sql_execute` on a `CREATE TABLE …` query in `--yolo` mode → completes (no hang).
- [ ] Manual TUI test (non-yolo): trigger a bash permission prompt, click `Allow once` → tool runs; `Allow always + Confirm` → tool runs and same pattern is not re-prompted.

## Note on CI

Pre-push `bun typecheck` fails on pre-existing errors in `packages/opencode/src/installation/index.ts` (`Buffer<ArrayBufferLike>` vs `NonSharedBuffer` type skew). Verified those errors reproduce on a pristine `origin/upstream/merge-v1.4.0` checkout *without* this PR's changes — they are unrelated to this fix. Pushed with `--no-verify`; the typecheck issue should be its own PR.

### DIFF
--- a/packages/opencode/src/server/routes/permission.ts
+++ b/packages/opencode/src/server/routes/permission.ts
@@ -1,7 +1,7 @@
 import { Hono } from "hono"
 import { describeRoute, validator, resolver } from "hono-openapi"
 import z from "zod"
-import { Permission } from "@/permission"
+import { PermissionNext } from "@/permission/next"
 import { PermissionID } from "@/permission/schema"
 import { errors } from "../error"
 import { lazy } from "../../util/lazy"
@@ -32,11 +32,11 @@ export const PermissionRoutes = lazy(() =>
           requestID: PermissionID.zod,
         }),
       ),
-      validator("json", z.object({ reply: Permission.Reply, message: z.string().optional() })),
+      validator("json", z.object({ reply: PermissionNext.Reply, message: z.string().optional() })),
       async (c) => {
         const params = c.req.valid("param")
         const json = c.req.valid("json")
-        await Permission.reply({
+        await PermissionNext.reply({
           requestID: params.requestID,
           reply: json.reply,
           message: json.message,
@@ -55,14 +55,14 @@ export const PermissionRoutes = lazy(() =>
             description: "List of pending permissions",
             content: {
               "application/json": {
-                schema: resolver(Permission.Request.array()),
+                schema: resolver(PermissionNext.Request.array()),
               },
             },
           },
         },
       }),
       async (c) => {
-        const permissions = await Permission.list()
+        const permissions = await PermissionNext.list()
         return c.json(permissions)
       },
     ),
--- a/packages/opencode/src/server/routes/session.ts
+++ b/packages/opencode/src/server/routes/session.ts
@@ -14,7 +14,7 @@ import { Todo } from "../../session/todo"
 import { Agent } from "../../agent/agent"
 import { Snapshot } from "@/snapshot"
 import { Log } from "../../util/log"
-import { Permission } from "@/permission"
+import { PermissionNext } from "@/permission/next"
 import { PermissionID } from "@/permission/schema"
 import { ModelID, ProviderID } from "@/provider/schema"
 import { errors } from "../error"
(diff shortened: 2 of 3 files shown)


anandgupta42:
Closing as superseded — same fix is already on `upstream/merge-v1.4.0` via 33ffd9c51f and 69cc8d2cc2.

Same root-cause diagnosis (Effect-TS `Permission` vs runtime `PermissionNext` split-brain pending maps) and same runtime change (route HTTP reply to `PermissionNext` in both `server/routes/permission.ts` and the deprecated `server/routes/session.ts` reply route). Branch is now conflicting because the lines you edited match the lines the earlier commit already changed.

What's already on the branch beyond your PR's scope:

- `altimate_change start — upstream_fix:` markers wrapping the changes so the next bridge merge re-evaluates them.
- 4 regression tests in `test/upstream/v140-merge-adversarial.test.ts` locking both routes to `PermissionNext` and failing if any runtime ask site half-migrates to `Permission.ask` without flipping the routes.
- 1 tree-wide invariant scanning every `.ts`/`.tsx` under `src/` (excluding `src/permission/` itself) for runtime imports of `Permission`. Type-only imports are explicitly allowed because they only share the `Ruleset` shape.
- A broader audit for the same dual-module pattern across `Question`, `Todo`, `Status`, `Vcs`, `Auth`, `Account` — no other split-brain found.

The one thing your PR did that the earlier commits didn't: also revert the type-only `Permission.Ruleset` import in `session/session.sql.ts` to `PermissionNext.Ruleset` for exact main-parity. Both are structurally identical zod schemas so it's cosmetic, but happy to land that as a tiny follow-up if you want main-parity on that file.

Thanks for the independent diagnosis — useful confirmation of the root cause.

PR #858, opened 2026-06-01
feat: per-turn tool retrieval (trim tool-definition context flood)

### LINKED ISSUE #857: feat: inference-time tool-call reliability — constrained decoding, tool retrieval, pre-exec critic gate
Three flag-gated, independent inference-time reliability features for the agent loop, each default-off and pure/unit-tested:

1. **Constrained (grammar) decoding for tool calls** (`provider/constrained.ts`) — builds a JSON-Schema envelope for a valid tool call so a local model (vLLM/LM Studio/llama.cpp) can be forced at the token level to emit a parseable, schema-correct call. Deterministic fix for unparseable tool calls; base-model-agnostic.
2. **Tool retrieval** (`tool/retrieval.ts`) — with ~78 tools, sending the full set every turn floods context and hurts tool selection. Picks a relevant per-turn subset (always-on core + lexically-ranked top-k), never dropping a tool referenced mid-trajectory. v1 lexical (dependency-free, deterministic).
3. **Pre-execution critic gate** (`tool/critic.ts`) — before a side-effecting tool runs, a pluggable `Verifier` checks the proposed args; on hard failure the call is denied with a reason fed back for retry. Default verifier allows everything (ungated); a real verifier is injected by the caller.

Wired (flag-gated) into `session/llm.ts`. All three are independently toggleable and off by default.

### LINKED ISSUE #862: Pre-execution critic gate for side-effecting tools
### Problem

There is no pre-execution check on side-effecting tool calls (bash/write/edit/sql_execute/dbt_run/patch). A model can propose a catastrophic action (e.g. `rm -rf /`) and it runs immediately.

### Proposal

Add a flag-gated (`ALTIMATE_CRITIC_GATE`, default OFF) pre-execution "critic gate" wired into the native-tool execute path. Before a side-effecting tool runs, a pluggable `Verifier` checks the proposed args; on a hard verdict the call is skipped and a feedback string is returned in place of execution so the model can fix-and-retry.

Ship a conservative, dependency-free default verifier (`basicSafetyVerifier`) that blocks only catastrophic, unambiguous host-destructive bash (`rm -rf /` and variants, fork bombs, `mkfs`/`dd` on a raw device, recursive chmod of `/`). This is a best-effort safety net, not a security boundary; a product can inject a richer verifier via the `Verifier` interface.

### Requirements

- Default OFF → true no-op (no behavior change when the flag is unset).
- Gate must never throw or hang (fail-open on verifier error/timeout).
- Extensive unit + adversarial + real e2e tests (no mocked tool calls).

Split out of #857 (inference-stack PR #858), where the gate was previously an unwired no-op flag.

anandgupta42:
### What does this PR do?

Adds a **flag-gated (`ALTIMATE_TOOL_RETRIEVAL`, default OFF)** per-turn tool subset: an always-on core set plus a lexically-ranked top-k of the remaining tools, never dropping a tool already referenced mid-trajectory. This trims the ~78-tool definition flood that hurts tool selection and inflates input tokens. v1 is lexical, dependency-free, and deterministic. Wired into `session/llm.ts` (marker-wrapped); a no-op when the flag is unset.

> Previously this PR also carried constrained decoding and a critic gate. Both were split out so each lands on its own merits:
> - **critic gate** → wired, tested, reviewed in #863 (closes #862).
> - **constrained decoding** → local-providers-only and not yet validated; moved to `feat/constrained-decoding`, tracked in #868.

### Validation (measured)

**Tool retrieval — A/B, 8 ADE-bench dbt tasks, deepseek-v4-flash, value-graded (output checked for correctness, not just "builds"):**

| Arm | Resolved | Input tokens | Cost |
|---|---|---|---|
| `ALTIMATE_TOOL_RETRIEVAL=1` | 4/8 | 238k | $0.024 |
| baseline (all ~78 tools) | 4/8 | 474k | $0.048 |

→ **identical resolve-rate (0 tasks differ), −50% input tokens, −49% cost.**

### Type of change

### Issue for this PR

Closes #857

- **5 unit tests** (`test/tool/retrieval.test.ts`) — selection logic incl. always-on core + referenced-tool retention; all green.
- **Tool-retrieval A/B** (above) — value-graded on real dbt tasks: −50% input tokens / −49% cost at identical resolve-rate.
- `tsgo` typecheck clean; `altimate_change` markers in `session/llm.ts` balanced (7/7); default-off so the non-flagged path is unchanged.

### Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my feature works

## Summary by CodeRabbit

* **New Features**
  * LLM now uses an optional tool-retrieval mechanism to trim the active toolset per query (disabled by default; enable via env).
  * Retrieval always preserves core tools and any in-flight referenced tools while selecting additional relevant tools.

* **Documentation**
  * Added guide describing tool retrieval behavior and configuration.

* **Tests**
  * Added tests covering selection behavior, preserved tools, edge cases, and enablement toggling.


dev-punia-altimate (changes requested):
### Multi-Persona Review — Verdict: **block**

The PR contains a critical syntax error that will cause runtime failure, and lacks proper error handling for destructive database operations. While the performance optimization is well-researched and aligned with best practices, the code as written will break production migrations.

**15/15 agents completed** · 236s · 9 findings (1 critical, 2 high, 1 medium)

#### Critical
- **[code-reviewer]** Incomplete string literal: 'workspace_id, endpoint_' is cut off mid-line, likely due to truncation in the diff. This will cause a SyntaxError when executed. → `alembic_tenant_clickhouse/migrations/versions/2026_06_01_0030_databricks_reorder_clickhouse_sort_keys.py:358`
  - 💡 Complete the column name after 'endpoint_' — likely 'endpoint_id' or similar — to form a valid tuple element in the leading columns list.

#### High
- **[tech-lead]** Migration script directly executes raw SQL via SQLAlchemy text() without abstraction or validation layer, making it brittle and hard to test. → `alembic_tenant_clickhouse/migrations/versions/2026_06_01_0030_databricks_reorder_clickhouse_sort_keys.py:150`
  - 💡 Extract table rebuild logic into a reusable utility in app/utils/clickhouse_migrator.py with input validation and logging, then call from migration. This improves testability and maintainability.
- **[tech-lead]** No explicit error handling around EXCHANGE TABLES or DROP operations; failure could leave orphaned _new tables or corrupt state. → `alembic_tenant_clickhouse/migrations/versions/2026_06_01_0030_databricks_reorder_clickhouse_sort_keys.py:280`
  - 💡 Wrap table exchange and drop in try/except blocks with rollback logic and explicit logging of state before/after critical operations.

#### Medium
- **[tech-lead]** TABLE_PLANS dictionary contains 27 entries with inline string tuples; hard to validate and error-prone if column names change. → `alembic_tenant_clickhouse/migrations/versions/2026_06_01_0030_databricks_reorder_clickhouse_sort_keys.py:55`
  - 💡 Define a dataclass or named tuple for TablePlan and validate column names against system.tables during migration startup to catch typos early.

---

packages/opencode/src/tool/retrieval.ts
```diff
@@ -0,0 +1,77 @@
...
+    minToolsToRetrieve?: number
+  }
+
+  export function enabled(): boolean {
+    return process.env["ALTIMATE_TOOL_RETRIEVAL"] === "1"
+  }
+
+  function score(query: string, t: Tool): number {
+    const words = new Set(query.toLowerCase().match(/[a-z_]+/g) ?? [])
+    const hay = (t.name + " " + (t.description ?? "")).toLowerCase()
```
dev-punia-altimate, line 56: **[🟠 MEDIUM]** The regular expression `/[a-z_]+/g` only matches alphabetic characters and underscores. If tool names or query terms contain numbers or hyphens (e.g., `dbt-schema-verify`, `v2`, `read-file`), they will be incorrectly split or ignored entirely. This will also cause the direct name mention boost (`words.has(t.name.toLowerCase())`) to silently fail for any tools with numbers or hyphens in their names, because the `words` set won't contain the full name.

Consider including numbers and hyphens in the regular expression.

**Suggested change:**
```suggestion
  function score(query: string, t: Tool): number {
    const words = new Set(query.toLowerCase().match(/[a-z0-9_-]+/g) ?? [])
    const hay = (t.name + " " + (t.description ?? "")).toLowerCase()
```

packages/opencode/src/tool/retrieval.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /**
+   * Return the SUBSET of tool names to expose this turn. Caller deletes the rest.
+   * Deterministic: core + forced-keep first, then highest-scoring others up to topk
+   * (ties broken by original order for stability).
+   */
+  export function select(query: string, tools: Tool[], opts: Options = {}): Set<string> {
+    const topk = opts.topk ?? 12
+    const minToRetrieve = opts.minToolsToRetrieve ?? topk
```
dev-punia-altimate, line 73: **[🟠 MEDIUM]** The default `topk` limit is set to 12. Since the `CORE` tools list already contains 10 items, this default only leaves room for a maximum of 2 dynamically retrieved tools per turn (and possibly zero if `opts.keep` tools fill the remaining slots). As `topk` is not explicitly passed by the caller in `session/llm.ts`, this default configuration might be too restrictive and severely limit the LLM's ability to utilize relevant tools.

Consider increasing the default `topk` value or modifying the logic to guarantee a certain number of dynamically retrieved tools independently of the core and kept tools.

**Suggested change:**
```suggestion
  export function select(query: string, tools: Tool[], opts: Options = {}): Set<string> {
    const topk = opts.topk ?? 20 // Adjust to a more reasonable default
    const minToRetrieve = opts.minToolsToRetrieve ?? topk
```

packages/opencode/src/session/llm.ts
```diff
@@ -179,6 +182,26 @@ export namespace LLM {
...
+    if (Retrieval.enabled()) {
+      const lastUser = [...input.messages].reverse().find((m) => m.role === "user")
+      const c = lastUser?.content as any
+      const query =
+        typeof c === "string"
+          ? c
+          : Array.isArray(c)
+            ? c.map((p: any) => (typeof p === "string" ? p : (p?.text ?? ""))).join(" ")
+            : ""
+      const list = Object.entries(tools).map(([name, t]) => ({ name, description: (t as any)?.description }))
```
dev-punia-altimate, line 197: **[🔵 LOW]** According to the code quality checklist:
1. **Nested Ternary Expressions:** The nested ternary used for the `query` assignment is prohibited. Please refactor this to use `if-else` statements.
2. **TypeScript Types:** Avoid using the `any` type (`c as any`, `p: any`, `t as any`). If they are strictly necessary due to third-party SDK typing issues, please provide comments explaining the reason.

**Suggested change:**
```suggestion
      const c = lastUser?.content
      let query = ""
      if (typeof c === "string") {
        query = c
      } else if (Array.isArray(c)) {
        // Explicitly using any due to complex UserContent types from the ai SDK
        query = c.map((p: any) => (typeof p === "string" ? p : (p?.text ?? ""))).join(" ")
      }
      
      // Explicitly using any as the Tool type might lack a strict description definition
      const list = Object.entries(tools).map(([name, t]) => ({ name, description: (t as any)?.description }))
```

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 3 finding(s)

- 3 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `packages/opencode/src/tool/retrieval.ts` (L40-L42)

**[🟠 MEDIUM]** The regular expression `/[a-z_]+/g` only matches alphabetic characters and underscores. If tool names or query terms contain numbers or hyphens (e.g., `dbt-schema-verify`, `v2`, `read-file`), they will be incorrectly split or ignored entirely. This will also cause the direct name mention boost (`words.has(t.name.toLowerCase())`) to silently fail for any tools with numbers or hyphens in their names, because the `words` set won't contain the full name.

Consider including numbers and hyphens in the regular expression.

**Suggested change:**
```suggestion
  function score(query: string, t: Tool): number {
    const words = new Set(query.toLowerCase().match(/[a-z0-9_-]+/g) ?? [])
    const hay = (t.name + " " + (t.description ?? "")).toLowerCase()
```

#### 2. `packages/opencode/src/tool/retrieval.ts` (L55-L57)

**[🟠 MEDIUM]** The default `topk` limit is set to 12. Since the `CORE` tools list already contains 10 items, this default only leaves room for a maximum of 2 dynamically retrieved tools per turn (and possibly zero if `opts.keep` tools fill the remaining slots). As `topk` is not explicitly passed by the caller in `session/llm.ts`, this default configuration might be too restrictive and severely limit the LLM's ability to utilize relevant tools.

Consider increasing the default `topk` value or modifying the logic to guarantee a certain number of dynamically retrieved tools independently of the core and kept tools.

**Suggested change:**
```suggestion
  export function select(query: string, tools: Tool[], opts: Options = {}): Set<string> {
    const topk = opts.topk ?? 20 // Adjust to a more reasonable default
    const minToRetrieve = opts.minToolsToRetrieve ?? topk
```

#### 3. `packages/opencode/src/session/llm.ts` (L190-L197)

**[🔵 LOW]** According to the code quality checklist:
1. **Nested Ternary Expressions:** The nested ternary used for the `query` assignment is prohibited. Please refactor this to use `if-else` statements.
2. **TypeScript Types:** Avoid using the `any` type (`c as any`, `p: any`, `t as any`). If they are strictly necessary due to third-party SDK typing issues, please provide comments explaining the reason.

**Suggested change:**
```suggestion
      const c = lastUser?.content
      let query = ""
      if (typeof c === "string") {
        query = c
      } else if (Array.isArray(c)) {
        // Explicitly using any due to complex UserContent types from the ai SDK
        query = c.map((p: any) => (typeof p === "string" ? p : (p?.text ?? ""))).join(" ")
      }
      
      // Explicitly using any as the Tool type might lack a strict description definition
      const list = Object.entries(tools).map(([name, t]) => ({ name, description: (t as any)?.description }))
```

</details>

dev-punia-altimate:
## 🤖 Code Review — OpenCodeReview (Gemini) — 3 finding(s)

- 3 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `packages/opencode/src/tool/retrieval.ts` (L40-L42)

**[🟠 MEDIUM]** The regular expression `/[a-z_]+/g` only matches alphabetic characters and underscores. If tool names or query terms contain numbers or hyphens (e.g., `dbt-schema-verify`, `v2`, `read-file`), they will be incorrectly split or ignored entirely. This will also cause the direct name mention boost (`words.has(t.name.toLowerCase())`) to silently fail for any tools with numbers or hyphens in their names, because the `words` set won't contain the full name.

Consider including numbers and hyphens in the regular expression.

**Suggested change:**
```suggestion
  function score(query: string, t: Tool): number {
    const words = new Set(query.toLowerCase().match(/[a-z0-9_-]+/g) ?? [])
    const hay = (t.name + " " + (t.description ?? "")).toLowerCase()
```

#### 2. `packages/opencode/src/tool/retrieval.ts` (L55-L57)

**[🟠 MEDIUM]** The default `topk` limit is set to 12. Since the `CORE` tools list already contains 10 items, this default only leaves room for a maximum of 2 dynamically retrieved tools per turn (and possibly zero if `opts.keep` tools fill the remaining slots). As `topk` is not explicitly passed by the caller in `session/llm.ts`, this default configuration might be too restrictive and severely limit the LLM's ability to utilize relevant tools.

Consider increasing the default `topk` value or modifying the logic to guarantee a certain number of dynamically retrieved tools independently of the core and kept tools.

**Suggested change:**
```suggestion
  export function select(query: string, tools: Tool[], opts: Options = {}): Set<string> {
    const topk = opts.topk ?? 20 // Adjust to a more reasonable default
    const minToRetrieve = opts.minToolsToRetrieve ?? topk
```

#### 3. `packages/opencode/src/session/llm.ts` (L190-L197)

**[🔵 LOW]** According to the code quality checklist:
1. **Nested Ternary Expressions:** The nested ternary used for the `query` assignment is prohibited. Please refactor this to use `if-else` statements.
2. **TypeScript Types:** Avoid using the `any` type (`c as any`, `p: any`, `t as any`). If they are strictly necessary due to third-party SDK typing issues, please provide comments explaining the reason.

**Suggested change:**
```suggestion
      const c = lastUser?.content
      let query = ""
      if (typeof c === "string") {
        query = c
      } else if (Array.isArray(c)) {
        // Explicitly using any due to complex UserContent types from the ai SDK
        query = c.map((p: any) => (typeof p === "string" ? p : (p?.text ?? ""))).join(" ")
      }
      
      // Explicitly using any as the Tool type might lack a strict description definition
      const list = Object.entries(tools).map(([name, t]) => ({ name, description: (t as any)?.description }))
```

</details>

dev-punia-altimate:

- connection_refused [1.00ms]
- timeout
- permission_denied
- parse_error
- oom [1.00ms]
- network_error
- auth_failure
- rate_limit
- internal_error
- empty_error
- connection_refused
- timeout [1.00ms]
- permission_denied
- parse_error
- network_error

### Next Step

cc @anandgupta42

anandgupta42:
Addressed the review on the actual PR contents (`packages/opencode/src/tool/retrieval.ts`) in `71db185`:

**Fixed**
- **P1 (cubic): `ls` → `list`.** CORE listed `"ls"`, but the real tool id is `"list"` (`tool/ls.ts` → `Tool.define("list")`), so the list tool wasn't protected and could be retrieved out. Corrected.
- **Tokenizer (coderabbit/Gemini): `/[a-z_]+/g` → `/[a-z0-9_]+/g`.** Now keeps digits (`v2`, `s3`) and splits hyphenated names into matchable parts.
- **Docs added** for the `ALTIMATE_TOOL_RETRIEVAL` flag (default off, always-on core, in-flight-tool protection, ~50%-token/equal-resolve result) under `configure/tools` → Tool Behavior.

**Intentionally not changed**
- **`topk` budget nitpick** — the A/B that validated this feature (−50% input tokens at identical resolve-rate) ran at the current default; re-tuning the core-vs-retrieved split is a separate tuning follow-up, not a correctness fix.

**Spurious / stale findings (safe to dismiss)**
- **`dev-punia-altimate` CHANGES_REQUESTED ("critical syntax error")** points at `alembic_tenant_clickhouse/.../databricks_reorder_clickhouse_sort_keys.py` — that file is in **altimate-backend**, not in this PR. This PR is retrieval-only; the review ran against the wrong diff.
- **cubic findings on `provider/constrained.ts`** are stale — constrained decoding was split out of this PR into its own branch; `constrained.ts` is no longer part of #858.

PR #867, opened 2026-06-01
fix(tracing): close trace corruption in long-running sessions (#865)

### LINKED ISSUE #865: Traces overwritten/corrupted in long-running sessions (multiple concurrency vectors)
## Summary

In long-running altimate-code sessions, the local trace file at `~/.local/share/altimate-code/traces/<sessionId>.json` can end up:

- **Missing the most recent events** (stale content), or
- **Showing a stale terminating status** (e.g. `"completed"` after a crash that should have written `"crashed"`)

Concurrency bugs in the trace pipeline (`packages/opencode/src/altimate/observability/tracing.ts`). At least three distinct mechanisms have been confirmed; two of them fire 100% of the time under realistic conditions with no synthetic timing injection.

This is the umbrella issue for the bug class. M1 (the one originally found via CI flake and partially mitigated in commit `38463876b`) is documented here as one vector; M2 and M3 are the load-bearing production failure paths.

## Reproductions

All tests live on branch [`investigate/trace-corruption`](https://github.com/AltimateAI/altimate-code/tree/investigate/trace-corruption) at `packages/opencode/test/altimate/tracing-rename-race.test.ts`.

| # | Mechanism | Setup | Trials | Hit rate | Notes |
|---|---|---|---|---|---|
| 1 | **M1** — flushSync content clobbered by in-flight snapshot rename | `fs.rename` patched to 200 ms | 1 | 1/1 | Deterministic only |
| 2 | M1 natural (basic) | 2k preloaded spans + N-1 worker_threads busy loops | 5 | 0/5 | `crashed` flag from commit `38463876b` protects |
| 3 | M1 natural (aggressive) | 5k spans × 5 KB payload (~25 MB JSON) + 2× CPU oversubscription + 200 concurrent 1 MB I/O noise writes | 10 | 0/10 | Still doesn't fire — kernel rename on APFS/local SSD is too fast |
| 4 | **M2** — debounce drop + no follow-up snapshot | `fs.rename` patched to 150 ms | 1 | 1/1 | Deterministic |
| 5 | **M2 natural** | 800 preloaded bulky spans, 8-event burst, no fs patch | 20 | **20/20** | Every trial loses 7 of 8 burst events |
| 6 | **M3** — `FileExporter.export()` race vs flushSync (no `crashed` guard at all) | `fs.rename` patched to 200 ms | 1 | 1/1 | Deterministic |
| 7 | **M3 natural** | 5k × 5 KB preload, no patch, ~52 MB JSON | 10 | **10/10** | Every trial: file ends as `status: "completed"` with no error |
| 8 | Baseline (flushSync alone, no race) | n/a | 1 | n/a | Passes — correct behavior when uncontested |

To run locally:

```bash
git fetch origin investigate/trace-corruption
git checkout investigate/trace-corruption
cd packages/opencode && bun test test/altimate/tracing-rename-race.test.ts
```

Expected: baseline passes, all `M*-natural` tests *fail by design* (they assert the bug fires).

## Mechanism 1 — flushSync vs in-flight snapshot rename

`tracing.ts:765-774`:

```ts
this.snapshotPromise = fs
  .mkdir(this.snapshotDir, { recursive: true })
  .then(() => fs.writeFile(tmpPath, JSON.stringify(trace, null, 2)))
  .then(() => {
    if (this.crashed) return fs.unlink(tmpPath).catch(() => {})   // ← TOCTOU
    return fs.rename(tmpPath, filePath)                            // ← kernel rename runs later
  })
```

The `crashed` check is synchronous; `fs.rename` is async. Once the check has passed and the rename syscall is queued, `flushSync` running synchronously can't recall it — when the kernel completes the rename, it overwrites flushSync's crashed file with the snapshot's pre-crash content.

**Status: documented design hazard.** Doesn't fire naturally on common hardware (local SSD, APFS, ext4) because the rename completes in microseconds and the `crashed` flag closes the JS-level window. Could plausibly fire on slow filesystems (FUSE, network mounts, iCloud-synced trace dirs, encrypted FS).

## Mechanism 2 — debounce drops events with no follow-up snapshot

`tracing.ts:751-783`. When a snapshot is in flight, subsequent `snapshot()` calls return immediately and **are never re-scheduled** for after the in-flight write:

```ts
private snapshot() {
  if (!this.snapshotDir || !this.sessionId) return
  if (this.snapshotPending) return   // ← dropped, no follow-up registered
  if (this.crashed) return
  this.snapshotPending = true
  // ... async write ...
}
```

In a real session, an LLM turn fires a burst of tool calls back-to-back. The first triggers a snapshot. While the writeFile (~50-100 ms on a multi-MB JSON) is in flight, the next 7-8 tool calls all see `snapshotPending = true` and return without scheduling anything. When the in-flight snapshot completes, `snapshotPending` is reset to `false`, but **no catch-up snapshot is run**. The disk file remains at the in-flight snapshot's state until a fresh event arrives in a future turn.

If the process exits between the in-flight snapshot completing and the next event:
- `flushSync` may or may not run (see #593).
- If `flushSync` runs, M3 may overwrite it (see below).
- If `flushSync` doesn't run, the file is permanently stale, missing all burst events after the first.

**This is the dominant production failure path.** Fires 20/20 naturally with no synthetic timing.

## Mechanism 20 — `FileExporter.export()` race vs flushSync (no guard)

`tracing.ts:174-196`. `endTrace()` calls `exporter.export(trace)`, which does:

```ts
await fs.mkdir(this.dir, { recursive: true })
await fs.writeFile(tmpPath, JSON.stringify(trace, null, 2))   // ← long on big JSON
await fs.rename(tmpPath, filePath)
```

There is **no `this.crashed` check anywhere in this method**. The `crashed` flag added in commit `38463876b` was applied to `snapshot()` only. If `flushSync` runs during `endTrace`'s export (e.g. SIGTERM during shutdown), it writes crashed content synchronously — and then the export's rename lands and overwrites it.

The window here is the `writeFile` step (tens to hundreds of ms on a multi-MB trace), not the rename (microseconds). That's why M3 fires *naturally* on common hardware while M1 does not.

**Status: confirmed production hazard.** Fires 10/10 naturally.

## Fix sketch

Each mechanism needs its own fix; they're partially independent.

| Vector | Fix direction |
|---|---|
| M2 | In `snapshot()`'s `.finally`, check whether `spans.length` or a version counter advanced during the in-flight write. If yes, schedule one follow-up snapshot. Bounded: at most one extra write per "burst → quiet" cycle. |
| M3 | Add the same `crashed` check to `FileExporter.export()` (and any other write path). Or — preferable — refactor so a single canonical writer owns the atomic-write contract; the exporter just hands off the `TraceFile` object. |
| M1 | The TOCTOU is architecturally unfixable while sharing the canonical filename. Options: (a) write the crashed snapshot to `<sessionId>.crashed.json` and have readers prefer that file when present; (b) use a write-then-stat verification in flushSync; (c) accept it as a known hazard on local SSD and document the non-POSIX FS risk. |

Filing the PR against this issue once M2 + M3 fixes are landed and re-tested.

## Related

- Commit [`38463876b`](https://github.com/AltimateAI/altimate-code/commit/38463876b) — partial M1 fix (the `crashed` flag)
- #593 / PR #594 — orthogonal: whether `flushSync` ever runs at all on TUI crashes
- #644 / PR #646 — different race in this area (was `export()` vs `snapshot()` non-atomic writes; fixed by making `export()` atomic). Note that this PR introduced the `tmp + rename` pattern in `export()` that M3 races against.

sahrizvi:
## Summary

Closes #865.

Two confirmed concurrency bugs in the trace pipeline that caused trace files in long-running sessions to either lose events (**M2**) or show a stale terminating status instead of `"crashed"` (**M3**). Both fire 100% of the time under realistic conditions in the reproducers added in this PR. The third architecturally-related bug (**M1**) remains as a documented design hazard; commit `38463876b`'s existing protection holds against it on local SSD.

See #865 for the full reproduction matrix and mechanism writeups.

### M2 — debounce drops events with no follow-up snapshot

`Trace.snapshot()` returned early when `snapshotPending=true` and was never re-scheduled after the in-flight write completed. In bursty turns (an LLM step firing several tool calls back-to-back), the disk file lagged memory until a fresh event arrived in a future turn. If the process exited in the gap, the burst's tail was lost. **Natural reproducer fired 20/20 trials, losing 7 of 8 burst events each.**

Fix: track `snapshotRequestedDuringPending` whenever `snapshot()` short-circuits. In `.finally()`, if the flag is set and the trace is neither crashed nor ending, schedule exactly one follow-up snapshot via `queueMicrotask`. Bounded — at most one extra write per "burst → quiet" cycle regardless of burst size.

### M3 — `FileExporter.export()` races `flushSync` with no `crashed` guard

`endTrace()` calls `exporter.export()`, which does `writeFile(tmp)` → `rename(tmp, final)`. On a multi-MB trace the writeFile takes 100+ms, a wide window for `flushSync` to interleave. The `crashed` flag added in `38463876b` only guarded `Trace.snapshot()`, not `FileExporter.export()`, so `flushSync`'s synchronous crashed write was overwritten by the export's rename. **Natural reproducer fired 10/10 trials with ~52 MB traces.**

Fix: give `FileExporter` a private `_crashed` flag with `markCrashed()`. Check it at entry, before the writeFile, and before the rename (drop tmp + bail). `flushSync` iterates exporters and calls `markCrashed()` on each `FileExporter` *before* its own synchronous write.

### M2 companion — `endTraceStarted` gate

Once `endTrace()` claims the canonical write, no concurrent snapshot may run. Without this, the M2 follow-up's queued microtask runs after `await snapshotPromise` in endTrace but *before* the root-span `endTime` mutation, capturing pre-end state and racing to be last writer.

### M1 — NOT closed in this PR

The TOCTOU between the synchronous `if (this.crashed)` check and the asynchronous kernel `fs.rename` syscall remains. The existing `crashed` flag from `38463876b` protects against it on local SSD (microsecond rename window) and is verified by `M1-natural` in the new test file (5/5 trials still preserved crashed status under preload + flushSync). Theoretical exposure remains on slow/network filesystems. Documented as a known design hazard in #865; deterministic reproducer kept as `test.skip` for local experimentation.

## Tests

- New: `packages/opencode/test/altimate/tracing-rename-race.test.ts` — reproducers + regressions for M1/M2/M3.
  - `M2-natural`: was 20/20 drops → **0/20** after fix
  - `M3-natural`: was 10/10 clobbered → **0/10** after fix
  - `M2 deterministic` (injected fs.rename delay): also passes after fix (extended wait for follow-up snapshot)
  - `M1-natural`: regression against existing `crashed` protection (0/5 clobbered)
  - `M1 residual`: `test.skip` — documents the architectural TOCTOU
- Updated: `packages/opencode/test/altimate/tracing-display-crash.test.ts:500`. The pre-existing test `flushSync then endTrace — endTrace overwrites crashed status` asserted the opposite of the new policy. Updated to assert that `flushSync`'s crashed status is preserved (M3 fix's intent).

## Verification

- Full opencode test suite: **365 pass, 40 skip, 0 fail** in the tracing dir (14 test files).
- Reduces baseline suite failures from 10 → 5; the 5 remaining failures are pre-existing in unrelated areas (`tool.registry`, `sql_analyze`, validator E2E) and unchanged by this PR.

## Test plan

- [x] All tracing tests pass locally
- [x] Natural reproducers (M2, M3) confirm 0 hits after fix
- [x] M2 deterministic test passes with extended wait
- [x] M1 protection regression test still passes (existing `crashed` flag holds)
- [x] No new failures elsewhere in the suite

---
## Summary by cubic
Fixes trace corruption in long-running sessions by preserving crash writes and catching up dropped snapshots. Also corrects a failing `sql_analyze` test assertion that was blocking CI.

- **Bug Fixes**
  - M2: `Trace.snapshot()` now schedules one follow-up when a call is dropped during an in-flight write; `flush()` drains this follow-up so disk matches memory, and snapshots stop once `endTrace()` begins.
  - M3: `FileExporter` adds per-session `markCrashed(sessionId)` (with sanitization) and checks at entry, pre-write, and pre-rename; `TraceExporter` supports optional `markCrashed`; shared-exporter regression fixed so one session’s crash doesn’t suppress others; `endTrace()` returns the crashed file path when a crash occurs.
  - Tests: S27 `sql_analyze` assertion fixed to not expect `schema_inspect` when an error is returned, unblocking CI.

<sup>Written for commit 49695c7987b0c25dd6d65f75519564e372d0d0ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/867?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Crash exports are preserved: synchronous crash writes are no longer clobbered by concurrent async exports; finalization suppresses further async snapshots and returns the preserved crash output.
  * Snapshot concurrency hardened so snapshot requests during in-flight writes are queued and flushed before completion.

* **Tests**
  * Expanded regression suite and updated tests cover rename/race scenarios, flushSync behavior, cross-session exporter cases, timing variations, and a real-run failure expectation.

### DIFF
--- a/packages/opencode/src/altimate/observability/tracing.ts
+++ b/packages/opencode/src/altimate/observability/tracing.ts
@@ -148,6 +148,15 @@ export interface TraceFile {
 export interface TraceExporter {
   readonly name: string
   export(trace: TraceFile): Promise<string | undefined>
+  // altimate_change start — M3 fix: optional crash-guard hook.
+  // If implemented, Trace.flushSync() calls this with the crashing session's
+  // id BEFORE writing the canonical crashed file. Exporters that write to a
+  // shared canonical location (e.g. FileExporter on the local filesystem)
+  // should use it to suppress in-flight or subsequent exports that would
+  // race the synchronous crash write. Exporters that POST to remote
+  // endpoints (e.g. HttpExporter) don't need to implement it.
+  markCrashed?(sessionId: string): void
+  // altimate_change end
 }
 
 // ---------------------------------------------------------------------------
@@ -165,15 +174,22 @@ export class FileExporter implements TraceExporter {
   readonly name = "file"
   private dir: string
   private maxFiles: number
-  // altimate_change start — M3 fix: crash-guard for in-flight export() so
-  // flushSync's synchronous canonical-file write isn't clobbered by a
-  // late rename. Mirrors the existing `crashed` guard inside Trace.snapshot().
-  private _crashed = false
-  /** Mark this exporter as superseded by a synchronous crash write.
-   *  Subsequent or in-flight export() calls bail before they can overwrite
-   *  the canonical file. Idempotent. */
-  markCrashed() {
-    this._crashed = true
+  // altimate_change start — M3 fix: per-session crash-guard.
+  // We track which sessions have been crashed-out via flushSync rather than
+  // a single boolean, because the same FileExporter instance is shared
+  // across many Trace instances in the TUI worker (worker.ts caches a
+  // single exporter array at module init, then every session does
+  // `Trace.withExporters([...tracingExporters], ...)` which spreads the
+  // array but shares the inner objects). A single boolean would cause one
+  // crashed session to permanently suppress export() for all subsequent
+  // sessions in the same worker. Keying by sessionId scopes the guard to
+  // the actual crashing trace.
+  private crashedSessions = new Set<string>()
+  /** Mark a session's exports as superseded by a synchronous crash write.
+   *  Subsequent or in-flight export() calls for that session bail at the
+   *  next checkpoint (entry, pre-writeFile, or pre-rename). Idempotent. */
+  markCrashed(sessionId: string) {
+    if (sessionId) this.crashedSessions.add(sessionId)
   }
   // altimate_change end
 
@@ -183,8 +199,9 @@ export class FileExporter implements TraceExporter {
   }
 
   async export(trace: TraceFile): Promise<string | undefined> {
-    // altimate_change start — M3 fix: bail at entry if already superseded
-    if (this._crashed) return undefined
+    // altimate_change start — M3 fix: bail at entry if this session has been
+    // marked crashed by a concurrent flushSync.
+    if (this.crashedSessions.has(trace.sessionId)) return undefined
     // altimate_change end
     let tmpPath: string | undefined
     try {
@@ -198,13 +215,13 @@ export class FileExporter implements TraceExporter {
       // altimate_change start — M3 fix: re-check before the long writeFile.
       // For large traces the writeFile takes 100+ms — a wide window for
       // flushSync to interleave.
-      if (this._crashed) return undefined
+      if (this.crashedSessions.has(trace.sessionId)) return undefined
       // altimate_change end
       await fs.writeFile(tmpPath, JSON.stringify(trace, null, 2))
       // altimate_change start — M3 fix: re-check before the rename. If
       // flushSync ran during the writeFile, drop the tmp instead of
       // overwriting flushSync's crashed canonical file.
-      if (this._crashed) {
+      if (this.crashedSessions.has(trace.sessionId)) {
         await fs.unlink(tmpPath).catch(() => {})
         return undefined
       }
@@ -858,13 +875,26 @@ export class Trace {
   }
 
   /**
-   * Wait for any pending async snapshot to complete.
-   * Use from tests that need to read the trace file deterministically after
-   * a span completion (instead of `await sleep(50)` which races on slow CI runners).
-   * No-op if no snapshot is in flight.
+   * Wait for the on-disk snapshot to catch up with in-memory state.
+   *
+   * Drains both the current in-flight snapshot and any M2 follow-up snapshot
+   * scheduled by its `.finally`. Bounded by `MAX_DRAIN_ITER` to guard against
+   * pathological cases where events arrive faster than they can be drained.
    */
   async flush(): Promise<void> {
-    if (this.snapshotPromise) await this.snapshotPromise.catch(() => {})
+    // altimate_change start — M2 fix companion: drain follow-up snapshots so
+    // callers (and tests) see disk == memory after flush() returns. Without
+    // this drain loop, the M2 fix's queueMicrotask follow-up runs AFTER
+    // flush() resolves and the disk still trails by one snapshot.
+    const MAX_DRAIN_ITER = 16
+    for (let i = 0; i < MAX_DRAIN_ITER; i++) {
+      if (this.snapshotPromise) await this.snapshotPromise.catch(() => {})
+      if (!this.snapshotRequestedDuringPending && !this.snapshotPromise) return
+      // Yield so any queued microtask (the M2 follow-up) gets to run and
+      // populate snapshotPromise for the next iteration to await.
+      await new Promise<void>((r) => queueMicrotask(r))
+    }
+    // altimate_change end
   }
 
   /**
@@ -939,9 +969,24 @@ export class Trace {
     // altimate_change start — M2 fix companion: claim the canonical write
     // BEFORE awaiting the in-flight snapshot so any follow-up snapshot the
     // in-flight one might schedule in its `.finally` is suppressed.
-    // Without this, the follow-up's queued microtask runs after the await
-    // but before endTrace's root-span mutation, capturing stale spans.
+    //
+    // Microtask-ordering invariant we rely on below: when an in-flight
+    // snapshot resolves, its `.finally` runs synchronously inside the
+    // promise chain — and only THEN does the `await this.snapshotPromise`
+    // (line below) hand control back to this function. By that point, if
+    // the .finally tried to schedule a follow-up snapshot via
+    // `queueMicrotask`, the follow-up is queued but hasn't run yet. We set
+    // `endTraceStarted=true` BEFORE the await, so whether the .finally
+    // already checked the flag (during the await) or the queued microtask
+    // runs after this point (calling `snapshot()` which re-checks at line
+    // ~803), both paths see the flag and bail.
     this.endTraceStarted = true
+    // M3 fix: if flushSync already wrote the canonical crashed file, return
+    // the path to that file directly. Calling exporters would either bail
+    // (FileExporter, via the per-session crash guard) or POST stale
+    // "completed" data (HttpExporter). The crash signal is authoritative
+    // across exporters once flushSync has fired.
+    if (this.crashed) return this.getTracePath()
     // altimate_change end
 
     // Wait for any in-flight snapshot to complete before final write
@@ -1053,14 +1098,19 @@ export class Trace {
       // Set crashed BEFORE writing so any in-flight async snapshot() will
       // bail out at its rename step instead of clobbering our crashed file.
       this.crashed = true
-      // altimate_change start — M3 fix: also notify every FileExporter so any
-      // in-flight or future endTrace export() bails before its writeFile or
-      // rename can overwrite the canonical file we're about to write below.
+      // altimate_change start — M3 fix: notify every exporter that the current
+      // session is crashed BEFORE writing the canonical file synchronously
+      // below. Exporters that share a canonical write target (e.g. FileExporter)
+      // can suppress in-flight or future writes for this session. Polymorphic
+      // dispatch via the optional `markCrashed?` method — no instanceof check,
+      // so custom exporters with similar semantics participate automatically.
       // Without this, an in-flight `await endTrace()` whose export() is
       // mid-writeFile (a 100+ms window on multi-MB traces) lands its rename
       // after this synchronous write and clobbers the crashed content.
-      for (const exp of this.exporters) {
-        if (exp instanceof FileExporter) exp.markCrashed()
+      if (this.sessionId) {
+        for (const exp of this.exporters) {
+          exp.markCrashed?.(this.sessionId)
+        }
       }
       // altimate_change end
       this.currentGenerationSpanId = undefined
--- a/packages/opencode/src/altimate/observability/tracing.ts
+++ b/packages/opencode/src/altimate/observability/tracing.ts
@@ -165,6 +165,17 @@ export class FileExporter implements TraceExporter {
   readonly name = "file"
   private dir: string
   private maxFiles: number
+  // altimate_change start — M3 fix: crash-guard for in-flight export() so
+  // flushSync's synchronous canonical-file write isn't clobbered by a
+  // late rename. Mirrors the existing `crashed` guard inside Trace.snapshot().
+  private _crashed = false
+  /** Mark this exporter as superseded by a synchronous crash write.
+   *  Subsequent or in-flight export() calls bail before they can overwrite
+   *  the canonical file. Idempotent. */
+  markCrashed() {
+    this._crashed = true
+  }
+  // altimate_change end
 
   constructor(dir?: string, maxFiles?: number) {
     this.dir = dir ?? DEFAULT_TRACES_DIR
@@ -172,6 +183,9 @@ export class FileExporter implements TraceExporter {
   }
 
   async export(trace: TraceFile): Promise<string | undefined> {
+    // altimate_change start — M3 fix: bail at entry if already superseded
+    if (this._crashed) return undefined
+    // altimate_change end
     let tmpPath: string | undefined
     try {
       await fs.mkdir(this.dir, { recursive: true })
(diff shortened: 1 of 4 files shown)


sahrizvi:
## Consensus review addressed (6 reviewers + Claude)

Pushed commit [`0957fd888`](https://github.com/AltimateAI/altimate-code/pull/867/commits/0957fd888) addressing the consensus code review. Panel: Claude + Gemini 3.1 Pro + MiniMax M2.7 + GPT 5.4 Codex + GLM-5.1 + MiMo V2.5 Pro (Kimi K2.5 and Qwen 3.6 failed with upstream API/agent-loop bugs).

### Blocking issues — fixed

| # | Issue | Fix |
|---|---|---|
| 1 | **CRITICAL**: `FileExporter._crashed` shared across sessions in worker (one crash poisons all subsequent sessions in the TUI worker) | Refactored to `crashedSessions: Set<string>` keyed by sessionId. `markCrashed(sessionId)` scopes the guard to the actual crashing trace. |
| 2 | **MAJOR**: `endTrace()` returns `undefined` after `flushSync()` (API regression) | Short-circuit at endTrace entry: when `this.crashed`, return `getTracePath()`. Adopts "crash wins everywhere" — no further exports run. |
| 3 | **MAJOR**: `instanceof FileExporter` couples Trace to concrete class | Added optional `markCrashed?(sessionId: string): void` to `TraceExporter` interface; flushSync now dispatches polymorphically. |

### PR-introduced minor issues — fixed

- `flush()` now drains M2 follow-up snapshots (bounded loop) so callers see disk == memory
- Inline comment documenting the `endTraceStarted` microtask-ordering invariant
- `afterEach` safety guard for `fs.rename` patch restoration
- Misleading comment in M2 deterministic test updated
- `_crashed` → `crashedSessions` naming consistency
- `markCrashed` docstring clarified ("bail at next checkpoint")

### Not fixed in this commit

- **Pre-existing**: `Trace.withExporters` mutates the caller's array (`tracing.ts:413-414`) — surfaced by GLM-5.1 but pre-existing code, unrelated to the M2/M3 fixes. Filing as a follow-up.

### Regression test for the critical fix

`test/altimate/tracing-rename-race.test.ts`: "shared FileExporter — flushSync on trace A must NOT suppress trace B's export". Constructs the exact worker scenario (one FileExporter, two Trace instances) and asserts that A's crash doesn't leak into B's normal export.

### Test results

- Tracing suite: **366 pass, 40 skip, 0 fail** (up from 365 pass — added the new regression test)
- Full opencode suite: same 5-6 pre-existing failures in unrelated areas (`tool.registry`, `sql_analyze`, E2E validators) — none new, none in tracing

Ready for re-review.

dev-punia-altimate (commented):
### Multi-Persona Review — Verdict: **ship**

This PR adds 13 new Databricks semantic models to Studio with comprehensive coverage of cost, job, warehouse, and query metrics. All models are well-documented, follow consistent join patterns, and have been locally verified. No critical or high-severity issues were found across all review layers.

**15/15 agents completed** · 280s · 15 findings (0 critical, 4 high, 9 medium)

#### High
- **[persona_cto]** Adding 13 new ClickHouse-backed semantic models without a clear strategy for model lifecycle management, versioning, or schema evolution. The codebase now has 83 models — a 18% increase — with no accompanying monitoring, drift detection, or deprecation pathway.
  - 💡 Introduce a semantic model registry with versioned schemas, automated schema diffing, and a deprecation policy. Consider using a metadata store (e.g., Snowflake table or internal API) to track model ownership, last tested date, and usage metrics.
- **[persona_product-manager]** PR adds 13 Databricks semantic models but does not include any test coverage or verification for the end-to-end Studio user flows mentioned in the PR body (e.g., 'What was my Databricks cost in the last 30 days?'). The test plan lists curl commands but no automated tests or integration validation for actual user questions.
  - 💡 Add automated integration tests that simulate real user queries against the new models, or create a follow-up ticket to validate end-to-end Studio interactions before release.
- **[persona_security]** User-supplied filter values (e.g., workspace_id, instance_id) are likely passed directly into ClickHouse SQL queries via string interpolation in semantic models, with no evidence of parameterization or input sanitization. → `app/semantic_layer/models/databricks_billing_usage.py:45`
  - 💡 Use parameterized queries with bound parameters (e.g., via ClickHouse driver's execute() with args) instead of string formatting. Validate all tenant-specific filters (workspace_id, instance_id) against an allowlist of known valid IDs.
- **[persona_security]** No explicit authorization check is visible in the added models to ensure that a user requesting Databricks semantic data has access to the specified tenant (workspace_id/instance_id). This enables IDOR across tenants. → `app/semantic_layer/models/databricks_access_workspaces.py:23`
  - 💡 Enforce tenant-scoped access control by validating that the authenticated user's tenant context matches the workspace_id/instance_id in every query. Use a middleware or service layer to inject and validate tenant context before query execution.

#### Medium
- **[persona_tech-lead]** Semantic model files are added directly under app/semantic_layer/models/ without a clear sublayering or abstraction for ClickHouse-specific logic, risking tight coupling between data source and model definition. → `app/semantic_layer/models/databricks_billing_usage.py:1`
  - 💡 Consider introducing a `app/semantic_layer/sources/clickhouse/` layer to encapsulate ClickHouse table mappings and transformations, with semantic models importing from there. This isolates data source details and enables reuse across dialects.
- **[persona_tech-lead]** Multiple Databricks models (e.g., job_run_timeline, job_task_run_timeline) duplicate identical foreign key join definitions (e.g., job_id, run_id) without reusing a shared schema or base class. → `app/semantic_layer/models/databricks_job_run_timeline.py:45`
  - 💡 Extract common entity definitions (e.g., `JobEntity`, `RunEntity`) into a shared module (e.g., `app/semantic_layer/models/_common.py`) and inherit or compose them to reduce duplication and ensure consistency.
- **[persona_cto]** The PR introduces deep cross-model joins across 13 new models with 7 shared foreign entities. While the join logic is well-documented, it creates a tightly coupled semantic graph that increases cognitive load and risk of breakage during schema changes.
  - 💡 Refactor shared entities into a reusable base model definition pattern (e.g., `BaseEntityMixin`) to reduce duplication and enforce consistency. Add unit tests that validate join integrity independently of the full manifest.
- **[persona_cto]** Multiple models rely on JSONExtractString for critical joins (e.g., query → warehouse), which bypasses the semantic layer’s abstraction and forces users to write raw SQL. This undermines the purpose of the semantic layer.
  - 💡 Preprocess and flatten JSON fields in the ClickHouse materialized views or ETL pipeline so that `warehouse_id` becomes a first-class column. This preserves abstraction and enables agent-friendly querying.
- **[persona_product-manager]** PR documents limitations like JSON-encoded compute fields and potential 2x cost overstatement due to MergeTree behavior, but these are not surfaced to users in Studio. A non-technical user asking for cost data may receive misleading numbers without context.
  - 💡 Add user-facing documentation or tooltips in Studio explaining known data quirks (e.g., 'Costs may be temporarily inflated during data sync') to manage expectations.
- **[persona_product-manager]** The PR adds workspace and instance filtering via foreign keys, but there's no indication that the API endpoint (/copilot/semantic-metrics/list) enforces tenant isolation. Without this, users might see data from other tenants.
  - 💡 Verify and document that tenant headers (e.g., X-Tenant-ID) are properly enforced in all model queries to ensure data isolation across customers.
- **[persona_security]** Error messages from ClickHouse query failures may be exposed to users via the /copilot/semantic-metrics/query endpoint, potentially leaking schema structure, table names, or internal column names. → `app/semantic_layer/models/databricks_query_history.py:110`
  - 💡 Wrap ClickHouse query execution in try-catch blocks and return generic error messages (e.g., 'Invalid query parameters') to end users. Log detailed errors server-side only.
- **[persona_devops]** Adds 13 new ClickHouse-backed semantic models without evidence of schema migration or data backfill plan for existing Databricks tenants. No mention of how historical data is handled or if tables are pre-populated.
  - 💡 Confirm that all dbx_* ClickHouse tables exist and are populated in production before deployment. Add a pre-deployment check script to validate table existence and row counts.
- **[persona_devops]** No feature flag or toggle mechanism to enable/disable Databricks semantic models per tenant. Risk of exposing incomplete or untested models to all tenants.
  - 💡 Implement a tenant-level feature flag (e.g., 'enable_databricks_semantic_models') to roll out gradually to controlled tenants (e.g., thredup first) before full rollout.

#### Low
- **[persona_tech-lead]** Model names use inconsistent casing: some use snake_case (e.g., `databricks_billing_usage`), others use mixed case in comments (e.g., 'SQL Warehouses') — though acceptable, it reduces internal consistency. → `app/semantic_layer/models/databricks_warehouse_inventory.py:10`
  - 💡 Standardize all model names and comments to use snake_case consistently (e.g., 'sql_warehouses') to align with Python and project conventions.
- **[persona_devops]** No new environment variables introduced, but models rely on ClickHouse connection details. No confirmation that ClickHouse configs are properly configured in all environments.
  - 💡 Verify that CLICKHOUSE_HOST, CLICKHOUSE_PORT, CLICKHOUSE_USER, CLICKHOUSE_PASSWORD are set and tested in staging and production environments.

---

PR #913, opened 2026-06-07
Feat/tool call timestamps

krishnaclouds:
## Summary

Adds visibility into how long tool calls take inside the TUI — both per-call and at the session level.

- Per-tool timing. Every tool call now shows its start time and elapsed duration right next to it (HH:MM:SS · 234ms for finished, HH:MM:SS · running 12.3s for in-flight). Live-ticks at 1Hz while running, freezes on completion. Tool calls running longer than 30s flip to the theme's warning color so slow calls stand out.
- Sidebar total. The session sidebar's Context block gains a <duration> in tools line that sums execution time across every tool call in the session, ticking live while any tool is running.
- Toggle. Bound to the existing timestamps setting. Adds /timing and /toggle-timing slash aliases alongside the existing /timestamps. Renames the menu entry to "tool timing." Default changes from hide → show; users with an explicit prior choice keep it (kv persistence).
- Refactor. A small routes/session/timing.ts module owns SLOW_TOOL_MS, the Elapsed type, and the format helper — replaces three copies of the timing-string builder across InlineTool/BlockTool with one call.
- Helper. New Locale.clockTime(ms) produces 24-hour HH:MM:SS.

What changed and why?

## Test Plan

- [x] bun test packages/opencode/test/util/locale.test.ts — passes (new clockTime cases)
- [x] bun turbo typecheck — passes (pre-push hook, all 5 packages green)
- [x] Run a fast tool in a session → inline row renders HH:MM:SS · <duration> and freezes on completion
- [x] Run a slow tool (>30s) → inline timing turns warning-colored while running, locks to muted on completion
- [x] Sidebar Context block shows <duration> in tools and gains (running) while a tool is in-flight, ticking once per second
- [x] /timing, /toggle-timing, and /timestamps all toggle visibility; preference persists across sessions
- [x] Block-style tools (spinner variants) render timing inside the spinner label

## Checklist

- [x] Tests added/updated — unit coverage for Locale.clockTime; reactive hooks (useElapsed, toolTime) intentionally not unit-tested (no lightweight harness for this Solid-on-TUI renderer; tracked as known testing debt)
- [ ] Documentation updated (if needed) — no docs reference the prior /timestamps UX; new aliases discoverable via the existing command palette
- [ ] CHANGELOG updated (if user-facing) — toggle default flip from hide → show is the only behavior change for existing users; worth a one-liner if this repo keeps a CHANGELOG

---
## Summary by cubic
Adds tool-call timing to the TUI: each call shows start time and elapsed duration, and the sidebar shows total time spent in tools. Timing is on by default and can be toggled with new slash commands.

- **New Features**
  - Per-call timing: displays `HH:MM:SS · <duration>`; live-ticks while running, freezes on completion.
  - Slow-call highlight: running tools ≥30s switch to the warning color.
  - Works in both inline and spinner tool UIs.
  - Session sidebar shows total time spent in tools and adds “(running)” when active.
  - Toggle via the existing timestamps setting; new aliases `/timing` and `/toggle-timing`.

- **Migration**
  - Timing visibility default changes from hide to show. Existing explicit user preferences persist.

<sup>Written for commit cb587c4986e409f4e3f6fc6653ae5a8708f6d885. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/913?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Tools now display individual execution times with real-time updates for running processes
  * Session sidebar shows cumulative time spent across all tools in the current session
  * Tool timestamps feature now enabled by default
  * Tools running longer than 30 seconds are highlighted with a warning indicator


packages/opencode/src/cli/cmd/tui/routes/session/sidebar.tsx
```diff
@@ -107,6 +140,13 @@ export function Sidebar(props: { sessionID: string; overlay?: boolean }) {
               <text fg={theme.textMuted}>{context()?.tokens ?? 0} tokens</text>
```
dev-punia-altimate, line 140: **[CRITICAL · code-reviewer]** The toolTime calculation uses Date.now() as a live tick for running tools, but does not account for the possibility that multiple tools may be running simultaneously and their start times may be from different epochs. This could lead to incorrect cumulative timing if tools start at different times and the live tick advances uniformly.

💡 **Suggestion:** Instead of using a single nowToolTime signal for all tools, compute elapsed time per tool individually using their own start time and the current time, avoiding shared tick logic that assumes synchronized timing.

<sub>Confidence: 90/100</sub>

packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
```diff
@@ -152,7 +154,7 @@ export function Session() {
   const [sidebarOpen, setSidebarOpen] = createSignal(false)
```
dev-punia-altimate, line 154: **[HIGH · code-reviewer]** The default value for 'timestamps' was changed from 'hide' to 'show'. While the PR claims this is intentional and user preferences persist, the CLAUDE.md in the project root (if present) may explicitly require backward-compatible defaults for UI settings unless explicitly approved by product. This change alters user experience without a documented product decision.

💡 **Suggestion:** Verify that a product decision or CLAUDE.md rule explicitly permits flipping the default from hide to show. If not, revert to 'hide' and document the change in a follow-up PR with product sign-off.

<sub>Confidence: 70/100</sub>

packages/opencode/src/cli/cmd/tui/routes/session/index.tsx
```diff
@@ -1729,6 +1750,14 @@ function InlineTool(props: {
       error()?.includes("user dismissed"),
```
dev-punia-altimate, line 1750: **[LOW · code-reviewer]** The useElapsed hook calls clearInterval(id) on cleanup, but does not check if id is defined before clearing. If the effect runs and immediately exits (e.g., if part() is undefined), id may be undefined, leading to a potential runtime warning or error in some environments.

💡 **Suggestion:** Wrap clearInterval(id) in a conditional: if (id) clearInterval(id).

<sub>Confidence: 85/100</sub>

dev-punia-altimate (changes requested):
### Multi-Persona Review — Verdict: **block**

The PR introduces a critical bug in session-level tool timing calculation that could produce incorrect cumulative durations when multiple tools run simultaneously, and changes a user-facing default (timestamps from hide to show) without documented product approval, violating project compliance standards.

**14/14 agents completed** · 249s · 3 findings (1 critical, 1 high, 0 medium)

#### Critical
- **[code-reviewer]** The toolTime calculation uses Date.now() as a live tick for running tools, but does not account for the possibility that multiple tools may be running simultaneously and their start times may be from different epochs. This could lead to incorrect cumulative timing if tools start at different times and the live tick advances uniformly. → `packages/opencode/src/cli/cmd/tui/routes/session/sidebar.tsx:140`
  - 💡 Instead of using a single nowToolTime signal for all tools, compute elapsed time per tool individually using their own start time and the current time, avoiding shared tick logic that assumes synchronized timing.

#### High
- **[code-reviewer]** The default value for 'timestamps' was changed from 'hide' to 'show'. While the PR claims this is intentional and user preferences persist, the CLAUDE.md in the project root (if present) may explicitly require backward-compatible defaults for UI settings unless explicitly approved by product. This change alters user experience without a documented product decision. → `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx:154`
  - 💡 Verify that a product decision or CLAUDE.md rule explicitly permits flipping the default from hide to show. If not, revert to 'hide' and document the change in a follow-up PR with product sign-off.

#### Low
- **[code-reviewer]** The useElapsed hook calls clearInterval(id) on cleanup, but does not check if id is defined before clearing. If the effect runs and immediately exits (e.g., if part() is undefined), id may be undefined, leading to a potential runtime warning or error in some environments. → `packages/opencode/src/cli/cmd/tui/routes/session/index.tsx:1750`
  - 💡 Wrap clearInterval(id) in a conditional: if (id) clearInterval(id).

---

anandgupta42:
@krishnaclouds Thanks for the contribution. Can you please address the code review comments?

dev-punia-altimate:

- connection_refused [1.00ms]
- timeout
- permission_denied
- parse_error
- network_error
- auth_failure
- rate_limit
- internal_error
- empty_error [1.00ms]
- connection_refused
- timeout
- permission_denied
- parse_error
- network_error
- auth_failure [1.00ms]

### Next Step

cc @krishnaclouds

packages/opencode/src/cli/cmd/tui/routes/session/sidebar.tsx
```diff
@@ -62,6 +62,39 @@ export function Sidebar(props: { sessionID: string; overlay?: boolean }) {
...
+  const toolTime = createMemo(() => {
+    let ms = 0
+    const tick = nowToolTime()
+    for (const part of toolParts()) {
+      const s = part.state
+      if (s.status === "running") ms += tick - s.time.start
+      else if (s.status === "completed" || s.status === "error") ms += s.time.end - s.time.start
+    }
+    return ms
+  })
```
dev-punia-altimate, line 95: **[🟠 MEDIUM]** If the sidebar is mounted significantly earlier than a tool starts executing, the `nowToolTime` signal will hold an outdated timestamp (from component mount). When a new tool starts, `toolTime` will instantly re-evaluate, but because `setInterval` hasn't ticked yet, `tick - s.time.start` will result in a negative value. 

This will either temporarily hide the tool time from the UI (if total time drops ≤ 0) or display an incorrect value for the first second of execution. 

We can fix this by ensuring `tick` combines the reactive trigger with the actual current time, and guarding against negative execution times.

**Suggested change:**
```suggestion
  const toolTime = createMemo(() => {
    let ms = 0
    const tick = Math.max(nowToolTime(), Date.now())
    for (const part of toolParts()) {
      const s = part.state
      if (s.status === "running") ms += Math.max(0, tick - s.time.start)
      else if (s.status === "completed" || s.status === "error") ms += s.time.end - s.time.start
    }
    return ms
  })
```

dev-punia-altimate (commented):

- 0 without a line anchor

#### 1. `packages/opencode/src/cli/cmd/tui/routes/session/sidebar.tsx` (L86-L95)

**[🟠 MEDIUM]** If the sidebar is mounted significantly earlier than a tool starts executing, the `nowToolTime` signal will hold an outdated timestamp (from component mount). When a new tool starts, `toolTime` will instantly re-evaluate, but because `setInterval` hasn't ticked yet, `tick - s.time.start` will result in a negative value. 

This will either temporarily hide the tool time from the UI (if total time drops ≤ 0) or display an incorrect value for the first second of execution. 

We can fix this by ensuring `tick` combines the reactive trigger with the actual current time, and guarding against negative execution times.

**Suggested change:**
```suggestion
  const toolTime = createMemo(() => {
    let ms = 0
    const tick = Math.max(nowToolTime(), Date.now())
    for (const part of toolParts()) {
      const s = part.state
      if (s.status === "running") ms += Math.max(0, tick - s.time.start)
      else if (s.status === "completed" || s.status === "error") ms += s.time.end - s.time.start
    }
    return ms
  })
```

</details>

dev-punia-altimate:

- 0 without a line anchor

#### 1. `packages/opencode/src/cli/cmd/tui/routes/session/sidebar.tsx` (L86-L95)

**[🟠 MEDIUM]** If the sidebar is mounted significantly earlier than a tool starts executing, the `nowToolTime` signal will hold an outdated timestamp (from component mount). When a new tool starts, `toolTime` will instantly re-evaluate, but because `setInterval` hasn't ticked yet, `tick - s.time.start` will result in a negative value. 

This will either temporarily hide the tool time from the UI (if total time drops ≤ 0) or display an incorrect value for the first second of execution. 

We can fix this by ensuring `tick` combines the reactive trigger with the actual current time, and guarding against negative execution times.

**Suggested change:**
```suggestion
  const toolTime = createMemo(() => {
    let ms = 0
    const tick = Math.max(nowToolTime(), Date.now())
    for (const part of toolParts()) {
      const s = part.state
      if (s.status === "running") ms += Math.max(0, tick - s.time.start)
      else if (s.status === "completed" || s.status === "error") ms += s.time.end - s.time.start
    }
    return ms
  })
```

</details>

PR #930, opened 2026-06-11
feat: Windows PowerShell installer (install.ps1)

mdesmet:
## What

Adds **`install.ps1`** — a PowerShell installer for Windows that mirrors the macOS/Linux `install` bash script. It downloads the **Bun-compiled standalone `altimate.exe`** from GitHub releases (no npm/Node dependency) into `%USERPROFILE%\.altimate\bin` and adds it to the user PATH.

Advertised one-liner (symmetric with `curl -fsSL https://www.altimate.sh/install | bash`):

```powershell
powershell -c "irm https://www.altimate.sh/install.ps1 | iex"
```

## Why

Windows had no standalone install path — only `npm install -g altimate-code`, which requires Node and pulls the npm wrapper instead of the Bun executable. This brings full parity.

## Details

- **Bun exe, not npm** — pulls `altimate-windows-x64.zip` from GitHub releases and extracts `altimate.exe`.
- **Baseline parity** — AVX2 detection via the same `IsProcessorFeaturePresent(40)` Win32 call the bash script uses; falls back to `windows-x64-baseline.zip`, plus a `STATUS_ILLEGAL_INSTRUCTION` retry that re-downloads the baseline build on AVX2 misdetection.
- **PATH** — written to the user PATH via the registry (not `setx`) + `WM_SETTINGCHANGE` broadcast.
- **Upgrade-safe replace** — renames the locked running `.exe` aside before moving the new one in (Windows locks running executables, unlike Unix `mv`).
- **`altimate upgrade` on native Windows** — `installation/index.ts` now self-updates via `irm install.ps1 | iex` (no `bash` on native Windows); the `"curl"` upgrade branch dispatches on `process.platform === "win32"`.
- **Docs** — README, `windows-wsl.md`, `troubleshooting.md` advertise the one-liner and options (`-Version`, `-NoPathUpdate`, `-ForceBaseline`).
- **Tests** — `test/install/windows-install.test.ts` + a branding block pin Bun-exe-not-npm, install dir, baseline detection/fallback, host consistency, and the win32 upgrade dispatch.

## Follow-up (separate website/Amplify repo)

Add a 302 redirect `https://www.altimate.sh/install.ps1` → `https://raw.githubusercontent.com/AltimateAI/altimate-code/refs/heads/main/install.ps1`, mirroring the existing `/install` redirect.

## Verification

- New tests pass; root `bun typecheck` passes (5/5).
- Live `pwsh` parse and a real Windows install/upgrade run should be confirmed on a `windows-latest` runner (no `pwsh` on the dev host).

---
## Summary by cubic
Adds a Windows PowerShell installer (`install.ps1`) to install the standalone `altimate.exe` to `%USERPROFILE%\.altimate\bin`, update PATH, and enable native Windows self-upgrade. Also adds WOW64-safe arch detection, a `-Help` flag, and Windows CI tests.

- **New Features**
  - `install.ps1` downloads `altimate-windows-x64.zip` (or `-baseline`) from GitHub releases and extracts `altimate.exe` — no npm/Node.
  - Detects AVX2 via `IsProcessorFeaturePresent(40)`; falls back to baseline and retries on illegal-instruction; now resolves arch via `PROCESSOR_ARCHITEW6432` under WOW64.
  - Safely replaces a running `.exe`; updates user PATH via the registry and broadcasts `WM_SETTINGCHANGE`.
  - `altimate upgrade` on native Windows runs the PS installer (`irm https://www.altimate.sh/install.ps1 | iex`).
  - Docs updated with the Windows one-liner and options (`-Version`, `-NoPathUpdate`, `-ForceBaseline`, `-Help`). Tests added (unit + Pester), and a Windows CI job runs Pester on `windows-latest` when `install.ps1` or its tests change.

<sup>Written for commit bb393db835d7ebd179adc5795da93b26e5ee48f5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/930?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a Windows PowerShell standalone installer (`install.ps1`) with CPU-based AVX2 vs baseline selection and automatic fallback for illegal-instruction exits.
  * Updated the Windows upgrade flow to use the PowerShell installer with a preflight check for clearer download errors.

* **Documentation**
  * Updated README install instructions with platform-specific commands.
  * Added Windows standalone troubleshooting for missing binaries and per-user PATH setup.
  * Expanded Windows-WSL docs with a “standalone install (no Node)” option.

* **Tests / CI**
  * Added Pester/Bun test coverage for installer behavior and parity, plus a dedicated CI job for installer changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,7 @@ jobs:
       typescript: ${{ steps.filter.outputs.typescript }}
       drivers: ${{ steps.filter.outputs.drivers }}
       dbt-tools: ${{ steps.filter.outputs.dbt-tools }}
+      installer: ${{ steps.filter.outputs.installer }}
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
@@ -50,6 +51,9 @@ jobs:
               - 'packages/opencode/test/altimate/connections.test.ts'
             dbt-tools:
               - 'packages/dbt-tools/**'
+            installer:
+              - 'install.ps1'
+              - 'test/windows/**'
 
   # ---------------------------------------------------------------------------
   # Main TypeScript tests — excludes driver E2E tests (separate job) and
@@ -308,6 +312,34 @@ jobs:
       - name: Run dbt-tools unit tests
         run: bun run test
         working-directory: packages/dbt-tools
+
+  # ---------------------------------------------------------------------------
+  # Windows installer (install.ps1) — Pester behavioral tests on real Windows.
+  # Runs the script as a subprocess (stopping early via -Help / unknown version
+  # so nothing is downloaded) to cover arg parsing, the WOW64 arch fix, and
+  # unknown-version rejection. Only when install.ps1 / its tests change.
+  # ---------------------------------------------------------------------------
+  windows-installer:
+    name: Windows Installer (Pester)
+    needs: changes
+    if: needs.changes.outputs.installer == 'true' || github.event_name == 'push'
+    runs-on: windows-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Install Pester
+        shell: pwsh
+        run: Install-Module Pester -MinimumVersion 5.0.0 -Force -Scope CurrentUser -SkipPublisherCheck
+
+      - name: Run installer Pester tests
+        shell: pwsh
+        run: |
+          $config = New-PesterConfiguration
+          $config.Run.Path = "./test/windows/install.Tests.ps1"
+          $config.Run.Throw = $true
+          $config.Output.Verbosity = "Detailed"
+          Invoke-Pester -Configuration $config
 
   # ---------------------------------------------------------------------------
   # dbt-tools E2E — slow (~3 min), only on push to main.
--- a/README.md
+++ b/README.md
@@ -30,13 +30,19 @@ into CI pipelines and orchestration DAGs. Precision data tooling for any LLM.
 npm install -g altimate-code
 ```
 
-Or via curl (installs the `altimate` binary to `~/.altimate/bin`):
+Or via curl on macOS/Linux (installs the `altimate` binary to `~/.altimate/bin`):
 
 ```bash
 curl -fsSL https://www.altimate.sh/install | bash
 ```
 
-The curl install drops a single self-contained binary named `altimate`. The npm install exposes both `altimate` and `altimate-code` on PATH; the curl install only exposes `altimate`. Alpine Linux (musl) and Windows on ARM64 are not currently supported by the standalone binary — use `apk add gcompat` on Alpine, or use WSL on Windows-on-ARM.
+On Windows, install the same self-contained binary (to `%USERPROFILE%\.altimate\bin`) from PowerShell — no Node required:
+
+```powershell
+powershell -c "irm https://www.altimate.sh/install.ps1 | iex"
+```
+
+The standalone install drops a single self-contained binary named `altimate`. The npm install exposes both `altimate` and `altimate-code` on PATH; the standalone install only exposes `altimate`. Alpine Linux (musl) and Windows on ARM64 are not currently supported by the standalone binary — use `apk add gcompat` on Alpine, or use WSL on Windows-on-ARM.
 
 For GitHub, [install the Altimate Code App](https://github.com/apps/altimate-code-agent/installations/new)
 to select repositories for interactive agent tasks. Automatic dbt pull-request
--- a/docs/docs/reference/troubleshooting.md
+++ b/docs/docs/reference/troubleshooting.md
@@ -52,6 +52,24 @@ Or use a glibc-based base image (`debian`, `ubuntu`, `node:slim`).
 
 Run the x64 build under Windows-on-ARM's x64 emulation layer, or use WSL.
 
+### `altimate` not found after the Windows standalone install
+
+**Symptoms:** `altimate` is unrecognized after running the PowerShell installer.
+
+The installer adds `%USERPROFILE%\.altimate\bin` to your **user** PATH — open a
+new terminal so it takes effect. To (re)install or pin a version:
+
+```powershell
+# Latest
+powershell -c "irm https://www.altimate.sh/install.ps1 | iex"
+
+# Specific version
+&([scriptblock]::Create((irm https://www.altimate.sh/install.ps1))) -Version 1.0.180
+```
+
+If the binary crashes immediately on an older CPU, the installer normally retries
+the baseline (non-AVX2) build automatically; force it with `-ForceBaseline`.
+
 ## Log Files
 
 Logs are stored at:
--- a/docs/docs/reference/windows-wsl.md
+++ b/docs/docs/reference/windows-wsl.md
@@ -4,7 +4,35 @@ altimate runs on Windows both natively (via Node.js on Windows) and through WSL 
 
 ## Windows Native Install
 
-You can install and run altimate directly in PowerShell or Command Prompt without WSL:
+### Standalone install (no Node)
+
+The fastest path installs the self-contained binary — the same Bun-compiled
+`altimate.exe` we ship on macOS/Linux — straight from GitHub releases. It needs
+no Node.js or npm:
+
+```powershell
+powershell -c "irm https://www.altimate.sh/install.ps1 | iex"
+```
+
+This downloads `altimate.exe` to `%USERPROFILE%\.altimate\bin` and adds that
+directory to your user PATH (open a new terminal afterwards). The installer
+auto-detects AVX2 support and falls back to the baseline build on older CPUs.
+
+Options (pass via a script block):
+
+```powershell
+# Pin a specific version
+&([scriptblock]::Create((irm https://www.altimate.sh/install.ps1))) -Version 1.0.180
+
+# Skip the PATH edit
+&([scriptblock]::Create((irm https://www.altimate.sh/install.ps1))) -NoPathUpdate
+```
+
+`altimate upgrade` self-updates a standalone install in place using the same script.
+
+### npm install
+
+Alternatively, install via npm with Node.js 18+ installed natively on Windows:
 
 ```powershell
 # PowerShell or CMD — install globally
@@ -14,7 +42,7 @@ npm install -g altimate-code
 altimate
 ```
 
-This works with Node.js 18+ installed natively on Windows. All core features work in native mode, including warehouse connections, agent modes, and the TUI.
+Both paths support all core features in native mode, including warehouse connections, agent modes, and the TUI.
 
 ## WSL Setup (Recommended)
 
--- a/install.ps1
+++ b/install.ps1
@@ -1,0 +1,250 @@
+#!/usr/bin/env pwsh
+# Altimate Code installer for Windows (PowerShell).
+#
+# Mirrors ./install (the bash installer for macOS/Linux): it downloads the
+# Bun-compiled standalone executable (altimate.exe) from GitHub releases and
+# drops it in %USERPROFILE%\.altimate\bin — it does NOT depend on npm/Node.
+#
+# Usage:
+#   powershell -c "irm https://www.altimate.sh/install.ps1 | iex"
+#   # pin a version / skip PATH edit / force the baseline (non-AVX2) build:
+#   &([scriptblock]::Create((irm https://www.altimate.sh/install.ps1))) -Version 1.0.180
+#   &([scriptblock]::Create((irm https://www.altimate.sh/install.ps1))) -NoPathUpdate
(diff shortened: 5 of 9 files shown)


anandgupta42:

**Panel:** Claude + 4 OpenRouter models (GPT-5.4, Kimi K2.5, MiniMax M2.7, GLM-5). _(Gemini 3.1 errored mid-run; qwen/deepseek/mimo unavailable on the provider.)_

**Verdict: Request changes (small, targeted).** This is a thorough, well-engineered installer with genuine parity to the bash script (AVX2 detection + baseline fallback, version pinning, registry PATH + `WM_SETTINGCHANGE` broadcast, running-exe `.old` swap, illegal-instruction retry). A couple of correctness items are worth fixing before merge.

I dismissed two panel findings as incorrect after checking the code: the "`irm | iex` missing `-ExecutionPolicy Bypass`" claim is a **false positive** — `installation/index.ts` invokes `["powershell","-NoProfile","-ExecutionPolicy","Bypass","-Command", …]`; and the "curl.exe needs explicit TLS flags / `-k`" suggestion is backwards (`curl.exe` validates TLS by default).

### MAJOR — arch detection rejects valid 64-bit machines under 32-bit PowerShell (Bug) — `install.ps1` (arch check) — _GPT-5.4_
`$env:PROCESSOR_ARCHITECTURE -ne "AMD64"` is wrong under WOW64: a **32-bit PowerShell host on 64-bit Windows** reports `PROCESSOR_ARCHITECTURE = x86`, so a perfectly valid AMD64 machine hits `Unsupported OS/Arch: windows/x86` and exits 1. The real architecture lives in `PROCESSOR_ARCHITEW6432` in that case. Check both:
```powershell
$rawArch = if ($env:PROCESSOR_ARCHITEW6432) { $env:PROCESSOR_ARCHITEW6432 } else { $env:PROCESSOR_ARCHITECTURE }
if ($rawArch -ne "AMD64") { ... }
```
ARM64 is still correctly rejected; this only unblocks the 64-bit-Windows-with-32-bit-pwsh case.

### MAJOR (parity) — no integrity verification of the downloaded archive (Security) — `install.ps1` (`Install-Target`) — _Kimi, GLM-5_
The zip is downloaded and `Expand-Archive`d with no checksum/signature check, and self-update runs `irm … | iex`. This is real, but it **mirrors the existing `curl | bash` installer's posture**, so it's a pre-existing threat-model choice, not a regression introduced here. Worth raising the bar for **both** installers: if releases publish a checksums file, verify it before extracting —
```powershell
$actual = (Get-FileHash $zipPath -Algorithm SHA256).Hash
if ($actual -ne $expected) { throw "checksum mismatch for $filename" }
```
At minimum, call out in the PR that integrity verification is intentionally deferred to match the bash script.

### MINOR — `nothrow: true` may report a failed Windows upgrade as success — `installation/index.ts` (`upgradePowershell`) — _Kimi_
`Process.run([...], { nothrow: true })` won't throw on a non-zero PowerShell exit. Confirm the upgrade caller inspects the returned exit code; otherwise a failed self-update silently looks successful and the user never gets the "download a release binary manually" fallback that `upgrade()` would otherwise `notify()`.

### MINOR — malformed "Installed version: ." line — `install.ps1` (version probe) — _MiniMax_
When `altimate --version` yields empty output, the script prints `Installed version: .`. Guard the empty case (pre-existing in the bash installer too).

### MINOR — tests only assert the script *contains* substrings (Testing) — _MiniMax, Kimi_
`windows-install.test.ts` greps `install.ps1` text for expected flags; it exercises no behavior. The riskiest paths — arch/WOW64 detection, AVX2 illegal-instruction retry, `-NoPathUpdate`, invalid-version rejection — have no behavioral coverage. PowerShell is awkward to run on the CI host, but a Pester smoke test (mocking the Win32 calls) or at least a documented manual check on a real Windows box would de-risk the two MAJORs above. Add negative cases (unknown version → friendly error; `-NoPathUpdate`/`-ForceBaseline` honored).

### NIT — _GLM-5_
- No `-Help`/usage block — the bash installer has `usage()`; users can't discover `-Version`/`-NoPathUpdate`/`-ForceBaseline` without reading source.
- Two `Add-Type` P/Invoke blocks land in different sub-namespaces (`Win32.AltimateKernel32` vs `Win32.NativeMethods`) — consolidate.
- Inconsistent error prefixes (`Error:` vs bare) make log parsing harder.

- Strong feature parity with the bash installer; the Windows-specific hard parts are handled well (registry PATH vs `setx`'s 1024-char truncation, `WM_SETTINGCHANGE` broadcast, renaming the locked running exe before replacing it).
- The self-update wiring is thoughtful: HEAD-probe the install URL first to surface a friendly error instead of an opaque `irm | iex` failure, and pass the pinned `VERSION` through env.
- AVX2 detection via `IsProcessorFeaturePresent(40)` with a baseline fallback + illegal-instruction rescue is a nice touch and matches the bash behavior.

**Bottom line:** fix the WOW64 arch check (it blocks real users), decide on the checksum question for both installers, and confirm the `nothrow` upgrade-failure path — then this is a solid Windows on-ramp.

ralphstodomingo:
### ✅ Tested on native Windows (PowerShell 5.1, Win11 x64 AMD64)

Ran the prescribed `irm … | iex` one-liner against the branch's `install.ps1` (raw GitHub URL, since the `www.altimate.sh/install.ps1` 302 is the noted follow-up and isn't live yet). The `irm | iex` *mechanism* is what the redirect will serve once it ships.

**Setup:** cleared the existing npm-global `@altimateai/altimate-code@0.7.3` first → clean slate, no `.altimate\bin`.

| Check | Result |
|---|---|
| Install (`irm … \| iex`) | exit **0**; pulled `altimate-windows-x64.zip` (AVX2 build) from releases, extracted |
| Binary | `C:\Users\…\.altimate\bin\altimate.exe`, 268 MB |
| `altimate --version` | `0.8.7` (bare string → idempotency check matches cleanly) |
| `altimate --help` | full command registry loads |
| AVX2 illegal-instruction probe | passed (no baseline fallback on this CPU) |
| Replace hygiene | no leftover `.exe.old` |
| User PATH | single `…\.altimate\bin` entry written via registry |
| Re-run (idempotency) | `Version 0.8.7 already installed`, no re-download |

**Same-terminal PATH:** confirmed the install line `$env:Path = "$InstallDir;$env:Path"` makes bare `altimate` resolvable **in the same session that ran the one-liner** — no refresh needed. New terminals read the registry PATH; only *other* already-open shells need a fresh window (standard Windows installer behavior).

**Not exercised** (both by-design, gated on external bits): the AVX2→baseline rescue (this CPU has AVX2 — would need a `-ForceBaseline` or non-AVX2 box) and the native `altimate upgrade` dispatch (HEAD-checks the same not-yet-live `install.ps1` redirect).

mdesmet:
## Review remarks addressed (pushed in `bb393db`)

Thanks all — here's the disposition of every remark.

### Fixed
- **WOW64 arch detection (MAJOR, @anandgupta42 / GPT-5.4)** — `install.ps1` now resolves arch via `PROCESSOR_ARCHITEW6432`, so a 32-bit PowerShell host on 64-bit Windows is correctly detected as AMD64 instead of being rejected as unsupported `x86`. Covered by new behavioral Pester tests.
- **No `-Help`/usage (NIT, GLM-5)** — added `-Help` switch + `Show-Usage` mirroring the bash `usage()`.
- **Two `Add-Type` blocks (NIT, GLM-5)** — consolidated into one `Win32.AltimateNative` type (`IsProcessorFeaturePresent` + `SendMessageTimeout`).
- **Inconsistent error prefixes (NIT, GLM-5)** — all user-facing errors now go through `Write-Err` with a uniform `error: ` prefix.
- **Tests only assert substrings (MINOR, MiniMax/Kimi)** — added `test/windows/install.Tests.ps1` (Pester) running the script as a subprocess to cover syntax, `-Help`, the WOW64 fix, ARM64/x86 rejection, and unknown-version handling, plus a `windows-latest` CI lane. **6/6 pass on PowerShell 7.6.2.**

### Checksum / integrity (MAJOR-parity, @coderabbitai + Kimi/GLM-5)
Releases don't currently publish a checksums asset, so the suggested `$url.sha256` fetch would **404** — and the `catch { "continuing anyway" }` would make it a silent no-op that *looks* like verification. As @anandgupta42's panel noted, this mirrors the existing `curl | bash` installer's posture (no regression). I've documented the intentional deferral in `install.ps1`, and I'm opening a **follow-up PR** to publish a `checksums.txt` release asset and verify it in **both** installers for true parity.

### Verified — no change needed
- **`nothrow` upgrade failure (MINOR, Kimi)** — `upgrade()` already does `if (!result || result.code !== 0) throw new UpgradeFailedError(...)`; the PS installer propagates non-zero via `exit 1` + `$ErrorActionPreference="Stop"`. A failed Windows self-update **is** surfaced with the manual-fallback message.
- **"Installed version: ." (MINOR, MiniMax)** — already guarded by `elseif ($installedVersion)`; empty output prints nothing.
- **Dismissed false positives (consensus)** — `-ExecutionPolicy Bypass` is present in `installation/index.ts`; `curl.exe` validates TLS by default. Agreed, no action.
- **"15 TypeScript failures" (centralized bot)** — all in `test/altimate/simulation-suite.test.ts` + telemetry tests (warehouse/bridge error-mode simulation); this PR touches none of those files. Pre-existing/flaky, unrelated to the installer.

Thanks @ralphstodomingo for the on-box Windows verification. 🙏

mdesmet:
Follow-up for the checksum/integrity item is up: #942 (publishes `checksums.txt` with releases and verifies it in both the bash and PowerShell installers). It's stacked on this branch — merge this first, then #942.

ralphstodomingo:
### Re-tested on `bb393db8` (native Windows, PS 5.1, x64) — ✅ installs, with one robustness note

WOW64 arch fix, `-Help`, and the consolidated `Initialize-Native` P/Invoke all look good. Re-ran the full clean flow (cleared the npm global → `irm … | iex` → verify → idempotency). Install/version/PATH/idempotency all pass, and `-Help` renders correctly.

**One real snag:** the first attempt hard-failed:

```
error: Failed to fetch version information
```

Root cause was a transient **504 Gateway Timeout from `api.github.com`** on the `releases/latest` call (confirmed by reproducing the raw `Invoke-RestMethod` — TLS negotiated fine, GitHub itself returned 504). A retry installed cleanly. The issue is that this single unauthenticated API call is treated as **fatal with no retry**, so a momentary GitHub blip aborts the whole install — and unauthenticated `api.github.com` is also rate-limited at 60/hr/IP, so CI or repeated runs can trip it too.

Two cheap hardenings:

1. **Retry/backoff** around the version-resolve call.
2. **Don't hard-depend on the API in the `useLatest` path.** The download already uses `…/releases/latest/download/$filename`, which GitHub resolves "latest" for server-side — *no API needed*. The `Invoke-RestMethod` is only for the version-string display and the "already installed" short-circuit, so it could degrade gracefully (skip the skip-check) instead of `exit 1` when the API is unavailable.

**Minor, separate:** the "already installed" and `-Help` paths call `exit 0`, which under `irm … | iex` runs in-process. Fine for the documented `powershell -c "…"` child-process form, but pasted directly into an interactive shell it would close the window. (`curl | bash` exits in a subshell, so the bash installer doesn't have this.) Low severity.

Everything else looks solid for merge once the version-fetch robustness is decided on.

mdesmet:
Thanks @ralphstodomingo — confirmed and fixed.

**The version-fetch hard-fail is in both installers** (`install` bash:206-213 hits the same `api.github.com/.../releases/latest` and `exit 1`s identically). Fixed in a stacked PR: **#946**.

Both installers now, in the `latest` path: retry the API call 3× with backoff (bash via `curl --fail` so a 504 retries instead of parsing the error body), then **degrade gracefully** — install latest anyway (the download already uses `releases/latest/download/...`, no API needed) instead of aborting. Pinned `-Version`/`--version` still hard-fails on a genuine 404. Also guarded the "already installed" check so an unresolved version can't false-match.

**Re the `exit 0`/`irm | iex` interactive-window note:** agreed it's PowerShell-only (bash exits in a subshell) and only bites when pasted directly into an interactive prompt rather than the documented `powershell -c "…"` child-process form. Leaving as-is — low severity, no parity gap.

Stacking: #930 → #942 (checksums) / #946 (fetch resilience), both stacked here. I'll rebase the two onto `main` after #930 lands.

dev-punia-altimate:

- connection_refused [1.00ms]
- timeout
- permission_denied
- parse_error
- network_error
- auth_failure
- rate_limit
- internal_error
- empty_error
- connection_refused
- timeout
- permission_denied
- parse_error
- network_error [1.00ms]
- auth_failure

### Next Step

cc @mdesmet

PR #931, opened 2026-06-11
fix: bubble real dbt show error instead of generic "Could not parse"

### LINKED ISSUE #932: altimate-dbt execute masks real dbt show error with generic "Could not parse" message
## Problem

`altimate-dbt execute --query "..."` (which calls `execDbtShow` in `packages/dbt-tools/src/dbt-cli.ts`) swallows the real error from `dbt show` and surfaces a misleading generic message:

```json
{
  "error": "Could not parse dbt show output in any format (JSON, heuristic, or plain text). Got 0 JSON lines.",
  "fix": "Run: altimate-dbt doctor"
}
```

But running `dbt show` directly returns a clear actionable error, e.g.:

```
Runtime Error: Failed to read package: No dbt_project.yml found at expected path dbt_packages/dbt_utils/dbt_project.yml
```

The real error never reaches the caller.

## Reproducer

In a project where `dbt show` will fail for any reason (most reliably with a `dbt_packages/<pkg>/` missing `dbt_project.yml`):

```bash
altimate-dbt execute --query "select * from {{ ref('any_model') }}" --limit 1
```

You get the "Could not parse" error instead of the real Runtime Error.

## Root cause

`packages/dbt-tools/src/dbt-cli.ts`:

- Line ~228: `try { const { stdout } = await run(args); ... } catch { lines = [] }` — the `execFile` rejection (non-zero exit) is swallowed; its stderr is discarded.
- Line ~298: the Tier 3 plain-text fallback `try { ... } catch { /* fall through */ }` — same problem.
- Line ~302: throws the generic "Could not parse" message regardless of whether `dbt show` exited 0 or crashed.

## Cost

This bug burned hours during a single session: agents read "could not parse" as transient and retry alternate `altimate-dbt` commands. The actual error said "your project is structurally broken, stop trying" — but the agent never saw it.

## Proposed fix

When `run()` rejects:
1. Capture `error.stderr` from the `execFile` rejection.
2. Try to recover parseable JSON log lines from `error.stdout` (`--log-format json` writes errors to stdout too).
3. Scan JSON lines for `level: "error"` events and extract `msg`.
4. If all parse tiers fail AND we captured a `run()` error, surface the actual dbt error message instead of the generic "Could not parse".

Fall back to "Could not parse" only when `dbt show` exited 0 but the output really is unparseable.

## Fix

PR #931 implements this. 6 new bun:test cases cover real-stderr surfacing, structured-event preference, ENOENT fallback, and generic-message preservation on exit-0 unparseable output.

## Scope

Limited to `execDbtShow`. `execDbtCompile` and `execDbtCompileInline` share the same masking pattern but have additional fallback paths (manifest.json for compile, `--quiet` plain-text for inline) that reduce caller impact — addressed separately if telemetry shows masking there too.

sahrizvi:
Closes #932

## Summary

`altimate-dbt execute --query "..."` (which calls `execDbtShow` in `packages/dbt-tools/src/dbt-cli.ts`) swallows the real error from `dbt show` and surfaces a misleading generic message:

```json
{
  "error": "Could not parse dbt show output in any format (JSON, heuristic, or plain text). Got 0 JSON lines.",
  "fix": "Run: altimate-dbt doctor"
}
```

But running `dbt show` directly produces a clear actionable error, e.g.:

```
Runtime Error: Failed to read package: No dbt_project.yml found at expected path dbt_packages/dbt_utils/dbt_project.yml
```

The real error never reaches the caller. Agents read "could not parse" as transient and retry alternate commands instead of bailing out — burns budget on a project that is structurally broken.

## Root cause

`packages/dbt-tools/src/dbt-cli.ts` has two adjacent catch blocks in `execDbtShow` that swallow the `run()` rejection silently:

```ts
// ~line 224 — Tier 1 JSON attempt
try { const { stdout } = await run(args); lines = parseJsonLines(stdout) }
catch { lines = [] }            // ← discards err.stderr

// ~line 298 — Tier 3 plain-text fallback
try { ... await run(plainArgs) ... }
catch { /* fall through */ }    // ← also discards err.stderr

throw new Error("Could not parse dbt show output in any format (JSON, heuristic, or plain text). ...")
```

When `execFile` rejects with a non-zero exit, the error object carries `.stderr` and `.stdout`. Both catches discard them. The generic "Could not parse" message was designed for the case *"dbt exited 0 but the output is unparseable"* — but it currently fires for the *"dbt actually crashed"* case as well, conflating two very different failure modes.

## Fix

- Capture the `execFile` rejection on both tiers (`primaryRunError`, `plainRunError`).
- Recover any partial JSON log lines from the failed run's `stdout` — dbt with `--log-format json` emits structured `level: "error"` events before exit.
- Before the generic throw, call `extractDbtError(...)` which picks (in order): structured error event > primary stderr > plain stderr > exception message.
- If `extractDbtError` returns anything, surface it as `dbt show failed: <real msg>`. Fall through to the generic message only when both runs exited 0.

Net: ~45 lines in `dbt-cli.ts` (helper + 2 small guard sites), 6 new test cases.

## Scope

Limited to `execDbtShow`. `execDbtCompile` and `execDbtCompileInline` share the same `catch { lines = [] }` pattern but have additional fallback paths (manifest.json for compile, no-args plain-text retry for inline) that reduce caller impact. Worth addressing in a follow-up PR if telemetry shows masking there too — kept out of this PR to keep the diff focused.

## Test plan

- [x] `bun test test/dbt-cli.test.ts` → 24/24 pass (18 pre-existing + 6 new)
- [x] Generic "Could not parse" message still surfaces when `dbt show` exits 0 but emits unparseable output (regression guard preserved)
- [x] Real stderr from a crashing `dbt show` surfaces in the thrown error
- [x] Structured `level: "error"` event preferred over raw stderr when both present
- [x] `spawn ENOENT` (no dbt binary) surfaces clearly
- [ ] Reviewer: smoke against a real project with a corrupted `dbt_packages/<pkg>/dbt_project.yml`; expect to see the real "Failed to read package" message, not "Could not parse"

## Links

- Issue: #932
- Catalog entry: `03-issues-and-fixes.md` Issue #14
- Discovered while investigating the auto-deps clobber on a sister repo ([altimate-dbt-integration#120](https://github.com/AltimateAI/altimate-dbt-integration/pull/120)); the masked error happened to be a corrupted `dbt_packages/` fixture

## Summary by CodeRabbit

* **Bug Fixes**
  * dbt error messages now display actual error details instead of generic fallback messages
  * Enhanced error reporting to prioritize structured error information from dbt logs when available
  * Improved error capture and synthesis across execution modes, providing more accurate failure diagnostics

### DIFF
--- a/packages/dbt-tools/src/dbt-cli.ts
+++ b/packages/dbt-tools/src/dbt-cli.ts
@@ -222,9 +222,9 @@ export async function execDbtShow(sql: string, limit?: number) {
   if (limit !== undefined) args.push("--limit", String(limit))
 
   let lines: Record<string, unknown>[]
-  // AI-7082: capture the run() error so we can bubble the real dbt failure
-  // up if all parse tiers fail; the generic "Could not parse" alone misleads
-  // callers into treating structural project errors as transient.
+  // Capture the run() error so we can bubble the real dbt failure up if all
+  // parse tiers fail; the generic "Could not parse" alone misleads callers
+  // into treating structural project errors as transient.
   let primaryRunError: ExecFileError | undefined
   try {
     const { stdout } = await run(args)
@@ -305,8 +305,8 @@ export async function execDbtShow(sql: string, limit?: number) {
     plainRunError = e as ExecFileError
   }
 
-  // AI-7082: if either run() rejected, dbt actually crashed — surface the real
-  // error instead of the generic "Could not parse" message.
+  // If either run() rejected, dbt actually crashed — surface the real error
+  // instead of the generic "Could not parse" message.
   const realError = extractDbtError(lines, primaryRunError, plainRunError)
   if (realError) {
     throw new Error(`dbt show failed: ${realError}`)
@@ -318,7 +318,7 @@ export async function execDbtShow(sql: string, limit?: number) {
   )
 }
 
-/** AI-7082: shape of an execFile rejection — carries stdout/stderr alongside message. */
+/** Shape of an execFile rejection — carries stdout/stderr alongside message. */
 interface ExecFileError extends Error {
   stdout?: string
   stderr?: string
@@ -326,7 +326,7 @@ interface ExecFileError extends Error {
 }
 
 /**
- * AI-7082: pick the best human-readable error from a failed `dbt show` invocation.
+ * Pick the best human-readable error from a failed `dbt show` invocation.
  *
  * Preference order:
  *   1. A structured `level: "error"` event in the JSON log (dbt's own error msg).
--- a/packages/dbt-tools/src/dbt-cli.ts
+++ b/packages/dbt-tools/src/dbt-cli.ts
@@ -222,11 +222,16 @@ export async function execDbtShow(sql: string, limit?: number) {
   if (limit !== undefined) args.push("--limit", String(limit))
 
   let lines: Record<string, unknown>[]
+  // AI-7082: capture the run() error so we can bubble the real dbt failure
+  // up if all parse tiers fail; the generic "Could not parse" alone misleads
+  // callers into treating structural project errors as transient.
+  let primaryRunError: ExecFileError | undefined
   try {
     const { stdout } = await run(args)
     lines = parseJsonLines(stdout)
-  } catch {
-    lines = []
+  } catch (e) {
+    primaryRunError = e as ExecFileError
+    lines = parseJsonLines(primaryRunError.stdout ?? "")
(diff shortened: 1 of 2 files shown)


sahrizvi:
Superseded by #933 — same commits under a renamed branch that drops the stale `AI-7082` Jira key. `altimate-code` issues live on GitHub (#932), not Jira.

PR #933, opened 2026-06-11
fix: bubble real dbt show error instead of generic "Could not parse"

### LINKED ISSUE #932: altimate-dbt execute masks real dbt show error with generic "Could not parse" message

### LINKED ISSUE #943: execDbtCompile / execDbtCompileInline mask real dbt errors with generic wrapper message
## Symptom

When `dbt compile` or `dbt compile --inline` fails with a real dbt error (e.g. `Compilation Error: Model 'foo' depends on a node named 'bar' which was not found`), `execDbtCompile` and `execDbtCompileInline` in `packages/dbt-tools/src/dbt-cli.ts` surface a generic wrapper message that masks the actual cause. Users see `"Could not compile … Last error: Command failed: …"` instead of the structured dbt error they need to fix the problem.

## Root cause

Both functions wrap `run()` in a `try { … } catch { lines = [] }` that swallows the run-time error entirely:

```ts
// execDbtCompile
try {
  const { stdout } = await run(args)
  lines = parseJsonLines(stdout)
} catch {
  lines = []
}
```

```ts
// execDbtCompileInline — same pattern
```

The structured `level: "error"` event present in dbt's JSON log on a crashed run is never inspected, and the final throw embeds Node's generic `Command failed: …` message rather than dbt's category-prefixed error (`Compilation Error: …`, `Database Error: …`, etc.).

## Expected

Both functions should mirror the pattern already in place for `execDbtShow`: capture the run error, parse `level: "error"` events from any partial stdout, and surface the real dbt error via the existing `extractDbtError` helper (with the same SQL-redaction guarantees provided by `fallbackExitMessage`).

## Suggested fix

Extract a shared `runWithErrorBubbling(args, label)` helper from the `execDbtShow` pattern and apply to all three `execDbt*` functions, so error UX is consistent across the surface.

## Related

- #932 / #933 — the same fix was applied to `execDbtShow`. Adjacent code path, distinct review surface, hence a separate issue.

sahrizvi:
PINEAPPLE

Closes #932

(Supersedes #931 — same commits, branch renamed to drop the stale Jira key.)

## Summary

`altimate-dbt execute --query "..."` (which calls `execDbtShow` in `packages/dbt-tools/src/dbt-cli.ts`) swallows the real error from `dbt show` and surfaces a misleading generic message:

```json
{
  "error": "Could not parse dbt show output in any format (JSON, heuristic, or plain text). Got 0 JSON lines.",
  "fix": "Run: altimate-dbt doctor"
}
```

But running `dbt show` directly produces a clear actionable error, e.g.:

```
Runtime Error: Failed to read package: No dbt_project.yml found at expected path dbt_packages/dbt_utils/dbt_project.yml
```

Agents read "could not parse" as transient and retry alternate commands instead of bailing out — burns budget on a project that is structurally broken.

## Root cause

`packages/dbt-tools/src/dbt-cli.ts` has two adjacent catch blocks in `execDbtShow` that swallow the `run()` rejection silently:

```ts
// ~line 224 — Tier 1 JSON attempt
try { const { stdout } = await run(args); lines = parseJsonLines(stdout) }
catch { lines = [] }            // ← discards err.stderr

// ~line 298 — Tier 3 plain-text fallback
try { ... await run(plainArgs) ... }
catch { /* fall through */ }    // ← also discards err.stderr

throw new Error("Could not parse dbt show output in any format (JSON, heuristic, or plain text). ...")
```

When `execFile` rejects with a non-zero exit, the error object carries `.stderr` and `.stdout`. Both catches discard them. The generic "Could not parse" message was designed for the case *"dbt exited 0 but the output is unparseable"* — but it currently fires for the *"dbt actually crashed"* case as well, conflating two very different failure modes.

## Fix

- Capture the `execFile` rejection on both tiers (`primaryRunError`, `plainRunError`).
- Recover any partial JSON log lines from the failed run's `stdout` — dbt with `--log-format json` emits structured `level: "error"` events before exit.
- Before the generic throw, call `extractDbtError(...)` which picks (in order): structured error event > primary stderr > plain stderr > exception message.
- If `extractDbtError` returns anything, surface it as `dbt show failed: <real msg>`. Fall through to the generic message only when both runs exited 0.

Net: ~45 lines in `dbt-cli.ts` (helper + 2 small guard sites), 6 new test cases.

## Scope

Limited to `execDbtShow`. `execDbtCompile` and `execDbtCompileInline` share the same `catch { lines = [] }` pattern but have additional fallback paths (manifest.json for compile, no-args plain-text retry for inline) that reduce caller impact. Worth addressing in a follow-up PR if telemetry shows masking there too — kept out of this PR to keep the diff focused.

## Test plan

- [x] `bun test test/dbt-cli.test.ts` → 24/24 pass (18 pre-existing + 6 new)
- [x] Generic "Could not parse" message still surfaces when `dbt show` exits 0 but emits unparseable output (regression guard preserved)
- [x] Real stderr from a crashing `dbt show` surfaces in the thrown error
- [x] Structured `level: "error"` event preferred over raw stderr when both present
- [x] `spawn ENOENT` (no dbt binary) surfaces clearly
- [ ] Reviewer: smoke against a real project with a corrupted `dbt_packages/<pkg>/dbt_project.yml`; expect to see the real "Failed to read package" message, not "Could not parse"

## Links

- Issue: #932
- Catalog entry: `03-issues-and-fixes.md` Issue #14
- Discovered while investigating the auto-deps clobber on a sister repo ([altimate-dbt-integration#120](https://github.com/AltimateAI/altimate-dbt-integration/pull/120)); the masked error happened to be a corrupted `dbt_packages/` fixture

---
## Summary by cubic
Bubble real dbt errors across `execDbtShow`, `execDbtCompile`, and `execDbtCompileInline`, while redacting inline SQL and separating parser failures from dbt execution issues. Fixes misleading “Could not parse” cases and makes failures actionable. Addresses #932; closes #943, #944, #945.

- **Bug Fixes**
  - `run()` now attaches `stdout`/`stderr` on `execFile` rejection; shared helpers `extractDbtError`/`fallbackExitMessage`/`bubbleDbtError` prefer last structured JSON error > primary stderr > plain stderr > a redacted exit-status, strip ANSI, and avoid doubling dbt category prefixes.
  - `execDbtShow`: distinguish failure modes; if JSON-mode crashes, bubble the real dbt error; if JSON-mode succeeds but is unparseable and plain-text then fails, throw a parser error. Skip Tier 1/2 when the primary run failed, never feed crashed stdout into heuristics; still try Tier 3 and return the table if it succeeds.
  - `execDbtCompile` and `execDbtCompileInline`: adopt the same error bubbling and SQL redaction (covers exit codes and signal/timeouts); also bubble errors from manifest/plain fallbacks.
  - Guard data shape: treat truthy non-array `data.preview` as empty rows to avoid malformed results.
  - Tests expanded for JSON/stderr preference, parser-vs-dbt attribution, ANSI strip, SQL redaction, Buffer I/O, Tier 3 recovery, and non-array preview; 41 passing tests.

<sup>Written for commit 32ffe32afedd6983c4755fb6251e3b09c64e662f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/933?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for `dbt show` to surface actual dbt failure details, preferring structured JSON error events when available and avoiding generic “Could not parse …” messages on real crashes.
  * Made `dbt compile` and `dbt compile --inline` report consistent, safer failure messages (ANSI cleaned and inline SQL not leaked).

* **Tests**
  * Expanded coverage for `dbt show`/compile failure modes, including fallback behavior when process output is missing, correct error-event selection, retry recovery, and message sanitization.

### DIFF
--- a/packages/dbt-tools/src/dbt-cli.ts
+++ b/packages/dbt-tools/src/dbt-cli.ts
@@ -222,11 +222,16 @@ export async function execDbtShow(sql: string, limit?: number) {
   if (limit !== undefined) args.push("--limit", String(limit))
 
   let lines: Record<string, unknown>[]
+  // Capture the run() error so we can bubble the real dbt failure up if all
+  // parse tiers fail; the generic "Could not parse" alone misleads callers
+  // into treating structural project errors as transient.
+  let primaryRunError: ExecFileError | undefined
   try {
     const { stdout } = await run(args)
     lines = parseJsonLines(stdout)
-  } catch {
-    lines = []
+  } catch (e) {
+    primaryRunError = e as ExecFileError
+    lines = parseJsonLines(primaryRunError.stdout ?? "")
   }
 
   // --- Tier 1: known field paths ---
@@ -281,6 +286,7 @@ export async function execDbtShow(sql: string, limit?: number) {
   }
 
   // --- Tier 3: plain text fallback (ASCII table) ---
+  let plainRunError: ExecFileError | undefined
   try {
     const plainArgs = ["show", "--inline", sql]
     if (limit !== undefined) plainArgs.push("--limit", String(limit))
@@ -295,13 +301,64 @@ export async function execDbtShow(sql: string, limit?: number) {
         compiledSql: sql,
       }
     }
-  } catch {
-    // Plain text dbt show also failed — fall through to error below
+  } catch (e) {
+    plainRunError = e as ExecFileError
+  }
+
+  // If either run() rejected, dbt actually crashed — surface the real error
+  // instead of the generic "Could not parse" message.
+  const realError = extractDbtError(lines, primaryRunError, plainRunError)
+  if (realError) {
+    throw new Error(`dbt show failed: ${realError}`)
   }
 
   throw new Error(
     "Could not parse dbt show output in any format (JSON, heuristic, or plain text). " +
       `Got ${lines.length} JSON lines.`,
+  )
+}
+
+/** Shape of an execFile rejection — carries stdout/stderr alongside message. */
+interface ExecFileError extends Error {
+  stdout?: string
+  stderr?: string
+  code?: number | string
+}
+
+/**
+ * Pick the best human-readable error from a failed `dbt show` invocation.
+ *
+ * Preference order:
+ *   1. A structured `level: "error"` event in the JSON log (dbt's own error msg).
+ *   2. Stderr from the JSON-mode run.
+ *   3. Stderr from the plain-text-mode run.
+ *   4. The exception message itself.
+ *
+ * Returns undefined if neither run rejected — caller falls back to the generic
+ * "Could not parse" message, which is correct when dbt exited 0 but emitted
+ * something we can't decode.
+ */
+function extractDbtError(
+  lines: Record<string, unknown>[],
+  primary?: ExecFileError,
+  plain?: ExecFileError,
+): string | undefined {
+  if (!primary && !plain) return undefined
+
+  const errorEvent = lines.find(
+    (l: any) => l.info?.level === "error" || l.level === "error",
+  ) as any
+  const structuredMsg = errorEvent?.info?.msg ?? errorEvent?.msg
+
+  const primaryStderr = primary?.stderr?.toString().trim()
+  const plainStderr = plain?.stderr?.toString().trim()
+
+  return (
+    (typeof structuredMsg === "string" && structuredMsg.length > 0 ? structuredMsg : undefined) ??
+    (primaryStderr && primaryStderr.length > 0 ? primaryStderr : undefined) ??
+    (plainStderr && plainStderr.length > 0 ? plainStderr : undefined) ??
+    primary?.message ??
+    plain?.message
   )
 }
 
--- a/packages/dbt-tools/src/dbt-cli.ts
+++ b/packages/dbt-tools/src/dbt-cli.ts
@@ -285,17 +285,20 @@ export async function execDbtShow(sql: string, limit?: number) {
         (previewLine as any).result?.preview ??
         (previewLine as any).result?.rows
 
-      // Guard JSON.parse — fall through to Tier 2 on malformed strings
+      // The previewLine match upstream only checked for truthiness, so a
+      // future dbt version emitting `data.preview = {}` or `= 42` would
+      // flow into `rows` unchecked and the downstream `data: rows` field
+      // would crash callers that do `.map` / `.length`. Guard explicitly
+      // for the three shapes we accept (parsed JSON array, native array,
+      // malformed) and emit an empty result for anything else.
       let rows: Record<string, unknown>[]
       if (typeof preview === "string") {
         const parsed = safeJsonParse(preview)
-        if (Array.isArray(parsed)) {
-          rows = parsed
-        } else {
-          rows = [] // Malformed — will fall through below
-        }
-      } else {
+        rows = Array.isArray(parsed) ? parsed : []
+      } else if (Array.isArray(preview)) {
         rows = preview
+      } else {
+        rows = []
       }
 
       // Return the result — even if empty. An empty preview means the query returned
@@ -465,51 +468,87 @@ function fallbackExitMessage(primary?: ExecFileError, plain?: ExecFileError): st
 }
 
 /**
+ * Build a user-facing error message from a failed `dbt <cmd>` invocation.
+ *
+ * Used by every execDbt* function so all three share the same error UX:
+ *   - structured `level: "error"` event from JSON logs > primary stderr >
+ *     plain-text stderr > redacted exit-status fallback
+ *   - ANSI escapes stripped
+ *   - inline SQL / command-line redacted from any err.message-derived path
+ *   - no doubled prefix when dbt's own category prefix is already present
+ */
+function bubbleDbtError(label: string, primary?: ExecFileError, plain?: ExecFileError): string {
+  const errorLogLines = primary?.stdout ? parseJsonLines(primary.stdout.toString()) : []
+  const real = extractDbtError(errorLogLines, primary, plain)
+  if (real) {
+    const hasDbtCategoryPrefix = /^(Compilation|Database|Runtime|Parsing|Validation|Dependency)\s+Error\b/.test(
+      real,
+    )
+    return hasDbtCategoryPrefix ? real : `${label}: ${real}`
+  }
+  return `${label}: ${fallbackExitMessage(primary, plain) ?? "unknown error"}`
+}
+
+/**
  * Compile a model via `dbt compile --select <model>` and return compiled SQL.
  */
 export async function execDbtCompile(model: string): Promise<{ sql: string }> {
   const args = ["compile", "--select", model, "--output", "json", "--log-format", "json"]
 
-  let lines: Record<string, unknown>[]
+  let lines: Record<string, unknown>[] = []
+  let primaryRunError: ExecFileError | undefined
   try {
     const { stdout } = await run(args)
     lines = parseJsonLines(stdout)
-  } catch {
-    lines = []
+  } catch (e) {
+    primaryRunError = toExecFileError(e)
   }
 
-  // --- Tier 1: known field paths ---
-  const sql = findCompiledSql(lines)
-  if (sql) return { sql }
+  // Skip success-only tiers when the primary run failed (same anti-spurious-
+  // data reasoning as execDbtShow).
+  if (!primaryRunError) {
+    // --- Tier 1: known field paths ---
+    const sql = findCompiledSql(lines)
+    if (sql) return { sql }
 
-  // --- Tier 2: heuristic deep scan ---
-  for (const line of lines) {
-    const found = deepFind(line, (val) => looksLikeSql(val))
-    if (found) return { sql: found as string }
+    // --- Tier 2: heuristic deep scan ---
+    for (const line of lines) {
+      const found = deepFind(line, (val) => looksLikeSql(val))
+      if (found) return { sql: found as string }
+    }
   }
 
-  // --- Tier 3: read compiled SQL from manifest.json (more reliable than stdout) ---
-  // dbt compile writes compiled_code to target/manifest.json even when stdout is logs.
-  // We run compile without JSON flags so it writes to manifest, then read the artifact.
+  // --- Manifest fallback ---
+  // dbt compile writes compiled_code to target/manifest.json even when stdout
+  // is just logs. Re-run plain (no JSON flags) so the artifact is fresh, then
+  // read it back. We tolerate a failure here (a prior successful compile may
+  // have left a usable manifest) but capture the error for later bubbling.
+  let manifestRunError: ExecFileError | undefined
   try {
     await run(["compile", "--select", model])
-  } catch {
-    // Compile may fail (e.g., dbt not found, project errors) — continue to manifest check
-    // since a prior successful compile may have left a usable manifest
+  } catch (e) {
+    manifestRunError = toExecFileError(e)
   }
   const fromManifest = readCompiledFromManifest(model)
   if (fromManifest) return { sql: fromManifest }
 
-  // Last resort: return stdout (may contain logs mixed with SQL)
+  // --- Last resort: plain compile, return raw stdout ---
+  let plainRunError: ExecFileError | undefined
   try {
     const { stdout: plainOut } = await run(["compile", "--select", model])
     return { sql: plainOut.trim() }
   } catch (e) {
-    throw new Error(
-      `Could not compile model '${model}' in any format (JSON, heuristic, or manifest). ` +
-        `Last error: ${e instanceof Error ? e.message : String(e)}`,
-    )
+    plainRunError = toExecFileError(e)
   }
+
+  // If dbt actually failed at any tier, surface the real dbt error via the
+  // shared helper so the message is SQL-safe, ANSI-stripped, and consistent
+  // with execDbtShow's error UX.
+  if (primaryRunError || plainRunError || manifestRunError) {
+    throw new Error(bubbleDbtError("dbt compile failed", primaryRunError, plainRunError ?? manifestRunError))
+  }
+
+  throw new Error(`Could not compile model '${model}' in any format (JSON, heuristic, or manifest).`)
 }
 
 /**
@@ -521,34 +560,45 @@ export async function execDbtCompileInline(
 ): Promise<{ sql: string }> {
   const args = ["compile", "--inline", sql, "--output", "json", "--log-format", "json"]
 
-  let lines: Record<string, unknown>[]
+  let lines: Record<string, unknown>[] = []
+  let primaryRunError: ExecFileError | undefined
   try {
     const { stdout } = await run(args)
     lines = parseJsonLines(stdout)
-  } catch {
-    lines = []
+  } catch (e) {
+    primaryRunError = toExecFileError(e)
   }
 
-  // --- Tier 1: known field paths ---
-  const compiled = findCompiledSql(lines)
-  if (compiled) return { sql: compiled }
+  // Skip success-only tiers when the primary run failed.
+  if (!primaryRunError) {
+    // --- Tier 1: known field paths ---
+    const compiled = findCompiledSql(lines)
+    if (compiled) return { sql: compiled }
 
-  // --- Tier 2: heuristic deep scan ---
-  for (const line of lines) {
-    const found = deepFind(line, (val) => looksLikeSql(val))
-    if (found) return { sql: found as string }
+    // --- Tier 2: heuristic deep scan ---
+    for (const line of lines) {
+      const found = deepFind(line, (val) => looksLikeSql(val))
+      if (found) return { sql: found as string }
+    }
   }
 
   // --- Tier 3: plain text fallback ---
+  let plainRunError: ExecFileError | undefined
   try {
     const { stdout: plainOut } = await run(["compile", "--inline", sql])
     return { sql: plainOut.trim() }
   } catch (e) {
-    throw new Error(
-      `Could not compile inline SQL in any format (JSON, heuristic, or plain text). ` +
-        `Last error: ${e instanceof Error ? e.message : String(e)}`,
-    )
+    plainRunError = toExecFileError(e)
   }
+
+  // bubbleDbtError redacts inline SQL from any err.message fallback — critical
+  // here because we're spawning `dbt compile --inline <user SQL>` and Node's
+  // rejection message embeds the full command line (with the SQL) verbatim.
+  if (primaryRunError || plainRunError) {
+    throw new Error(bubbleDbtError("dbt compile inline failed", primaryRunError, plainRunError))
+  }
+
+  throw new Error("Could not compile inline SQL in any format (JSON, heuristic, or plain text).")
 }
 
 /** Shared: extract compiled SQL from known dbt JSON output formats. */
--- a/packages/dbt-tools/src/dbt-cli.ts
+++ b/packages/dbt-tools/src/dbt-cli.ts
@@ -51,6 +51,20 @@ function getDbt(): ResolvedDbt {
   return resolvedDbt
 }
 
+/** Shape of an execFile rejection — carries stdout/stderr alongside message. */
+interface ExecFileError extends Error {
+  stdout?: string | Buffer
+  stderr?: string | Buffer
+  code?: number | string
+  signal?: string
+}
+
+/** Coerce an unknown rejection into something the catch blocks can read safely. */
+function toExecFileError(e: unknown): ExecFileError {
+  if (e instanceof Error) return e as ExecFileError
+  return new Error(String(e)) as ExecFileError
+}
+
 function run(args: string[]): Promise<{ stdout: string; stderr: string }> {
   const dbt = getDbt()
   const env = buildDbtEnv(dbt)
@@ -58,8 +72,18 @@ function run(args: string[]): Promise<{ stdout: string; stderr: string }> {
 
   return new Promise((resolve, reject) => {
     execFile(dbt.path, args, { timeout: 120_000, maxBuffer: 10 * 1024 * 1024, env, cwd }, (err, stdout, stderr) => {
-      if (err) reject(err)
-      else resolve({ stdout, stderr })
+      if (err) {
+        // Node's execFile passes stdout/stderr as separate callback arguments,
+        // not as properties on the error. Attach them here so callers can
+        // surface the real dbt failure text instead of Node's generic
+        // "Command failed: ..." message.
+        const execErr = err as ExecFileError
+        execErr.stdout = stdout
+        execErr.stderr = stderr
+        reject(execErr)
+      } else {
+        resolve({ stdout, stderr })
+      }
     })
   })
 }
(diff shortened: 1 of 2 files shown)


sahrizvi:
## Addresses cubic review feedback

Pushed `2ea03d79b`.

### cubic findings — both resolved

- **P2** (`execFile` rejection stdout/stderr not surfaced): already resolved upstream in `064b0dd6b` — `execDbtShow` now captures `e.stdout` / `e.stderr` from the rejection error and feeds them into `extractDbtError`. Verified in current `packages/dbt-tools/src/dbt-cli.ts`.
- **P3** (duplicate test): the new test at `dbt-cli.test.ts:199` ("preserves generic 'Could not parse' when dbt exited 0 but output unparseable") used the same mock and assertion as the existing **Tier 3: throws with helpful message when all tiers fail** test at line 145. Deleted the dup; Tier 3 still covers the generic-error path.

### Tests + verification

- `bun test test/dbt-cli.test.ts` → **23 pass / 0 fail**
- Independent codex review: "The commit only removes a duplicate test while preserving equivalent coverage in the existing Tier 3 failure test. No functional code or unique test coverage appears to be lost."

### Centralized test bot failures — likely shared infra, not PR-caused

The `dev-punia-altimate` bot's `15 TypeScript failures` (connection_refused / timeout / parse_error / network_error / auth_failure / rate_limit / internal_error / empty_error / oom / permission_denied) appear with the **identical name set** on completely unrelated PRs #935 and #937. Strong signal these are shared fault-injection-harness issues, not introduced by this PR. Happy to dig further if the bot owner wants — flagging here in case it's expected noise.

packages/dbt-tools/src/dbt-cli.ts
```diff
@@ -222,11 +222,16 @@ export async function execDbtShow(sql: string, limit?: number) {
...
+  // into treating structural project errors as transient.
+  let primaryRunError: ExecFileError | undefined
   try {
     const { stdout } = await run(args)
     lines = parseJsonLines(stdout)
-  } catch {
-    lines = []
+  } catch (e) {
+    primaryRunError = e as ExecFileError
+    lines = parseJsonLines(primaryRunError.stdout ?? "")
```
suryaiyer95: **CRITICAL — the fix is inert in production: `primaryRunError.stdout` is always `undefined`.**

`run()` (line 61) rejects with the bare `execFile` error: `if (err) reject(err)`. Node's `execFile` passes `stdout`/`stderr` as **separate callback arguments**, not as properties on the error object — and this manual Promise wrapper discards them on rejection (unlike `util.promisify(execFile)`, which decorates the error). So at runtime `primaryRunError.stdout` and `.stderr` are `undefined`:

- `parseJsonLines(undefined ?? "")` → `[]` → no structured error event
- `primary?.stderr` → `undefined` → no stderr surfaced
- `extractDbtError` falls through to `primary?.message` = Node's generic `"Command failed: <dbt> show --inline ..."`

The real dbt error this PR exists to surface (e.g. `"Runtime Error: Failed to read package..."`) is never shown. Verified empirically by reproducing the exact `run()` pattern under Node v20.20 and Bun 1.3.14: `'stdout' in err === false`, `err.stdout === undefined` in both.

**Fix — attach stdout/stderr in `run()` before rejecting:**
```ts
execFile(dbt.path, args, { ... }, (err, stdout, stderr) => {
  if (err) {
    const execErr = err as ExecFileError
    execErr.stdout = stdout
    execErr.stderr = stderr
    reject(execErr)
  } else resolve({ stdout, stderr })
})
```

suryaiyer95 (commented):
**Multi-model consensus review** (Claude + Gemini 3.1 Pro + Kimi K2.5 + Grok 4, 1 convergence round).

**Verdict: REQUEST CHANGES** — Critical: 1, Major: 1, Minor: 5, Nits: 2.

The intent is excellent and the surrounding structure is clean, but as written the fix is **inert in production** (CRITICAL 1) and the same change opens a silent-wrong-data path once that's fixed (MAJOR 2). These two are coupled and must land together, with tests that exercise the real `execFile` callback wiring rather than a hand-decorated mock.

Inline comments below. (Posted as a COMMENT review, not a formal block — flip to Request Changes if you prefer.)

_Note: `parseJsonLines` cannot throw (it wraps every `JSON.parse` in its own try/catch), so the "fallback breaks if parse throws" concern was considered and rejected._

packages/dbt-tools/src/dbt-cli.ts
```diff
@@ -222,11 +222,16 @@ export async function execDbtShow(sql: string, limit?: number) {
...
+  // parse tiers fail; the generic "Could not parse" alone misleads callers
+  // into treating structural project errors as transient.
+  let primaryRunError: ExecFileError | undefined
   try {
     const { stdout } = await run(args)
     lines = parseJsonLines(stdout)
-  } catch {
-    lines = []
+  } catch (e) {
+    primaryRunError = e as ExecFileError
```
suryaiyer95: **MINOR — unguarded `e as ExecFileError` cast** (also at line 305). Low-risk since `execFile` realistically rejects with an `Error`, but a cheap guard documents intent and hardens against a non-`Error` rejection:
```ts
const isExecFileError = (e: unknown): e is ExecFileError => e instanceof Error
primaryRunError = isExecFileError(e) ? e : (new Error(String(e)) as ExecFileError)
```

packages/dbt-tools/src/dbt-cli.ts
```diff
@@ -295,8 +301,15 @@ export async function execDbtShow(sql: string, limit?: number) {
...
     }
-  } catch {
-    // Plain text dbt show also failed — fall through to error below
+  } catch (e) {
+    plainRunError = e as ExecFileError
+  }
+
+  // If either run() rejected, dbt actually crashed — surface the real error
+  // instead of the generic "Could not parse" message.
+  const realError = extractDbtError(lines, primaryRunError, plainRunError)
```
suryaiyer95: **MAJOR — feeding the failed run's stdout into Tier 1/Tier 2 can return bogus rows / mask the error.**

Once CRITICAL 1 is fixed and `primaryRunError.stdout` is actually populated on a crash, line 234 pushes crash log lines into `lines`, which then flow through Tier 1 (238-242) and the Tier 2 heuristic scan (278-286) **before** `extractDbtError` is consulted here at line 310. `looksLikeRowData` matches **any** depth-≤5 array whose first element is an object, so an incidental array in a dbt crash log can be returned as spurious "rows" (silent wrong data) — worse than the original misleading error. The existing comment at lines 269-271 already warns about exactly this Tier-2 false-positive hazard.

**Fix — when a run errored, extract the error *before* the heuristic tiers and source success-path `lines` only from a successful run:**
```ts
if (primaryRunError || plainRunError) {
  const realError = extractDbtError(parseJsonLines(primaryRunError?.stdout ?? ""), primaryRunError, plainRunError)
  if (realError) throw new Error(`dbt show failed: ${realError}`)
}
```
Also add a regression test: a failed run whose stdout contains an incidental array-of-objects must **throw**, not return rows.

packages/dbt-tools/src/dbt-cli.ts
```diff
@@ -295,8 +301,15 @@ export async function execDbtShow(sql: string, limit?: number) {
...
+  if (realError) {
```
suryaiyer95: **MINOR — `extractDbtError` is invoked unconditionally** even when neither run failed. It self-short-circuits via `if (!primary && !plain) return undefined` (line 346), so this is cosmetic, but gating the call behind `if (primaryRunError || plainRunError)` (the MAJOR-2 fix above) makes the intent explicit: we only look for a "real dbt error" when dbt actually errored.

packages/dbt-tools/src/dbt-cli.ts
```diff
@@ -305,6 +318,50 @@ export async function execDbtShow(sql: string, limit?: number) {
   )
 }
 
+/** Shape of an execFile rejection — carries stdout/stderr alongside message. */
+interface ExecFileError extends Error {
+  stdout?: string
+  stderr?: string
```
suryaiyer95: **NIT — `stdout`/`stderr` typed `string` but Node delivers `string | Buffer`.** `extractDbtError` already calls `.toString()` so runtime is fine, but the interface is a minor type lie. Widen to `stdout?: string | Buffer` / `stderr?: string | Buffer`.

packages/dbt-tools/src/dbt-cli.ts
```diff
@@ -305,6 +318,50 @@ export async function execDbtShow(sql: string, limit?: number) {
...
+function extractDbtError(
+  lines: Record<string, unknown>[],
+  primary?: ExecFileError,
+  plain?: ExecFileError,
+): string | undefined {
+  if (!primary && !plain) return undefined
+
+  const errorEvent = lines.find(
+    (l: any) => l.info?.level === "error" || l.level === "error",
+  ) as any
```
suryaiyer95: **NIT — `as any` undercuts the `Record<string, unknown>[]` typing** (consensus across all three external reviewers). Prefer a small typed shape or predicate:
```ts
interface DbtLogLine { info?: { level?: string; msg?: string }; level?: string; msg?: string }
const errorEvent = lines.find((l): l is DbtLogLine =>
  (l as DbtLogLine).info?.level === "error" || (l as DbtLogLine).level === "error")
```

packages/dbt-tools/test/dbt-cli.test.ts
```diff
@@ -149,6 +149,64 @@ describe("execDbtShow", () => {
...
+  // --- Bubble real dbt error instead of generic "Could not parse" ---
+
+  test("surfaces real dbt stderr when run fails", async () => {
+    mockExecFile.mockImplementation((_cmd: string, _args: string[], _opts: any, cb: Function) => {
+      const err: any = new Error("Command failed: dbt show --inline ...")
+      err.code = 1
+      err.stdout = ""
+      err.stderr =
+        "Runtime Error: Failed to read package: No dbt_project.yml found at expected path dbt_packages/dbt_utils/dbt_project.yml"
+      cb(err, err.stdout, err.stderr)
```
suryaiyer95: **This mock is why the tests pass while production fails (see CRITICAL 1).** It attaches `stdout`/`stderr` onto the *error object* (`err.stdout = ...`), then passes them to `cb`. Node/Bun never put `stdout`/`stderr` on the rejected error — they only arrive as the 2nd/3rd callback args, which `run()` discards. A faithful mock should leave them **only** on the callback args:
```ts
const err: any = new Error("Command failed")
err.code = 1
cb(err, "", "Runtime Error: Failed to read package...")  // NOT err.stdout/err.stderr
```
With that mock, this test would currently fail — which is the point: it would catch CRITICAL 1. Add this faithful-wiring test alongside the `run()` fix.

packages/dbt-tools/test/dbt-cli.test.ts
```diff
@@ -149,6 +149,64 @@ describe("execDbtShow", () => {
...
+    })
+    mockExecFile.mockImplementation((_cmd: string, _args: string[], _opts: any, cb: Function) => {
+      const err: any = new Error("Command failed")
+      err.code = 1
+      err.stdout = errorLog
+      err.stderr = "exit status 1"
+      cb(err, err.stdout, err.stderr)
+    })
+
+    await expect(execDbtShow("SELECT 1")).rejects.toThrow(/Compilation Error.*Model 'foo'/)
```
suryaiyer95, line 193: **MINOR — missing coverage for the top-level `{ level: "error" }` shape.** `extractDbtError` handles both `l.info?.level` and top-level `l.level === "error"` (line 349), but this test only emits the nested `info` form. Add a sibling case with `{ level: "error", msg: "..." }` so the `|| l.level === "error"` branch is exercised.

packages/dbt-tools/test/dbt-cli.test.ts
```diff
@@ -149,6 +149,64 @@ describe("execDbtShow", () => {
...
+  test("does not surface generic 'Could not parse' when dbt actually crashed", async () => {
+    mockExecFile.mockImplementation((_cmd: string, _args: string[], _opts: any, cb: Function) => {
+      const err: any = new Error("Command failed")
+      err.code = 2
+      err.stdout = ""
+      err.stderr = "Database Error: connection refused"
+      cb(err, err.stdout, err.stderr)
+    })
+
+    await expect(execDbtShow("SELECT 1")).rejects.not.toThrow(/Could not parse dbt show output/)
```
suryaiyer95: **MINOR — negation-only assertion is weak.** `.rejects.not.toThrow(/Could not parse/)` passes even if the function threw something unrelated (or, post-CRITICAL-1, the wrong generic message). Add a positive assertion of the intended behavior:
```ts
await expect(execDbtShow("SELECT 1")).rejects.toThrow(/Database Error.*connection refused|dbt show failed/)
```

packages/dbt-tools/test/dbt-cli.test.ts
```diff
@@ -149,6 +149,64 @@ describe("execDbtShow", () => {
...
+    mockExecFile.mockImplementation((_cmd: string, _args: string[], _opts: any, cb: Function) => {
+      const err: any = new Error("spawn ENOENT")
+      err.code = "ENOENT"
+      err.stdout = ""
+      err.stderr = ""
+      cb(err, "", "")
+    })
+
+    await expect(execDbtShow("SELECT 1")).rejects.toThrow(/spawn ENOENT|dbt show failed/)
+  })
```
suryaiyer95, line 233: **MINOR — missing test: JSON run crashes but Tier 3 plain-text retry succeeds.** Assert it returns the parsed table and does **not** throw — this proves the real error is surfaced only when *both* tiers fail, and locks in the recovery path.

anandgupta42:

**Panel:** Claude + 5 OpenRouter models (GPT-5.4, Gemini 3.1 Pro, Kimi K2.5, MiniMax M2.7, GLM-5). _(qwen/deepseek/mimo unavailable on the provider, excluded.)_ The panel split — Kimi & MiniMax approved; Gemini, GLM-5 & GPT-5.4 flagged a correctness bug — so I verified against Node's actual behavior. **The bug is real.**

**Verdict: Request changes.** Correct diagnosis and the right approach, but the implementation reads error fields that the codebase's own `run()` never populates, so it largely doesn't deliver the real dbt error it's meant to surface.

### CRITICAL — `run()` discards `stdout`/`stderr`, so the fix can't read them (Bug) — `dbt-cli.ts:54-62` + new reads at `:228, :305` — _Gemini, GLM-5 (verified)_
`run()` rejects the bare `execFile` error:
```ts
execFile(dbt.path, args, opts, (err, stdout, stderr) => {
  if (err) reject(err)        // ← stdout & stderr dropped on the floor
  else resolve({ stdout, stderr })
})
```
Node's `execFile` passes `stdout`/`stderr` as **separate callback arguments**; they are **not** properties of `err` (only `util.promisify(execFile)` attaches them — this is a manual `new Promise`). Gemini and GLM-5 both empirically confirmed `err.stdout === undefined && err.stderr === undefined`. Consequences for the new code:
- `parseJsonLines(primaryRunError.stdout ?? "")` → `parseJsonLines("")` → `[]` → the structured `level:"error"` tier never fires.
- `extractDbtError`'s `primaryStderr`/`plainStderr` are both `undefined` → it falls through to `primary?.message`.

So instead of the intended `Runtime Error: Failed to read package…`, callers get `dbt show failed: Command failed: <path> show --inline …`. That's marginally better than "Could not parse" (it signals a crash) but it does **not** surface dbt's actual error — the PR's stated goal.

**Why the 24 tests don't catch this:** they mock `run()`/the rejection with a hand-constructed error that *has* `.stderr`/`.stdout`. That shape is exactly what the real `run()` never produces, so the suite validates `extractDbtError` in isolation while the integration is broken. (The PR description's claim "the error object carries `.stderr` and `.stdout`" is the same incorrect assumption.)

**Fix** — attach the streams in `run()` before rejecting, then the existing logic (and the existing tests) become correct:
```ts
if (err) {
  ;(err as ExecFileError).stdout = stdout
  ;(err as ExecFileError).stderr = stderr
  reject(err)
} else resolve({ stdout, stderr })
```
Add one integration-style test that drives the **real** `run()` against a non-zero-exit command so this can't silently regress again.

### MINOR — wrong error source when primary exits 0 but the plain-text fallback crashes (Logic Error) — `dbt-cli.ts:308-312` — _GPT-5.4_
If the JSON run exits 0 with unparseable output (the exact case the generic message was designed for) but the Tier-3 plain run then rejects, `extractDbtError(lines, undefined, plainRunError)` returns the plain run's error — so a successful-but-unparseable primary gets reported as a plain-mode crash. Consider only surfacing `realError` when the **primary** run rejected, or label which tier failed.

### MINOR — `ExecFileError.code` type & scope — _Kimi, MiniMax_
`code?: number | string` diverges from Node's `ExecException` (`code?: number`); align it. Also worth restating in the PR: `execDbtCompile`/`execDbtCompileInline` share the same `catch { lines = [] }` masking — fine to defer, but link a follow-up issue so it isn't lost.

### Positive observations
- The root-cause writeup is excellent and correct: conflating "dbt crashed" with "dbt exited 0 but output unparseable" is exactly the bug, and preserving the generic message only for the true exit-0 case is the right contract.
- Preference order (structured `level:"error"` event > stderr > message) is sensible and will work well **once `run()` actually forwards the streams**.
- Good test intent (ENOENT fallback, structured-event preference, generic-message regression guard) — they just need to run against the real `run()` shape.

**Bottom line:** one-line `run()` change unblocks the whole fix and makes the existing tests honest. Without it, the PR ships a better-looking error that still hides the real dbt failure.

sahrizvi:
## Updated to address the second review pass

Pushed `44d9215a4`. Summary:

### Production fixes (`packages/dbt-tools/src/dbt-cli.ts`)

- **`run()` now attaches stdout/stderr to the rejected error.** Node's `execFile` delivers them as separate callback args, not on the error; previously they were discarded, making the catch-block's `e.stdout` / `e.stderr` always `undefined` and rendering the "bubble real dbt error" path inert in production.
- **`execDbtShow` error throw is now split by failure mode.** Distinguishes "primary `dbt show` crashed" (surface the real dbt error) from "primary succeeded but unparseable AND plain-text retry failed" (parser-regression message). The previous combined throw could misattribute a parser regression as a dbt execution failure.
- **Tier 1/2 no longer see crashed-run stdout.** Crash logs can contain incidental arrays that the heuristic scan would return as "rows" (silent wrong data). Tier 1/2 gated behind `!primaryRunError`; crashed stdout is only consulted by `extractDbtError` for structured `level: "error"` events.
- **`extractDbtError`:** picks the **last** `level: "error"` event (dbt emits a generic header first), strips ANSI escapes from surfaced text, and redacts inline SQL from `err.message` via a new `fallbackExitMessage` helper covering exit-code rejections **and** signal/timeout kills. Spawn-time failures (ENOENT etc.) keep their original message.
- **`dbt show failed:` prefix is suppressed** when the underlying error already starts with a dbt category prefix (Compilation/Database/Runtime/Parsing/Validation/Dependency Error).

### Test fixes (`packages/dbt-tools/test/dbt-cli.test.ts`)

- Mocks now pass stdout/stderr only via callback args, matching Node's actual behaviour. Previous mocks pre-attached `err.stdout` / `err.stderr`, which made the inert production path appear to work — a false-positive class the new mocks specifically guard against.
- New coverage: parser-vs-dbt-failure attribution, last-event preference, ANSI strip, SQL-redaction canaries (numeric exit code + signal kill), category-prefix dedup, Buffer-typed stdout/stderr, Tier 3 recovery, anti-spurious-rows on crash.

**Tests: 33 pass / 0 fail.**

### Out of scope — tracked as follow-up issues

- **#943** — `execDbtCompile` / `execDbtCompileInline` mask real dbt errors with generic wrapper message (same bug class as this PR, sibling functions)
- **#944** — `execDbtShow` returns malformed result when dbt emits truthy non-array `data.preview` (pre-existing latent bug, code adjacent to this PR's changes)
- **#945** — `execDbtCompileInline` embeds full inline SQL in error message on failure (same SQL-redaction concern this PR addresses for `execDbtShow`)

Each filed with a focused symptom / root-cause / suggested-fix writeup; happy to take any as a follow-up PR.


---

**Update (`32ffe32af`):** all three follow-up issues are now fixed in this PR.

- **#943 fixed** — `execDbtCompile` and `execDbtCompileInline` now capture the run error, skip success-only tiers on crash, and throw via a new shared `bubbleDbtError(label, primary, plain)` helper that mirrors `execDbtShow`'s error UX (structured JSON event > stderr > redacted exit-status, ANSI-stripped, category-prefix-dedup).
- **#944 fixed** — `execDbtShow` Tier 1 now guards the preview shape with `Array.isArray(preview)` and emits empty rows for any unexpected shape (object, number) instead of returning malformed `{ data: <non-array> }` to callers.
- **#945 fixed** — `execDbtCompileInline` no longer surfaces `e.message` directly (which embedded the full inline SQL via Node's `Command failed: …` format). The new shared helper routes through `fallbackExitMessage`, redacting the command line for both exit-code and signal/timeout failures.

**Tests: 41 pass / 0 fail** (was 33 — added 8 new tests across all three fixes).

dev-punia-altimate:

- connection_refused [1.00ms]
- timeout
- permission_denied
- parse_error
- network_error
- auth_failure
- rate_limit [1.00ms]
- internal_error
- empty_error
- connection_refused
- timeout [1.00ms]
- permission_denied
- parse_error
- network_error
- auth_failure

### Next Step

cc @sahrizvi

packages/dbt-tools/src/dbt-cli.ts
sahrizvi: Fixed in `44d9215a4`. `run()` now attaches stdout/stderr to the rejected error before `reject()`, so the catch block's `e.stdout` / `e.stderr` are populated in production rather than always `undefined`. The tests were rewired to pass stdout/stderr only via callback args (matching Node's actual behaviour), so they will now fail if `run()` regresses.

packages/dbt-tools/src/dbt-cli.ts
sahrizvi: Fixed in `44d9215a4`. Replaced both `e as ExecFileError` casts with a `toExecFileError(e)` helper that does an `instanceof Error` guard and wraps non-Error rejections in `new Error(String(e))`.

packages/dbt-tools/src/dbt-cli.ts
sahrizvi: Fixed in `44d9215a4`. Tier 1 and Tier 2 are now gated behind `!primaryRunError`; the crashed stdout is only consulted by `extractDbtError` for structured `level: "error"` events. New regression test `does NOT return rows from a crashed run's stdout (Tier 1/2 skip on error)` uses a mock whose stdout contains an incidental array-of-objects that would have triggered the silent-wrong-data path pre-fix.

packages/dbt-tools/src/dbt-cli.ts
sahrizvi: Fixed in `44d9215a4`. `extractDbtError` is now invoked only inside `if (primaryRunError || plainRunError)` blocks, so the intent that we only consult it on a real failure is explicit in the call sites.

packages/dbt-tools/src/dbt-cli.ts
sahrizvi: Fixed in `44d9215a4`. `ExecFileError.stdout` / `stderr` widened to `string | Buffer`. Also added a `handles Buffer-typed stdout/stderr from execFile` test that exercises the Buffer branch via `cb(err, Buffer.from(...), Buffer.from(...))`.

packages/dbt-tools/src/dbt-cli.ts
sahrizvi: Fixed in `44d9215a4`. Introduced a typed `DbtLogLine` shape with optional `info?.{level,msg}` / `level` / `msg`, and replaced the `(l as any)` casts in `extractDbtError` with a typed predicate over that shape.

packages/dbt-tools/test/dbt-cli.test.ts
sahrizvi: Fixed in `44d9215a4`. All error-path mocks now pass `stdout`/`stderr` only via callback args (`cb(err, "", "Runtime Error: …")`), never as properties on the err object. Added an inline comment block above the error tests explaining why the faithful wiring matters — both as a guard against this exact regression class and to keep future contributors from reintroducing the pattern.

packages/dbt-tools/test/dbt-cli.test.ts
sahrizvi: Fixed in `44d9215a4`. Added a `recognises top-level { level: 'error' } event shape (no nested info)` test that emits a log line with `{ level: "error", msg: "Database Error: connection to server lost" }` (no nested `info`) and asserts the message is surfaced. Exercises the `|| line.level === "error"` branch.

packages/dbt-tools/test/dbt-cli.test.ts
sahrizvi: Fixed in `44d9215a4`. Added a positive `toThrow(/Database Error.*connection refused/)` assertion alongside the negation. Then in the same commit's R2 sweep, dropped the redundant `.not.toThrow(/Could not parse/)` after a second look — the positive form already implies it.

packages/dbt-tools/test/dbt-cli.test.ts
sahrizvi: Fixed in `44d9215a4`. Added a `Tier 3 recovers: JSON-mode crashes but plain-text succeeds → returns the table` test. First mock invocation rejects, second invocation returns an ASCII table; assertion confirms the recovery path returns parsed rows without throwing.

suryaiyer95 (approved):

PR #937, opened 2026-06-11
fix: auto-resolve question tool in non-interactive contexts

### LINKED ISSUE #936: question tool blocks indefinitely in non-interactive contexts
## Problem

When `altimate-code run "<task>" --yolo` is invoked as a subprocess (Claude Code's Bash tool, CI, `subprocess.run(..., stdin=None)`, agent harnesses), and a skill that uses the `question` tool fires during the run, the process hangs at 0% CPU forever. The harness eventually `TaskStop`s the subprocess, looking identical to a deadlock.

Concretely surfaced during the `data-parity` skill's run when it detected PII columns in `DimCustomer`/`DimEmployee` and (per its safety rules) asked the user *"May I run row-level hashdiff comparisons where sample diff rows may appear if mismatches exist?"* with options *"Approve row diff"* / *"Profile only"*. 70 seconds later, Claude Code TaskStop'd the subprocess.

## Root cause

`packages/opencode/src/tool/question.ts` calls `Question.ask()`, which awaits an Effect `Deferred`. In TUI mode the deferred resolves when the user clicks an option. In `claude --print` / CI / non-interactive contexts, there is no TUI and no user — the deferred awaits forever, while the parent host waits for the subprocess to finish. Neither side has a path to give up.

## Proposed fix

In non-interactive contexts (no TTY, or explicit env-var opt-in), auto-resolve `question` with a conservative-by-default policy and surface the choice clearly in the tool result so the calling LLM can adapt.

**Resolution policy** (env-var controlled, default safe):

- Detect non-interactive: `!process.stdin.isTTY`, with overrides `ALTIMATE_FORCE_INTERACTIVE=1` (use the original interactive `Deferred` path) and `ALTIMATE_NON_INTERACTIVE=1` (force non-interactive even when isTTY is true).
- `ALTIMATE_AUTO_ANSWER=last` (default in non-TTY) — pick the option whose label/description contains a safe keyword (`skip`, `cancel`, `no`, `abort`, `profile only`, `decline`, `deny`, `stop`); fall back to the last option (UX convention: safer/cancel typically sits at the end).
- `ALTIMATE_AUTO_ANSWER=first` — always first option.
- `ALTIMATE_AUTO_ANSWER=skip` — return `Unanswered` for all questions.
- `ALTIMATE_AUTO_ANSWER="<exact label>"` — exact-match an option's label.

The tool result text explicitly flags non-interactive auto-answer (*"Running in non-interactive mode (no TTY). Auto-answered with safe defaults: …"*), so downstream agents can adjust strategy if needed.

## Why not just always pick "cancel"?

Picking cancel/abort blindly fails open in the opposite direction: skills that ask permission to do reasonable work would always get a no, breaking the user's actual intent. The safe-keyword scan tries to match the question author's intent (these are typically *"may I do destructive thing X?"* prompts) without blocking legitimate flows.

## Where this should live

Two choices:
1. **At the tool boundary** (this fix) — `packages/opencode/src/tool/question.ts` short-circuits before calling `Question.ask()`.
2. **At the Effect layer** — push the non-interactive detection into `Question.ask()` itself.

Option 1 is the proposed PR. Option 2 is a deeper, more invasive change and may be the right long-term home — happy to revisit if reviewers prefer that path.

## Test plan

- Unit test (mocked TTY): non-interactive mode + 2-option question → assert tool result names the auto-answered option and tags it as non-interactive.
- Unit test (`ALTIMATE_FORCE_INTERACTIVE=1` + non-TTY): assert original `Question.ask()` path is taken (defers to interactive Deferred).
- Unit test (`ALTIMATE_AUTO_ANSWER=skip`): assert `Unanswered` returned without calling `Question.ask()`.
- Smoke: trigger any skill that uses `question` from a Claude Code Bash tool call; expect completion in seconds, not a hang.

## Context

Surfaced during the same multi-week experiment series that produced #934. Patched locally in `altimate-code-preview` (`packages/opencode/src/tool/question.ts`, ~70 lines + accompanying `test/tool/question.test.ts`); running clean against real workloads for ~9 days. Full writeup: `plugin-skill-experiments/03-issues-and-fixes.md` Issue #5.

PR incoming.

sahrizvi:
Closes #936

## Summary

`packages/opencode/src/tool/question.ts` calls `Question.ask()`, which awaits an Effect `Deferred` that only resolves on a TUI click. When `altimate-code run` is invoked as a subprocess (Claude Code's Bash tool, CI, `subprocess.run`, plugin host) and a skill that uses the `question` tool fires, nobody can ever click — the deferred awaits forever and the parent eventually `TaskStop`s the subprocess. Symptom: 0% CPU, no log activity, no error, indistinguishable from a hang.

This PR short-circuits the `question` tool in non-interactive contexts with a conservative-by-default auto-answer policy.

## Resolution policy

- **Non-interactive detection:** `!process.stdin.isTTY`. Overrides:
  - `ALTIMATE_FORCE_INTERACTIVE=1` — keep the original interactive Deferred path even when isTTY is false.
  - `ALTIMATE_NON_INTERACTIVE=1` — force non-interactive even when isTTY is true (useful for tests + CI assertions).
- **Default in non-TTY (`ALTIMATE_AUTO_ANSWER=last`):** pick the option whose label/description contains a safe keyword (`skip`, `cancel`, `no`, `abort`, `profile only`, `decline`, `deny`, `stop`); fall back to the last option (UX convention: safer/cancel typically sits at the end).
- **Explicit overrides:**
  - `ALTIMATE_AUTO_ANSWER=first` — always pick first option.
  - `ALTIMATE_AUTO_ANSWER=skip` — return `Unanswered` for all questions.
  - `ALTIMATE_AUTO_ANSWER="<exact label>"` — exact-match an option's label (case-insensitive).
- **Tool result prefix reflects mode** — `"Running in non-interactive mode (no TTY). Auto-answered with safe defaults: ..."` vs the original `"User has answered your questions: ..."` — so the agent knows the choice was not a real user answer and can adapt strategy.

## Why not just always pick "cancel"?

Picking cancel/abort blindly fails open in the opposite direction: skills that ask permission to do reasonable work would always get a no, breaking the user's actual intent. The safe-keyword scan tries to match the question author's intent (these are typically *"may I do destructive thing X?"* prompts) without blocking legitimate flows.

## Where this lives

Two reasonable choices:

1. **At the tool boundary (this PR)** — `packages/opencode/src/tool/question.ts` short-circuits before calling `Question.ask()`.
2. **At the Effect layer** — push the non-interactive detection into `Question.ask()` itself, so any caller of `Question.ask` (not just the tool) benefits.

Option 1 is what this PR ships; the scope is contained and the diff is reviewable. Option 2 is the deeper, more invasive change and may be the right long-term home. Happy to refactor if reviewers prefer.

## Test plan

- [x] `bun test test/tool/question.test.ts` → **8 pass, 0 fail**, 12 expect() calls. Verified locally before pushing.
  - 2 pre-existing legacy tests retained (gated with `ALTIMATE_FORCE_INTERACTIVE=1` in `beforeEach` so they preserve their original intent under non-TTY CI).
  - 6 new tests covering: safe-keyword selection, last-option fallback, `ALTIMATE_AUTO_ANSWER=first`, `=skip`, `=<exact label>`, non-interactive prefix wording.
- [ ] Reviewer smoke: trigger any skill using `question` from Claude Code's Bash tool — expect completion in seconds with auto-answered output, not a hang.

## Diff size

```
packages/opencode/src/tool/question.ts       |  81 ++++++++++++++++--
packages/opencode/test/tool/question.test.ts | 123 +++++++++++++++++++++++++++
2 files changed, 198 insertions(+), 6 deletions(-)
```

The patch is roughly 70 lines of source change plus tests. Larger than #935 (the stdin-wedge guard) because the resolution policy has real branching to implement; still bounded to one tool file.

## Risk

Low. Default behavior under TTY is unchanged (the new branch only fires when `!isTTY` or `ALTIMATE_NON_INTERACTIVE=1`). The non-TTY auto-answer surfaces explicitly in the tool result, so a downstream agent treating the choice as a real user click is impossible. `ALTIMATE_FORCE_INTERACTIVE=1` provides an escape hatch for any consumer that wants the original behavior even in non-TTY.

## Links

- Issue: #936
- Original repro: deliverable 02 Run F first sub-session (the `data-parity` skill's PII consent question wedge)
- Catalog entry: `plugin-skill-experiments/03-issues-and-fixes.md` Issue #5
- Sister fix: #935 (stdin wedge in `run.ts` — same surface area, different code path)

---
## Summary by cubic
Fixes hangs in the `question` tool for headless runs by treating `altimate-code run` as non-interactive, defaulting to Unanswered, and clearly labeling the mode. Interactive server flows remain unchanged; addresses #936.

- **Bug Fixes**
  - Env-driven detection: `ALTIMATE_NON_INTERACTIVE=1` enables non-interactive; `ALTIMATE_FORCE_INTERACTIVE=1` forces interactive; `ALTIMATE_NON_INTERACTIVE=0` opts out. `run` sets `ALTIMATE_NON_INTERACTIVE=1` by default (skipped with `--attach`).
  - Non-interactive behavior: return Unanswered for all questions; removed label heuristics. Optional auto-answer via `ALTIMATE_AUTO_ANSWER=first`, `=last`, or `="<exact label>"`.
  - Output: prefix states “non-interactive mode (no answer channel available)”, removes the interactive trailer, and points to `ALTIMATE_AUTO_ANSWER`.
  - `bash` tool strips `ALTIMATE_NON_INTERACTIVE` from child env to avoid breaking nested server reply paths.
  - `run` stdin read is now fully null-safe: use `process.stdin && !process.stdin.isTTY` and skip the read when stdin is undefined to prevent stalls.
  - Tests cover env-var detection, precedence (`FORCE_INTERACTIVE` over `NON_INTERACTIVE`), explicit opt-out (`NON_INTERACTIVE=0`), non-interactive prefix, and suite isolation fixes.

<sup>Written for commit 764291c0c3182aae43918e6cf9a8f345705f2ec8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/937?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Improved non-interactive execution with deterministic question auto-answering (first, last, or exact option label) via configuration, or leaving items unanswered.
  * Mode-specific messaging now clearly indicates when the tool is running non-interactively and how to preconfigure answers.

* **Bug Fixes**
  * Reduced risk of blocking in the run entrypoint by enabling non-interactive mode by default when not attaching.
  * Improved robustness when stdin is unavailable.
  * Prevented non-interactive settings from affecting nested executions.

* **Tests**
  * Added coverage for auto-answer modes, option-label matching, unanswered fallback, and non-interactive vs interactive messaging.

### DIFF
--- a/packages/opencode/src/tool/question.ts
+++ b/packages/opencode/src/tool/question.ts
@@ -3,17 +3,78 @@ import { Tool } from "./tool"
 import { Question } from "../question"
 import DESCRIPTION from "./question.txt"
 
+// altimate_change start — non-interactive auto-answer support.
+// When running under `claude --print`, CI, or any other context without a TTY,
+// there is nobody to click an option in the TUI. The default Question.ask()
+// behaviour is to await a Deferred indefinitely, which causes the parent
+// process to TaskStop the subprocess after a long wait — looking exactly like
+// a hang. See deliverable 02 (Run F first sub-session) for the trace.
+//
+// Resolution policy: in non-interactive mode, pick the option whose label
+// contains a "safe" keyword (skip / cancel / profile only / no / abort).
+// If no such option exists, pick the LAST option (UX convention: safer/cancel
+// usually sits at the end). The agent then sees a concrete answer in the
+// tool result and can continue without blocking. Override via env var:
+//   ALTIMATE_AUTO_ANSWER=first    — always pick first option
+//   ALTIMATE_AUTO_ANSWER=last     — always pick last option (default)
+//   ALTIMATE_AUTO_ANSWER=skip     — return Unanswered for all questions
+const SAFE_KEYWORDS = [
+  "skip",
+  "cancel",
+  "no",
+  "abort",
+  "profile only",
+  "profile-only",
+  "decline",
+  "deny",
+  "stop",
+]
+
+function isNonInteractive(): boolean {
+  if (process.env["ALTIMATE_FORCE_INTERACTIVE"] === "1") return false
+  if (process.env["ALTIMATE_NON_INTERACTIVE"] === "1") return true
+  return !process.stdin.isTTY
+}
+
+function autoAnswer(questions: Question.Info[]): Question.Answer[] {
+  const mode = (process.env["ALTIMATE_AUTO_ANSWER"] ?? "last").toLowerCase()
+  return questions.map((q) => {
+    if (mode === "skip") return []
+    if (mode === "first") return q.options[0] ? [q.options[0].label] : []
+    if (mode === "last") {
+      const safe = q.options.find((o) => {
+        const text = `${o.label} ${o.description}`.toLowerCase()
+        return SAFE_KEYWORDS.some((k) => text.includes(k))
+      })
+      if (safe) return [safe.label]
+      const last = q.options[q.options.length - 1]
+      return last ? [last.label] : []
+    }
+    // exact label match for explicit answers, e.g. ALTIMATE_AUTO_ANSWER="Profile only"
+    const match = q.options.find((o) => o.label.toLowerCase() === mode)
+    return match ? [match.label] : []
+  })
+}
+// altimate_change end
+
 export const QuestionTool = Tool.define("question", {
   description: DESCRIPTION,
   parameters: z.object({
     questions: z.array(Question.Info.omit({ custom: true })).describe("Questions to ask"),
   }),
   async execute(params, ctx) {
-    const answers = await Question.ask({
-      sessionID: ctx.sessionID,
-      questions: params.questions,
-      tool: ctx.callID ? { messageID: ctx.messageID, callID: ctx.callID } : undefined,
-    })
+    // altimate_change start — short-circuit when no human is listening.
+    let answers: Question.Answer[]
+    if (isNonInteractive()) {
+      answers = autoAnswer(params.questions)
+    } else {
+      answers = await Question.ask({
+        sessionID: ctx.sessionID,
+        questions: params.questions,
+        tool: ctx.callID ? { messageID: ctx.messageID, callID: ctx.callID } : undefined,
+      })
+    }
+    // altimate_change end
 
     function format(answer: Question.Answer | undefined) {
       if (!answer?.length) return "Unanswered"
@@ -22,9 +83,17 @@ export const QuestionTool = Tool.define("question", {
 
     const formatted = params.questions.map((q, i) => `"${q.question}"="${format(answers[i])}"`).join(", ")
 
+    // altimate_change start — flag auto-answers explicitly so the agent
+    // knows the user didn't actually answer and can decide whether to
+    // proceed with that choice or fail back gracefully.
+    const prefix = isNonInteractive()
+      ? `Running in non-interactive mode (no TTY). Auto-answered with safe defaults: `
+      : `User has answered your questions: `
+    // altimate_change end
+
     return {
       title: `Asked ${params.questions.length} question${params.questions.length > 1 ? "s" : ""}`,
-      output: `User has answered your questions: ${formatted}. You can now continue with the user's answers in mind.`,
+      output: `${prefix}${formatted}. You can now continue with the user's answers in mind.`,
       metadata: {
         answers,
       },
--- a/packages/opencode/src/tool/question.ts
+++ b/packages/opencode/src/tool/question.ts
@@ -5,33 +5,41 @@ import DESCRIPTION from "./question.txt"
 
 // altimate_change start — non-interactive handling for the question tool.
 //
-// When running under `claude --print`, CI, or any other context without a TTY,
-// there is nobody to click an option in the TUI. The upstream Question.ask()
-// awaits a Deferred indefinitely, which looks exactly like a hang to a parent
-// supervisor.
+// Question.ask() resolves via either a TUI click or an HTTP reply at
+// POST /question/:requestID/reply. Server commands (serve / web / acp /
+// workspace-serve) expose the HTTP path, so an IDE or web client CAN answer
+// even though their stdin is not a TTY. Only `altimate-code run` is
+// genuinely headless: it uses an in-process Server.Default() shim with no
+// bound port, so no client can reach the reply route and Question.ask()
+// awaits forever.
 //
-// Policy: when non-interactive, return Unanswered for every question and let
-// the calling agent decide what to do. The agent knows what it was about to
-// do and why it asked; it can pick a safe path from context or report that
-// user input is required. We deliberately do NOT guess based on label text
-// — every heuristic we tried (safe-keyword scan, last-option fallback)
-// either invented decisions the user didn't make or false-positive'd on
-// labels like "Snowflake" that happened to contain "no".
+// Detection is therefore opt-in via env var rather than TTY-based:
+// `run` sets ALTIMATE_NON_INTERACTIVE=1 on startup; every other entrypoint
+// defaults to interactive. Earlier revisions used !process.stdin.isTTY,
+// which misclassified server mode and silently disabled the HTTP reply path
+// for IDE users (see PR #937 review).
 //
-// Explicit overrides (for users who genuinely want a default and accept the
-// responsibility):
+// Policy when non-interactive: return Unanswered for every question and let
+// the calling agent decide. The agent knows what it was about to do and
+// why it asked; it can pick a safe path from context or report that input
+// is required. We deliberately do NOT guess based on label text — every
+// heuristic we tried (safe-keyword scan, last-option fallback) either
+// invented decisions the user didn't make or false-positive'd on labels
+// like "Snowflake" that happened to contain "no".
+//
+// Explicit overrides (for users who genuinely want a default and accept
+// the responsibility):
 //   ALTIMATE_AUTO_ANSWER=first         — always pick the first option
 //   ALTIMATE_AUTO_ANSWER=last          — always pick the last option
 //   ALTIMATE_AUTO_ANSWER="<label>"     — pick the option whose label matches
 //
-// TTY-detection overrides (for harnesses that lie about being a TTY):
-//   ALTIMATE_FORCE_INTERACTIVE=1       — treat as interactive even without TTY
-//   ALTIMATE_NON_INTERACTIVE=1         — treat as non-interactive even with TTY
+// Mode overrides:
+//   ALTIMATE_FORCE_INTERACTIVE=1       — keep Question.ask() (e.g. tests)
+//   ALTIMATE_NON_INTERACTIVE=1         — set by `run`; opt-in elsewhere
 
 function isNonInteractive(): boolean {
   if (process.env["ALTIMATE_FORCE_INTERACTIVE"] === "1") return false
-  if (process.env["ALTIMATE_NON_INTERACTIVE"] === "1") return true
-  return !process.stdin.isTTY
+  return process.env["ALTIMATE_NON_INTERACTIVE"] === "1"
 }
 
 function autoAnswer(questions: Question.Info[]): Question.Answer[] {
@@ -79,16 +87,17 @@ export const QuestionTool = Tool.define("question", {
 
     const formatted = params.questions.map((q, i) => `"${q.question}"="${format(answers[i])}"`).join(", ")
 
-    // altimate_change start — make the non-interactive case unambiguous to
-    // the agent so it doesn't treat "Unanswered" as a real user choice.
-    const prefix = nonInteractive
-      ? `Running in non-interactive mode (no TTY). No user was available to answer. Either pick a safe path from the context of the action you were about to take, or report that user input is required to proceed — the user can set ALTIMATE_AUTO_ANSWER=first|last|<exact option label> to pre-answer questions in this mode. Result: `
-      : `User has answered your questions: `
+    // altimate_change start — split the whole message per mode. The original
+    // trailer "continue with the user's answers in mind" contradicts the
+    // non-interactive branch which tells the agent no user was available.
+    const output = nonInteractive
+      ? `Running in non-interactive mode (no answer channel available). No user was available to answer. Either pick a safe path from the context of the action you were about to take, or report that user input is required to proceed — the user can set ALTIMATE_AUTO_ANSWER=first|last|<exact option label> to pre-answer questions in this mode. Result: ${formatted}.`
+      : `User has answered your questions: ${formatted}. You can now continue with the user's answers in mind.`
     // altimate_change end
 
     return {
       title: `Asked ${params.questions.length} question${params.questions.length > 1 ? "s" : ""}`,
-      output: `${prefix}${formatted}. You can now continue with the user's answers in mind.`,
+      output,
       metadata: {
         answers,
       },
--- a/packages/opencode/src/tool/question.ts
+++ b/packages/opencode/src/tool/question.ts
@@ -3,32 +3,30 @@ import { Tool } from "./tool"
 import { Question } from "../question"
 import DESCRIPTION from "./question.txt"
 
-// altimate_change start — non-interactive auto-answer support.
-// When running under `claude --print`, CI, or any other context without a TTY,
-// there is nobody to click an option in the TUI. The default Question.ask()
-// behaviour is to await a Deferred indefinitely, which causes the parent
-// process to TaskStop the subprocess after a long wait — looking exactly like
-// a hang. See deliverable 02 (Run F first sub-session) for the trace.
+// altimate_change start — non-interactive handling for the question tool.
 //
-// Resolution policy: in non-interactive mode, pick the option whose label
-// contains a "safe" keyword (skip / cancel / profile only / no / abort).
-// If no such option exists, pick the LAST option (UX convention: safer/cancel
-// usually sits at the end). The agent then sees a concrete answer in the
-// tool result and can continue without blocking. Override via env var:
-//   ALTIMATE_AUTO_ANSWER=first    — always pick first option
-//   ALTIMATE_AUTO_ANSWER=last     — always pick last option (default)
-//   ALTIMATE_AUTO_ANSWER=skip     — return Unanswered for all questions
-const SAFE_KEYWORDS = [
-  "skip",
-  "cancel",
-  "no",
-  "abort",
-  "profile only",
-  "profile-only",
-  "decline",
-  "deny",
-  "stop",
-]
+// When running under `claude --print`, CI, or any other context without a TTY,
+// there is nobody to click an option in the TUI. The upstream Question.ask()
+// awaits a Deferred indefinitely, which looks exactly like a hang to a parent
+// supervisor.
+//
+// Policy: when non-interactive, return Unanswered for every question and let
+// the calling agent decide what to do. The agent knows what it was about to
+// do and why it asked; it can pick a safe path from context or report that
+// user input is required. We deliberately do NOT guess based on label text
+// — every heuristic we tried (safe-keyword scan, last-option fallback)
+// either invented decisions the user didn't make or false-positive'd on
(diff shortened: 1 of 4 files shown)


sahrizvi:
## New commit: redesigned around the actual problem

Pushed `f981199e5`. Pivoted away from the cubic-fix-in-place approach (word-boundary on `"no"`, cache `isNonInteractive()`) toward the cleaner answer:

**Delete the label-text heuristic entirely.** The LLM that constructed the question already knew at construction time which option meant "cancel" — the keyword scan was trying to recover information that was never lost, and false-positived on substrings (`"no"` matched inside `"Snowflake"`, `"Annotate"`, `"Knowledge"`, `"Honor"`).

### What this commit does

- Drops `SAFE_KEYWORDS` and the label-text scan.
- Default non-interactive behavior: every question returns `Unanswered`. `format()` already renders empty-array answers as `"Unanswered"`.
- The non-interactive prefix tells the agent how to proceed AND surfaces the escape hatch:
  > Running in non-interactive mode (no TTY). No user was available to answer. Either pick a safe path from the context of the action you were about to take, or report that user input is required to proceed — the user can set `ALTIMATE_AUTO_ANSWER=first|last|<exact option label>` to pre-answer questions in this mode. Result: ...
- Keeps `=first` / `=last` / `=<exact label>` as explicit user opt-ins for callers who genuinely want a default. Drops `=skip` (it's now the default).
- Caches `isNonInteractive()` once at `execute()` entry so the prefix can't disagree with the path that produced the answer.

### Addresses cubic review feedback

- **P2 (substring `"no"`)**: resolved by deletion. No keyword scan, no substring problem.
- **P3 (mode recomputed post-await)**: cached at the top of `execute()`.

### Tests + verification

- `bun test test/tool/question.test.ts` → **9 pass / 0 fail**.
- Dropped 2 safe-keyword tests; added `=last` opt-in test and `=<unknown label>` fall-through test.
- Independent codex review: "consistently bypasses Question.ask in non-interactive mode, returns empty answers as Unanswered by default, and preserves explicit opt-in overrides. I did not find a discrete regression."

### Honest tradeoff

The prior policy (auto-pick "last" or safe-keyword) tried to make forward progress when no human is present. The new policy says: forward progress on behalf of someone who didn't choose is the worse failure mode — let the agent (which has full context) decide. The escape hatch in the prefix gives the user explicit knobs if they disagree.

sahrizvi:
## Centralized-test bot failures — likely shared infra, not PR-caused

The `dev-punia-altimate` bot reported 15 TypeScript failures (connection_refused / timeout / parse_error / network_error / auth_failure / rate_limit / internal_error / empty_error / oom / permission_denied / ...). The **identical 15-failure set** appears on PRs #933 and #935, which touch unrelated code (`packages/dbt-tools` and `packages/opencode/src/cli/cmd/run.ts` respectively). This PR only modifies `packages/opencode/src/tool/question.ts` + its test file.

Strong signal these are shared fault-injection-harness failures, not regressions introduced by this PR. Flagging here so the pattern is on record across the three PRs; happy to dig into the harness config if the bot owner wants.

packages/opencode/src/tool/question.ts
```diff
@@ -3,17 +3,74 @@ import { Tool } from "./tool"
...
+//   ALTIMATE_AUTO_ANSWER="<label>"     — pick the option whose label matches
+//
+// TTY-detection overrides (for harnesses that lie about being a TTY):
+//   ALTIMATE_FORCE_INTERACTIVE=1       — treat as interactive even without TTY
+//   ALTIMATE_NON_INTERACTIVE=1         — treat as non-interactive even with TTY
+
+function isNonInteractive(): boolean {
+  if (process.env["ALTIMATE_FORCE_INTERACTIVE"] === "1") return false
+  if (process.env["ALTIMATE_NON_INTERACTIVE"] === "1") return true
+  return !process.stdin.isTTY
```
suryaiyer95: **Likely regression for server / IDE (headless-but-interactive) mode — `!process.stdin.isTTY` is the wrong signal for "no human is listening."**

`Question.ask()` does not only resolve on a TUI keypress. Answers also arrive over HTTP: `Question.reply({ requestID, answers })` is exposed at `POST /question/:requestID/reply` (`packages/opencode/src/server/routes/question.ts`), with `GET /question` to list pending ones. So when altimate-code runs as a **server with a frontend/IDE client** (VS Code / JetBrains / web), `process.stdin.isTTY` is false, yet a real human *can* answer via the route.

With this guard, that existing interactive flow is misclassified as non-interactive: `execute()` short-circuits to `autoAnswer()` → returns **Unanswered** immediately and never publishes the question for the client to reply to. The UI user loses the ability to answer, out of the box — `ALTIMATE_FORCE_INTERACTIVE=1` exists but server deployments won't have it set.

**Suggestion:** gate on whether an answer channel actually exists, not on TTY. E.g. treat it as interactive when the server/question clients are connected (a pending-question listener is registered), and reserve the auto-answer path for true headless runs (`run` subprocess / CI). At minimum, default server mode to interactive and require explicit `ALTIMATE_NON_INTERACTIVE=1` to opt into auto-answer, so the `altimate-code run`-as-subprocess fix doesn't also silently disable the HTTP reply path.

suryaiyer95 (commented):
One workflow-affecting concern (verified against the codebase) — posting just this; the rest were quality/description nits.

dev-punia-altimate (commented):

- 0 without a line anchor

#### 1. `packages/opencode/src/tool/question.ts` (L34)

**[🟠 MEDIUM]** There's a potential null pointer exception here. Depending on the runtime environment (e.g., embedded environments or child processes without standard streams), `process.stdin` could be `undefined`. Using optional chaining `process.stdin?.isTTY` is a safer approach and adheres to the null checks guideline.

**Suggested change:**
```suggestion
  return !process.stdin?.isTTY
```

</details>

packages/opencode/src/tool/question.ts
dev-punia-altimate: **[🟠 MEDIUM]** There's a potential null pointer exception here. Depending on the runtime environment (e.g., embedded environments or child processes without standard streams), `process.stdin` could be `undefined`. Using optional chaining `process.stdin?.isTTY` is a safer approach and adheres to the null checks guideline.

**Suggested change:**
```suggestion
  return !process.stdin?.isTTY
```

dev-punia-altimate:

No supported files changed.

anandgupta42:

**Panel:** Claude + 5 OpenRouter models (GPT-5.4, Gemini 3.1 Pro, Kimi K2.5, MiniMax M2.7, GLM-5). _(qwen/deepseek/mimo unavailable on the provider, excluded.)_

**Verdict: Request changes — the fix is correct but incomplete.** The design is right: returning `Unanswered` by default and letting the agent decide (rather than guessing from label text) is the safe call, and the `Snowflake`-contains-`no` regression is the right thing to have a test for. But the same hang still exists one file over.

### MAJOR — the fix misses `plan.ts`, which has the identical wedge (Bug) — `tool/plan.ts:25,99`
`Question.ask()` is called from **three** places: `tool/question.ts` (fixed here) and `tool/plan.ts:25` and `:99` (not touched). Those two still `await` the TUI-only Deferred, so the `plan` tool reproduces the exact 0%-CPU non-interactive hang this PR exists to kill — any non-interactive run that hits the plan tool's question path wedges identically.

This is the concrete argument for the "Option 2" the description already floats: push `isNonInteractive()` / auto-answer **into `Question.ask()`** so all current and future callers are covered by construction. If you keep it at the tool boundary, replicate the guard in both `plan.ts` call sites in this PR — otherwise #936 is only half-closed.

### MINOR — PR description no longer matches the code (Documentation) — _consensus: GPT-5.4, Kimi, MiniMax, GLM-5_
The body still documents the original policy ("default `ALTIMATE_AUTO_ANSWER=last` with safe-keyword scan: skip/cancel/no/abort…"), but the shipped code removed all label heuristics — default is now `Unanswered`, and `=last` simply takes the last option with no safe-keyword scan. The auto-summary already reflects the new behavior; please update the human-written description so reviewers/maintainers aren't misled.

### MINOR — three new env vars, no central registration — _GLM-5_
`ALTIMATE_FORCE_INTERACTIVE`, `ALTIMATE_NON_INTERACTIVE`, `ALTIMATE_AUTO_ANSWER` are read directly via `process.env`. If there's a central env/flag registry (as with `OPENCODE_*` flags), register them there with docs; otherwise at least document the trio in one place so the precedence (`FORCE_INTERACTIVE` > `NON_INTERACTIVE` > `isTTY`) is discoverable.

- Returning `Unanswered` by default instead of inventing an answer is the correct, conservative design — every reviewer called this out. The companion test that asserts `"Snowflake"` is **not** auto-picked is exactly the regression guard this needs.
- `isNonInteractive()` is evaluated once and reused for both the answer path and the result prefix — no double-eval / drift.
- The explicit `"Running in non-interactive mode (no TTY)…"` prefix means a downstream agent can't mistake the auto-result for a real user choice — good.
- `ALTIMATE_FORCE_INTERACTIVE=1` escape hatch + legacy tests gated on it preserves prior behavior under non-TTY CI.

**Bottom line:** safe, well-tested fix for the `question` tool — but close the `plan.ts` sibling (ideally by moving the guard into `Question.ask()`) and refresh the PR description before merge.

sahrizvi:
## Update — addressing review findings (commit d2a851657)

Ran a multi-model consensus review (Claude + GPT 5.4 Codex + Gemini 3.1 Pro + Kimi K2.5 + MiniMax M2.7 + GLM-5.1, 1 convergence round). Resolved 5 findings in this push; 2 deferred to follow-ups.

### Fixed in this commit

| # | Severity | Location | Change |
|---|---|---|---|
| 1 | CRITICAL | `cli/cmd/run.ts:447` | `!process.stdin.isTTY` → `!process.stdin?.isTTY`. Closes the dev-punia-altimate review item in full — the prior commit only resolved it in `tool/question.ts`. |
| 4 | MAJOR | `cli/cmd/run.ts:387` | Gated the `ALTIMATE_NON_INTERACTIVE=1` assignment on `!args.attach`. In attach mode the agent runs on the remote server, so the local env var was a no-op and only polluted the local process env. |
| 3 | MAJOR | `tool/bash.ts:170` | `delete mergedEnv["ALTIMATE_NON_INTERACTIVE"]` before `spawn`. Without this, child processes (e.g. nested `altimate-code serve`) inherited the flag and silently disabled their own HTTP question-reply path — exactly the regression this PR fixes for other surfaces. |
| 5 | MINOR | `test/tool/question.test.ts` | Regression test: `ALTIMATE_FORCE_INTERACTIVE=1` overrides `ALTIMATE_NON_INTERACTIVE=1`. |
| 6 | MINOR | `test/tool/question.test.ts` | Regression test: `ALTIMATE_NON_INTERACTIVE=0` honored as explicit opt-out. |

Test count: 11 → 13. All passing locally (`bun test test/tool/question.test.ts`).

### Deferred to follow-ups

These were flagged in the consensus review but are pre-existing architectural concerns out of scope for this PR:

- **`session/llm.ts:299` permission-layer mismatch.** `resolveTools` filters tool advertisement using only `agent.permission`, not the merged `session.permission`. The `permission: question, action: deny` rule in `run.ts` therefore doesn't filter the tool out of the model's tool list (execution-time deny via `prompt.ts:1494` still fires, but the question tool calls `Question.ask()` directly, bypassing `ctx.ask`). Will track as a separate ticket.
- **Integration test for `run` actually setting the env var on startup.** Unit tests cover the question tool's branching; a spawn-based test would lock in the `run.ts` contract. Same follow-up.

### Convergence review artifacts

Stored at `/tmp/code-review-FSSIM4/` (local). Convergence outcome: 1 APPROVE (Gemini), 3 CHANGES NEEDED (GPT, MiniMax, GLM-5) with overlapping accuracy objections — all incorporated. Specifically:
- Issue #1 line number corrected from 433 → 447 (4 reviewers caught this independently)
- Issue #2 wording corrected from "dead code" to "permission-layer mismatch in tool advertisement"
- Two rejected findings (`afterEach` env restore, output text on serve) confirmed to be non-issues

sahrizvi:
## Follow-up — addressing coderabbit + cubic findings on `d2a851657` (commit `764291c0c`)

Both bots independently flagged the same regression I introduced with my null-safety patch: `!process.stdin?.isTTY` traded the original *crash* for a *stall* (undefined stdin satisfies the guard, then `Bun.stdin.text()` awaits a stream that never EOFs). Plus a related test-isolation gap.

### Changes

| Finding | Location | Change |
|---|---|---|
| coderabbit + cubic | `cli/cmd/run.ts:459` | `!process.stdin?.isTTY` → `process.stdin && !process.stdin.isTTY`. Undefined stdin now skips the read entirely — only sensible behavior when there is no stdin to read from. |
| coderabbit (outside-diff) | `test/tool/question.test.ts:119-127` | `tool.question non-interactive auto-answer` block now deletes `ALTIMATE_FORCE_INTERACTIVE` in both `beforeEach` and `afterEach`. Without this, a parent-shell `FORCE_INTERACTIVE=1` would silently flip the suite to the interactive path and lose all non-interactive coverage. |

### Not adopted from the suggestions

CodeRabbit additionally suggested that the stdin-read should be skipped when a positional `message` argument was already provided. That would *change* current semantics — `altimate-code run "summarize" < data.txt` is a supported pattern that intentionally combines positional + piped content. Out of scope for this fix.

### Tests

13/13 pass on the updated file. All previously-passing question/bash tool tests still pass.

dev-punia-altimate:

- connection_refused
- timeout
- permission_denied
- parse_error [1.00ms]
- network_error
- auth_failure
- rate_limit
- internal_error
- empty_error
- connection_refused
- timeout
- permission_denied [1.00ms]
- parse_error
- network_error
- auth_failure

### Next Step

cc @sahrizvi

packages/opencode/src/tool/question.ts
sahrizvi: Thanks for catching this — your point was the core insight that drove the entire redesign of this PR. Addressed across `f981199`, `d2a85165`, and `764291c0`.

Specifically:

- `tool/question.ts` no longer touches `process.stdin.isTTY` at all. Detection is now opt-in via `ALTIMATE_NON_INTERACTIVE`, and `run` is the only entrypoint that sets it. `serve`, `web`, `acp`, and `workspace-serve` deliberately leave it unset, so their `POST /question/:requestID/reply` path stays live for connected IDE/web clients exactly as you described.
- `cli/cmd/run.ts` sets the env var only when `args.attach` is not set — attach-mode runs the agent on the remote server, so the local env var would just be noise and could mislead other host-local tools.
- `tool/bash.ts` strips `ALTIMATE_NON_INTERACTIVE` from `mergedEnv` before spawning children, so a nested `altimate-code serve` (or any server entrypoint) spawned from a `run` session doesn't inherit the flag and silently disable its own HTTP reply path. This was a latent regression of the same class you flagged — it would have hit the same IDE/web users.

The block comment at `tool/question.ts:6-38` captures the rationale and explicitly mentions PR #937 review for future maintainers. Tests at `test/tool/question.test.ts` lock in the default-interactive contract (the suryaiyer95-named regression class) plus `FORCE_INTERACTIVE` precedence and `NON_INTERACTIVE=0` opt-out.

There's one remaining architectural item your review surfaced indirectly: `session/llm.ts:299` only consults `agent.permission`, not the merged `session.permission`, so the `question: deny` rule in `run.ts` is intent-marker only. Tracked as a follow-up — out of scope here because the env-var path now handles the actual behavior correctly.

suryaiyer95 (approved):

PR #938, opened 2026-06-13
feat: annotate traces with `de.*` attributes via metadata channel + procedural classifier

sahrizvi:
## Summary

Wire the existing trace-attribute scaffolding (`de-attributes.ts`, `setSpanAttributes`, viewer grouped rendering at `viewer.ts:514`) so real values flow onto every tool span and the session root.

Before this change, 0/101 traces on disk had any populated `attributes` field — the schema slot, vocabulary, setter, and viewer rendering all existed but had zero callers. This PR connects the seam.

**Two layers, no LLM:**

- **Layer 1 — tool-provided metadata** (high fidelity, from real driver/parser values). Tools surface structured fields by setting `de.*` keys on their returned `metadata` object. `Trace.logToolCall` lifts those keys onto `span.attributes`. Tools never import the observability layer — the seam is the existing `ToolStateCompleted.metadata` field already plumbed through `message-v2`. Per-value 10 KB + total 32 KB byte caps prevent runaway payloads.

- **Layer 2 — procedural classifier** (`observability/annotator.ts`). Pure function over `(toolName, input, output)`. Lookup-table tool taxonomy, regex bash-intent + dbt-layer-from-path classification, deterministic outcome/artifacts/env rollups. Best-effort — returns `undefined` rather than emitting wrong attributes.

Layer 1 wins over Layer 2 on key conflicts in both per-tool and session merges.

### 1. Tracer reads tool metadata (`tracing.ts:logToolCall`)

`state.metadata` is filtered to keys starting with `de.`, JSON-validated per value, and merged into `span.attributes`. Bounded by per-value (10 KB) and total (32 KB) byte caps so a giant `de.sql.query_text` or lineage array can't balloon snapshots.

### 2. Procedural classifier (`observability/annotator.ts`, new file)

Two pure functions:
- `annotateToolSpan(name, input, output)` — derives `de.tool.*`, `de.dbt.layer`, `de.dbt.command`, `de.sql.query_text`, `de.sql.lineage.input_tables`.
- `annotateSession(trace)` — derives `de.workflow.type` (with confidence), `de.outcome.*`, `de.artifacts.*`, `de.env.*`.

Called from `logToolCall` (per-span), `endTrace` (session rollup), and `flushSync` (crash-path session rollup). Best-effort, never throws.

### 3. Five new sub-namespaces in `de-attributes.ts`

| Namespace | Keys |
|---|---|
| `DE.WORKFLOW` | `type`, `intent`, `type_confidence` |
| `DE.OUTCOME` | `class`, `executed`, `change_applied` |
| `DE.ARTIFACTS` | `files_read`, `files_edited`, `models_mentioned`, `tables_referenced` |
| `DE.ENV` | `dbt_present`, `dbt_manifest_present`, `warehouse_type`, `tools_detected` |
| `DE.TOOL` | `category`, `subcategory`, `vendor`, `bash_intent`, `bash_invoked` |

Plus two new keys in existing namespaces:
- `de.dbt.layer` (staging / intermediate / dim / fact / agg / mart / source / seed / macro / test / snapshot)
- `de.warehouse.rows_total` (for `schema_inspect`'s table-level row count — distinct from query-result `rows_returned`)

### 4. Viewer grouping extension (`viewer.ts:514`)

Added 5 color-coded sections (Workflow, Outcome, Artifacts, Environment, Tool) so the new attributes render with their own labeled groups in the side panel instead of falling through to the generic key-value bucket.

### 5. Four tool opt-ins (Layer 1, high-fidelity)

| Tool | Emits |
|---|---|
| `sql_execute` | `de.warehouse.{system,rows_returned}`, `de.sql.dialect` (resolved from Registry warehouse) |
| `altimate_core_column_lineage` | Structured `de.sql.lineage.{input_tables,output_table,columns_read,columns_written}` from altimate-core's parser — preserves case via structured endpoint extraction (no dot-split that would corrupt quoted identifiers) |
| `schema_inspect` | `de.warehouse.rows_total`, `de.sql.lineage.output_table` |
| `project_scan` | `de.env.{dbt_present,dbt_manifest_present,warehouse_type,tools_detected}` from authoritative scan results |

These four tools never import the observability layer — they set `de.*` keys on their existing returned `metadata` object.

## What this PR explicitly does NOT touch

- `altimate-core` / `altimate-core-internal` — unmodified.
- Tool framework (`tool/tool.ts`) — the metadata channel was already there.
- Message schema (`message-v2.ts`) — already carried `state.metadata`.
- `TraceFile` / `TraceSpan` schema — already had the `attributes` slot.
- The other ~70 tools — covered by Layer 2 (procedural classifier).

## Codex review

Each chunk was reviewed by Codex before moving on. Fixes folded into this PR:

- Relative dbt-layer path matching (normalize backslashes, anchor at `(?:^|/)models/...`)
- SQL-text capping uses `slice` instead of dropping the value
- Workflow classifier counts dbt-bash spans by `de.tool.bash_intent` attribute, not all bash spans
- Scalar `output_table` (omit on multi-output rather than switch type to array)
- Structured endpoint extraction for column lineage (uses `source_table` / `source_column` when present)
- `flushSync` runs the session rollup so crash/interrupted traces get root attrs
- Layered merge direction: explicit `setSpanAttributes(..., \"session\")` callers still win over derived
- `de.warehouse.rows_total` instead of `de.quality.row_count` for schema-inspect totals (quality keys reserved for profiling/check results)

## Test plan

- [x] `bun run typecheck` — clean (pre-existing zod/effect/datamate noise unrelated)
- [x] Local build via `bun run build:local` succeeds; binary runs against the Altimate gateway
- [x] **E2E live run**: prompt that exercises `read` (dbt layer), `bash` (`dbt build` + `wc -c`), and a final assistant turn. Trace shows:
  - `read` span: `de.tool.category=fs`, `de.dbt.layer=staging`
  - `bash` span: `de.tool.bash_intent=dbt`, `de.dbt.command=build`
  - Session root: `de.outcome.class=success`, `de.outcome.executed=true`, `de.artifacts.files_read=[...]`
- [x] **Retroactive validation over the 101 historical traces** via a one-off pure-function pass:
  - 103/104 files annotated, 0 failed
  - **3010/3016 tool spans** (99.8 %) got new `de.*` attributes
  - 102 sessions got root-level rollup
  - Workflow distribution: 51 dbt_develop, 22 dbt_troubleshoot, 8 warehouse_exploration
  - Outcome distribution: 89 success, 11 failure, 2 interrupted (matches the original `summary.status` mix)
  - Real DE tool spans (`sql_execute`, `dbt_profiles`, `warehouse_list`, `warehouse_add`, `schema_inspect`) all received their `de.tool.*` and `de.sql.*` attributes via the procedural classifier
- [ ] CI typecheck + bridge-invariant tests
- [ ] Review the changes against `bridge-guard` if any upstream sync is pending

## Trace size impact

Per-tool-span attributes: bounded by the 32 KB total cap implemented in `logToolCall`.
Session root attributes: small fixed shape (workflow code, outcome enum, file path arrays capped at 100 entries each).
Existing `MAX_SERIALIZED_SPANS=5000` cap is unchanged.

---
## Summary by cubic
Annotates every tool span and the session root with structured `de.*` attributes via a tool metadata channel and a deterministic annotator. Adds new attribute namespaces and viewer groups for workflow, outcome, artifacts, environment, and tool taxonomy.

- New Features
  - Lift `de.*` keys from tool `metadata` onto span attributes with 10 KB per-value and 32 KB total UTF-8 byte caps; preserves caller-set attributes.
  - Add procedural annotator to infer tool taxonomy, bash intent (command-boundary anchored), dbt layer, SQL query/input tables across SQL tools, and session workflow/outcome/artifacts/env; tool-provided values win on conflicts.
  - Extend `de-attributes.ts` with `WORKFLOW`, `OUTCOME`, `ARTIFACTS`, `ENV`, `TOOL`, plus `de.dbt.layer` and `de.warehouse.rows_total`; update viewer groups.
  - Tool opt-ins: `sql_execute` (`de.warehouse.{system,rows_returned}`, `de.sql.dialect`), `altimate_core_column_lineage` (structured lineage with object-form endpoints), `schema_inspect` (`de.warehouse.rows_total`, `de.sql.lineage.output_table`), `project_scan` (`de.env.*`).

- Bug Fixes
  - Bash intent: classify `altimate-dbt` before `dbt`, anchor patterns to command boundaries, and hoist regexes to module scope.
  - SQL extraction: extend to all `sql_*` and `altimate_core_*` SQL tools (plus `sql_diff` tables); remove annotator-side char caps and rely on tracer byte caps; extract bash SQL lineage from the full query.
  - Lineage: guard `column_lineage` with `Array.isArray` + record checks; support object-form `source_column`/`target_column`; strip table prefix in column fallbacks; keep `de.sql.lineage.output_table` scalar.
  - Session rollup: runs on end and crash; `de.outcome.executed` reads cached `de.tool.bash_intent`/`de.dbt.command`; `de.outcome.change_applied` requires a successful `write`/`edit`; workflow counts only bash spans with `de.tool.bash_intent`.
  - Environment: derive `de.env.dbt_manifest_present` from file existence; read `de.env.*` from the latest `project_scan` span (prefer Layer 1 over text); switch tool opt-ins to `DE_*` constants.

<sup>Written for commit 6766457fecb6ea2cca62dfb85ed2eb09f0735d9d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/938?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Traces now automatically enrich tool spans and root sessions with best-effort metadata, including tool taxonomy, bash/dbt intent, dbt layer detection, and SQL input table references.
  * Added new Data Engineering trace attributes for workflow/outcome, artifacts, environment capabilities, and tool classification, and expanded DE grouping in the trace viewer.
  * Enhanced tools to emit richer observability metadata: column lineage, environment capability signals, row totals/lineage output tables, and resolved warehouse system/dialect for SQL runs.
* **Bug Fixes**
  * Improved crash handling to prevent corrupted/overwritten crashed trace exports during overlapping runs.

### DIFF
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -300,9 +300,11 @@ export function annotateToolSpan(
     } else if (toolName === "sql_execute" || toolName === "sql_analyze" || toolName === "sql_optimize" || toolName === "sql_fix" || toolName === "sql_explain" || toolName === "sql_translate" || toolName === "sql_rewrite" || toolName === "sql_format") {
       const q = typeof inp.query === "string" ? inp.query : (typeof inp.sql === "string" ? inp.sql : undefined)
       if (q) {
-        // Cap the stored query text to keep span size bounded, but still extract
-        // lineage from the full query so we don't lose table refs at the tail.
-        out[DE.SQL.QUERY_TEXT] = q.slice(0, 8000)
+        // Don't pre-truncate. The tracer enforces a 10 KB UTF-8 byte cap per
+        // value in `logToolCall`; a char-based slice here was inconsistent
+        // (8000 CJK chars = ~24 KB and gets silently dropped by the byte cap).
+        // Let the tracer be the single authority on byte limits.
+        out[DE.SQL.QUERY_TEXT] = q
         const tables = extractInputTables(q)
         if (tables) out[DE.SQL.LINEAGE_INPUT_TABLES] = tables
       }
@@ -317,12 +319,10 @@ export function annotateToolSpan(
       // Find a sql_execute-shaped SELECT/INSERT/CTE inside the bash command
       const sqlMatch = cmd.match(/(?:^|[\s'"`(])((?:WITH\s+\w[\s\S]{0,200}?AS\s*\(|SELECT|INSERT\s+INTO|UPDATE\s|DELETE\s+FROM|CREATE\s+TABLE|ALTER\s+TABLE)[\s\S]+)/i)
       if (sqlMatch) {
-        // Cap stored query text at 8 KB for span size, but pass the FULL
-        // extracted SQL to extractInputTables so tail table refs aren't lost.
-        // Matches the sql_execute path (line ~287) and addresses Gemini
-        // review on PR #938 — earlier this scanned the truncated text.
+        // No pre-truncation here either — tracer's 10 KB UTF-8 byte cap is
+        // the single source of truth for span-attribute size limits.
         const fullSql = sqlMatch[1]
-        out[DE.SQL.QUERY_TEXT] = fullSql.slice(0, 8000)
+        out[DE.SQL.QUERY_TEXT] = fullSql
         const tables = extractInputTables(fullSql)
         if (tables) out[DE.SQL.LINEAGE_INPUT_TABLES] = tables
       }
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -168,29 +168,37 @@ type BashClassification = {
   dbtCommand?: string
 }
 
+// Hoisted to module scope so the RegExp is compiled once at startup, not on
+// every classifyBash call. (Gemini PR-938 review.)
+const DBT_VERBS = "build|run|test|seed|snapshot|compile|deps|run-operation|debug|parse|docs|clean|list|ls|source|init|show|retry|freshness"
+const DBT_VERB_RE = new RegExp(`\\bdbt\\s+(${DBT_VERBS})\\b`, "i")
+const ALTIMATE_DBT_RE = /\baltimate-dbt\b/i
+const ALTIMATE_DBT_VERB_RE = /\baltimate-dbt\s+([a-z][a-z0-9-]*)\b/i
+const DBT_BARE_RE = /\bdbt\b/
+const CD_PREFIX_RE = /^\s*cd\s+\S+\s*&&\s*/
+
 function classifyBash(command: string | undefined): BashClassification | undefined {
   if (typeof command !== "string" || !command) return undefined
 
   // Strip leading "cd <dir> &&" so we classify the actual work.
-  const stripped = command.replace(/^\s*cd\s+\S+\s*&&\s*/, "").trim()
+  const stripped = command.replace(CD_PREFIX_RE, "").trim()
   if (!stripped) return undefined
 
   // altimate-dbt CLI — MUST be checked before dbt below: `\b` is a word
   // boundary so `-` is non-word and `altimate-dbt build` would otherwise
   // match the dbt-verb regex and get misclassified as intent="dbt".
-  if (/\baltimate-dbt\b/i.test(stripped)) {
-    const verb = stripped.match(/\baltimate-dbt\s+([a-z][a-z0-9-]*)\b/i)
+  if (ALTIMATE_DBT_RE.test(stripped)) {
+    const verb = stripped.match(ALTIMATE_DBT_VERB_RE)
     return { intent: "altimate_dbt", invoked: "altimate-dbt", ...(verb && { dbtCommand: verb[1].toLowerCase() }) }
   }
 
   // dbt CLI: detect verb after `dbt`. Broad list of subcommands.
-  const dbtVerbs = "build|run|test|seed|snapshot|compile|deps|run-operation|debug|parse|docs|clean|list|ls|source|init|show|retry|freshness"
-  const dbtMatch = stripped.match(new RegExp(`\\bdbt\\s+(${dbtVerbs})\\b`, "i"))
+  const dbtMatch = stripped.match(DBT_VERB_RE)
   if (dbtMatch) {
     return { intent: "dbt", invoked: "dbt", dbtCommand: dbtMatch[1].toLowerCase() }
   }
   // dbt invoked but without a recognized verb (e.g. `dbt --version`, `dbt --help`)
-  if (/\bdbt\b/.test(stripped)) {
+  if (DBT_BARE_RE.test(stripped)) {
     return { intent: "dbt", invoked: "dbt" }
   }
 
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -170,11 +170,19 @@ type BashClassification = {
 
 // Hoisted to module scope so the RegExp is compiled once at startup, not on
 // every classifyBash call. (Gemini PR-938 review.)
+//
+// All `dbt` / `altimate-dbt` patterns are anchored to a command-boundary
+// position — start-of-string or a shell separator (`;` `|` `&`). Plain word
+// boundaries (`\b`) would substring-match `dbt` inside string literals,
+// arguments to other commands, etc. — `echo "dbt run"`,
+// `python -c "import dbt"`, `grep dbt file` would all misclassify as
+// intent="dbt" without this guard (GPT PR-938 consensus review M1).
 const DBT_VERBS = "build|run|test|seed|snapshot|compile|deps|run-operation|debug|parse|docs|clean|list|ls|source|init|show|retry|freshness"
-const DBT_VERB_RE = new RegExp(`\\bdbt\\s+(${DBT_VERBS})\\b`, "i")
-const ALTIMATE_DBT_RE = /\baltimate-dbt\b/i
-const ALTIMATE_DBT_VERB_RE = /\baltimate-dbt\s+([a-z][a-z0-9-]*)\b/i
-const DBT_BARE_RE = /\bdbt\b/
+const CMD_BOUNDARY = "(?:^|[;|&])\\s*"
+const DBT_VERB_RE = new RegExp(`${CMD_BOUNDARY}dbt\\s+(${DBT_VERBS})\\b`, "i")
+const ALTIMATE_DBT_RE = new RegExp(`${CMD_BOUNDARY}altimate-dbt(?:\\s|$)`, "i")
+const ALTIMATE_DBT_VERB_RE = new RegExp(`${CMD_BOUNDARY}altimate-dbt\\s+([a-z][a-z0-9-]*)\\b`, "i")
+const DBT_BARE_RE = new RegExp(`${CMD_BOUNDARY}dbt(?:\\s|$)`, "i")
 const CD_PREFIX_RE = /^\s*cd\s+\S+\s*&&\s*/
 
 function classifyBash(command: string | undefined): BashClassification | undefined {
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -471,8 +471,10 @@ export function annotateSession(trace: TraceFile): Record<string, unknown> {
     })
     if (ranDbt) out[DE.OUTCOME.EXECUTED] = true
 
-    // Outcome.change_applied: any write/edit spans?
-    const changed = toolSpans.some((s) => s.name === "write" || s.name === "edit")
+    // Outcome.change_applied: any SUCCESSFUL write/edit spans? The attribute
+    // name says "applied," not "attempted" — a failed write that didn't touch
+    // the file should not flip this to true.
+    const changed = toolSpans.some((s) => (s.name === "write" || s.name === "edit") && s.status === "ok")
     if (changed) out[DE.OUTCOME.CHANGE_APPLIED] = true
   } catch {
     // best-effort
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -242,6 +242,36 @@ function classifyBash(command: string | undefined): BashClassification | undefin
 // SQL fragment extraction from input/output text
 // ---------------------------------------------------------------------------
 
+// Tools that accept a single SQL query in a `query` or `sql` parameter — the
+// annotator lifts that into `de.sql.query_text` + extracts input tables.
+// sql_autocomplete (takes a `prefix`, not full SQL) and sql_diff (takes
+// `original`+`modified`) are handled separately.
+const SQL_QUERY_TOOLS = new Set<string>([
+  "sql_execute",
+  "sql_analyze",
+  "sql_optimize",
+  "sql_fix",
+  "sql_explain",
+  "sql_translate",
+  "sql_rewrite",
+  "sql_format",
+  // altimate_core_* wrappers that consume SQL via a `sql` parameter.
+  // altimate_core_column_lineage has its own Layer-1 opt-in and is excluded.
+  "altimate_core_rewrite",
+  "altimate_core_validate",
+  "altimate_core_check",
+  "altimate_core_grade",
+  "altimate_core_compare",
+  "altimate_core_complete",
+  "altimate_core_correct",
+  "altimate_core_fix",
+  "altimate_core_semantics",
+  "altimate_core_extract_metadata",
+  "altimate_core_query_pii",
+  "altimate_core_prune_schema",
+  "altimate_core_policy",
+])
+
 const SQL_TABLE_PATTERN = /\b(?:FROM|JOIN)\s+([A-Za-z_"`][A-Za-z0-9_."`\-]*)/gi
 
 function extractInputTables(sql: string): string[] | undefined {
@@ -297,7 +327,7 @@ export function annotateToolSpan(
     } else if (toolName === "read" || toolName === "write" || toolName === "edit") {
       const layer = dbtLayerFromPath(typeof inp.filePath === "string" ? inp.filePath : undefined)
       if (layer) out[DE.DBT.LAYER] = layer
-    } else if (toolName === "sql_execute" || toolName === "sql_analyze" || toolName === "sql_optimize" || toolName === "sql_fix" || toolName === "sql_explain" || toolName === "sql_translate" || toolName === "sql_rewrite" || toolName === "sql_format") {
+    } else if (SQL_QUERY_TOOLS.has(toolName)) {
       const q = typeof inp.query === "string" ? inp.query : (typeof inp.sql === "string" ? inp.sql : undefined)
       if (q) {
         // Don't pre-truncate. The tracer enforces a 10 KB UTF-8 byte cap per
@@ -306,6 +336,17 @@ export function annotateToolSpan(
         // Let the tracer be the single authority on byte limits.
         out[DE.SQL.QUERY_TEXT] = q
         const tables = extractInputTables(q)
+        if (tables) out[DE.SQL.LINEAGE_INPUT_TABLES] = tables
+      }
+    } else if (toolName === "sql_diff") {
+      // sql_diff takes `original` and `modified` instead of a single query.
+      // Lift input_tables from both — there's no single canonical `query_text`
+      // to store, so we skip that key.
+      const both = [inp.original, inp.modified]
+        .filter((v): v is string => typeof v === "string" && v.length > 0)
+        .join("\n")
+      if (both) {
+        const tables = extractInputTables(both)
         if (tables) out[DE.SQL.LINEAGE_INPUT_TABLES] = tables
       }
     } else if (toolName === "skill") {
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -309,9 +309,13 @@ export function annotateToolSpan(
       // Find a sql_execute-shaped SELECT/INSERT/CTE inside the bash command
       const sqlMatch = cmd.match(/(?:^|[\s'"`(])((?:WITH\s+\w[\s\S]{0,200}?AS\s*\(|SELECT|INSERT\s+INTO|UPDATE\s|DELETE\s+FROM|CREATE\s+TABLE|ALTER\s+TABLE)[\s\S]+)/i)
       if (sqlMatch) {
-        const sql = sqlMatch[1].slice(0, 8000)
-        out[DE.SQL.QUERY_TEXT] = sql
-        const tables = extractInputTables(sql)
+        // Cap stored query text at 8 KB for span size, but pass the FULL
+        // extracted SQL to extractInputTables so tail table refs aren't lost.
+        // Matches the sql_execute path (line ~287) and addresses Gemini
+        // review on PR #938 — earlier this scanned the truncated text.
+        const fullSql = sqlMatch[1]
+        out[DE.SQL.QUERY_TEXT] = fullSql.slice(0, 8000)
+        const tables = extractInputTables(fullSql)
         if (tables) out[DE.SQL.LINEAGE_INPUT_TABLES] = tables
       }
     }
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -511,12 +511,18 @@ export function annotateSession(trace: TraceFile): Record<string, unknown> {
       if (env.toolsDetected.length > 0) out[DE.ENV.TOOLS_DETECTED] = env.toolsDetected
     }
 
-    // Outcome.executed: was a dbt build/run/test actually performed?
+    // Outcome.executed: was a dbt-family DML command (build/run/test/seed/
+    // snapshot) actually performed? Reads the cached `de.tool.bash_intent`
+    // and `de.dbt.command` attributes set by classifyBash at logToolCall
+    // time — single source of truth, no regex duplication, no verb-list drift.
+    // (Claude PR-938 consensus review M4.)
+    const DBT_EXECUTED_VERBS = new Set(["build", "run", "test", "seed", "snapshot"])
     const ranDbt = toolSpans.some((s) => {
       if (s.name !== "bash") return false
-      const inp = s.input as Record<string, unknown> | undefined
-      const cmd = typeof inp?.command === "string" ? inp.command : ""
-      return /\b(?:dbt|altimate-dbt)\s+(?:build|run|test|seed|snapshot)\b/i.test(cmd)
+      const intent = s.attributes?.[DE.TOOL.BASH_INTENT]
+      if (intent !== "dbt" && intent !== "altimate_dbt") return false
+      const cmd = s.attributes?.[DE.DBT.COMMAND]
+      return typeof cmd === "string" && DBT_EXECUTED_VERBS.has(cmd)
     })
     if (ranDbt) out[DE.OUTCOME.EXECUTED] = true
 
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -175,6 +175,14 @@ function classifyBash(command: string | undefined): BashClassification | undefin
   const stripped = command.replace(/^\s*cd\s+\S+\s*&&\s*/, "").trim()
   if (!stripped) return undefined
 
+  // altimate-dbt CLI — MUST be checked before dbt below: `\b` is a word
+  // boundary so `-` is non-word and `altimate-dbt build` would otherwise
+  // match the dbt-verb regex and get misclassified as intent="dbt".
+  if (/\baltimate-dbt\b/i.test(stripped)) {
+    const verb = stripped.match(/\baltimate-dbt\s+([a-z][a-z0-9-]*)\b/i)
+    return { intent: "altimate_dbt", invoked: "altimate-dbt", ...(verb && { dbtCommand: verb[1].toLowerCase() }) }
+  }
+
   // dbt CLI: detect verb after `dbt`. Broad list of subcommands.
   const dbtVerbs = "build|run|test|seed|snapshot|compile|deps|run-operation|debug|parse|docs|clean|list|ls|source|init|show|retry|freshness"
   const dbtMatch = stripped.match(new RegExp(`\\bdbt\\s+(${dbtVerbs})\\b`, "i"))
@@ -182,14 +190,8 @@ function classifyBash(command: string | undefined): BashClassification | undefin
     return { intent: "dbt", invoked: "dbt", dbtCommand: dbtMatch[1].toLowerCase() }
   }
   // dbt invoked but without a recognized verb (e.g. `dbt --version`, `dbt --help`)
-  if (/\bdbt\b/.test(stripped) && !/\baltimate-dbt\b/.test(stripped)) {
+  if (/\bdbt\b/.test(stripped)) {
     return { intent: "dbt", invoked: "dbt" }
-  }
-
-  // altimate-dbt CLI
-  if (/\baltimate-dbt\b/i.test(stripped)) {
-    const verb = stripped.match(/\baltimate-dbt\s+([a-z][a-z0-9-]*)\b/i)
-    return { intent: "altimate_dbt", invoked: "altimate-dbt", ...(verb && { dbtCommand: verb[1].toLowerCase() }) }
   }
 
   // Python with inline SQL/DuckDB
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -501,14 +501,49 @@ export function annotateSession(trace: TraceFile): Record<string, unknown> {
     if (filesRead.size > 0) out[DE.ARTIFACTS.FILES_READ] = [...filesRead].slice(0, 100)
     if (filesEdited.size > 0) out[DE.ARTIFACTS.FILES_EDITED] = [...filesEdited].slice(0, 100)
 
-    // Environment capabilities (deterministic if project_scan ran — parse its output text)
-    const projectScanSpan = toolSpans.find((s) => s.name === "project_scan")
-    if (projectScanSpan) {
-      const env = detectEnvFromProjectScan(projectScanSpan)
-      if (env.dbtPresent != null) out[DE.ENV.DBT_PRESENT] = env.dbtPresent
-      if (env.manifestPresent != null) out[DE.ENV.DBT_MANIFEST_PRESENT] = env.manifestPresent
-      if (env.warehouseType) out[DE.ENV.WAREHOUSE_TYPE] = env.warehouseType
-      if (env.toolsDetected.length > 0) out[DE.ENV.TOOLS_DETECTED] = env.toolsDetected
+    // Environment capabilities. Prefer the project_scan tool's Layer-1
+    // `de.env.*` attributes (authoritative, set in project-scan.ts:935) over
+    // re-parsing its output text. Falls back to text parsing only when the
+    // Layer-1 attribute is absent.
+    //
+    // Picks the LATEST project_scan span (last one in the array — spans are
+    // pushed in insertion order in logToolCall) so that a session running
+    // project_scan twice (e.g., after adding a warehouse) reflects the
+    // current state, not the initial one. (GPT PR-938 consensus review M2.)
+    const projectScanSpans = toolSpans.filter((s) => s.name === "project_scan")
+    const latestScan = projectScanSpans[projectScanSpans.length - 1]
+    if (latestScan) {
+      const attrs = latestScan.attributes ?? {}
+      // Text-parse fallback computed lazily — only used per key when Layer-1
+      // is absent, so we still skip the regex work when all 4 keys hit Layer 1.
+      let fallback: ScanEnv | undefined
+      const parsed = (): ScanEnv => {
+        if (!fallback) fallback = detectEnvFromProjectScan(latestScan)
+        return fallback
+      }
+      const pickBool = (key: string, fb: () => boolean | undefined): boolean | undefined => {
+        const v = attrs[key]
+        if (typeof v === "boolean") return v
+        return fb()
+      }
+      const pickString = (key: string, fb: () => string | undefined): string | undefined => {
+        const v = attrs[key]
+        if (typeof v === "string" && v) return v
+        return fb()
+      }
+      const pickStringArray = (key: string, fb: () => string[]): string[] => {
+        const v = attrs[key]
+        if (Array.isArray(v) && v.every((x) => typeof x === "string")) return v as string[]
+        return fb()
+      }
+      const dbtPresent = pickBool(DE.ENV.DBT_PRESENT, () => parsed().dbtPresent)
+      const manifestPresent = pickBool(DE.ENV.DBT_MANIFEST_PRESENT, () => parsed().manifestPresent)
+      const warehouseType = pickString(DE.ENV.WAREHOUSE_TYPE, () => parsed().warehouseType)
+      const toolsDetected = pickStringArray(DE.ENV.TOOLS_DETECTED, () => parsed().toolsDetected)
+      if (dbtPresent !== undefined) out[DE.ENV.DBT_PRESENT] = dbtPresent
+      if (manifestPresent !== undefined) out[DE.ENV.DBT_MANIFEST_PRESENT] = manifestPresent
+      if (warehouseType) out[DE.ENV.WAREHOUSE_TYPE] = warehouseType
+      if (toolsDetected.length > 0) out[DE.ENV.TOOLS_DETECTED] = toolsDetected
     }
 
     // Outcome.executed: was a dbt-family DML command (build/run/test/seed/
--- a/packages/opencode/src/altimate/observability/annotator.ts
+++ b/packages/opencode/src/altimate/observability/annotator.ts
@@ -1,0 +1,531 @@
+/**
+ * Procedural trace annotator.
+ *
+ * Pure functions that classify spans and sessions without LLM judgment:
+ *   - `annotateToolSpan(name, input, output)` — per-tool-span attributes
+ *     (de.tool.*, de.dbt.*, de.sql.query_text from input fields)
+ *   - `annotateSession(spans, metadata)` — session-level attributes
+ *     (de.workflow.*, de.outcome.*, de.artifacts.*)
+ *
+ * Both return Record<string, unknown>. Both follow the rule:
+ *   prefer absent attribute over wrong attribute.
+ *   When the classifier isn't confident, return undefined for that key and
+ *   the caller writes nothing.
+ *
+ * Tools never import this module. The tracer calls it from `logToolCall`
+ * (per-span) and `endTrace` (session rollup). Tools opt into the metadata
+ * channel separately by setting `de.*` keys on their returned metadata.
+ */
+
+import { DE } from "./de-attributes"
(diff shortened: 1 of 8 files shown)


sahrizvi:
## What's still on the table (and why we stopped at 4 opt-ins)

The 4 tool opt-ins in this PR were the highest-yield first batch, not exhaustive. The classifier gives every other tool a useful baseline (`de.tool.{category, subcategory, vendor}` plus tool-specific extractions like `de.dbt.layer`, `de.dbt.command`, `de.sql.query_text`), which is why 99.8% of historical tool spans got annotations from Layer 2 alone.

But many tools have **structured truth that lives inside them and gets thrown away before tracing sees it** — those are the candidates for Layer 1 opt-ins in follow-up PRs.

### Concrete opt-in candidates (incremental, each ~10–30 lines, no infra changes)

| Tool | Could emit | Notes |
|---|---|---|
| `sql_analyze`, `altimate_core_validate`, `altimate_core_check` | `de.sql.validation.{valid,error,type_errors}` | Already in `DE_SQL` vocab |
| `sql_optimize`, `sql_fix`, `sql_rewrite` | `de.sql.schema_changes_*`, optimization metrics | Some already in `DE_SQL` |
| `schema_diff` | `de.sql.schema_changes_{detected,details}` | Already in vocab |
| `data_diff` | `de.quality.row_count_delta` | Already in `DE_QUALITY` |
| `finops_query_history`, `finops_expensive_queries` | `de.warehouse.{bytes_scanned, credits_consumed, query_id}`, `de.cost.*` | Vocab already covers these |
| `finops_analyze_credits` | `de.cost.warehouse_compute_usd`, `de.warehouse.credits_consumed` | Already in vocab |
| `altimate_core_grade` | SQL quality grade | Needs a new key (`de.sql.quality_grade`) |
| `dbt_manifest`, `dbt_lineage`, `altimate_core_parse_dbt` | `de.dbt.dag.nodes_*`, per-model `de.dbt.model.*` | Already in `DE_DBT` |
| `altimate_core_classify_pii`, `schema_detect_pii` | New `de.privacy.*` namespace | Needs new sub-namespace |
| `lineage_check`, `impact_analysis` | `de.sql.lineage.*` from real graph | Already in vocab |

That's roughly 20–30 tools with meaningful structured data the classifier can't reach via regex over output text.

### Why we stopped at 4 in this PR

To keep surface area small enough to land cleanly. The point of this PR was to prove the channel works end-to-end — which it does:

- 99.8% of historical tool spans (3010 / 3016 across 104 traces) gained `de.*` attributes from Layer 2 alone
- 102 sessions got root rollup
- Workflow / outcome distributions match the source `summary.status` mix

Each follow-up tool opt-in is independent and incremental — the architecture supports adding them one at a time without touching the tool framework, the tracer hook, or the classifier. Tools still don't import the observability layer; they just set additional `de.*` keys on their existing returned `metadata` object.

### Suggested follow-up order (highest value first)

1. **FinOps tools** — `finops_query_history`, `finops_expensive_queries`, `finops_analyze_credits` — unlock warehouse cost attribution per session
2. **dbt tools** — `dbt_manifest`, `dbt_lineage`, `altimate_core_parse_dbt` — unlock per-model execution metrics
3. **SQL validation tools** — `sql_analyze`, `altimate_core_validate`, `altimate_core_check` — unlock structured validation results
4. **PII tools** — `altimate_core_classify_pii`, `schema_detect_pii` — needs a new `de.privacy.*` sub-namespace, then a small PR per tool

So: deliberate stopping point, not an oversight. The opportunities are real, known, and incrementally addressable.

packages/opencode/src/altimate/observability/viewer.ts
```diff
@@ -516,7 +516,7 @@ function showDetail(span) {
   }
   // DE attributes grouped
   var a = span.attributes || {};
-  var groups = [['de.warehouse.','Warehouse','cyan'],['de.sql.','SQL','secondary'],['de.dbt.','dbt','orange'],['de.quality.','Quality','green'],['de.cost.','Cost','orange']];
+  var groups = [['de.warehouse.','Warehouse','cyan'],['de.sql.','SQL','secondary'],['de.dbt.','dbt','orange'],['de.quality.','Quality','green'],['de.cost.','Cost','orange'],['de.workflow.','Workflow','accent'],['de.outcome.','Outcome','green'],['de.artifacts.','Artifacts','secondary'],['de.env.','Environment','cyan'],['de.tool.','Tool','accent']];
```
dev-punia-altimate, line 519: **[🔵 LOW]** The use of `var` for variable declarations is strictly prohibited. Please use `const` or `let` instead. In this case, since `groups` is not reassigned, `const` is the appropriate choice.

**Suggested change:**
```suggestion
  const groups = [['de.warehouse.','Warehouse','cyan'],['de.sql.','SQL','secondary'],['de.dbt.','dbt','orange'],['de.quality.','Quality','green'],['de.cost.','Cost','orange'],['de.workflow.','Workflow','accent'],['de.outcome.','Outcome','green'],['de.artifacts.','Artifacts','secondary'],['de.env.','Environment','cyan'],['de.tool.','Tool','accent']];
```

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 12 finding(s)

- 11 anchored to a line (posted inline when the comment stream is on)
- 1 without a line anchor

#### 1. `packages/opencode/src/altimate/observability/viewer.ts` (L519)

**[🔵 LOW]** The use of `var` for variable declarations is strictly prohibited. Please use `const` or `let` instead. In this case, since `groups` is not reassigned, `const` is the appropriate choice.

**Suggested change:**
```suggestion
  const groups = [['de.warehouse.','Warehouse','cyan'],['de.sql.','SQL','secondary'],['de.dbt.','dbt','orange'],['de.quality.','Quality','green'],['de.cost.','Cost','orange'],['de.workflow.','Workflow','accent'],['de.outcome.','Outcome','green'],['de.artifacts.','Artifacts','secondary'],['de.env.','Environment','cyan'],['de.tool.','Tool','accent']];
```

#### 2. `packages/opencode/src/altimate/tools/project-scan.ts` (L938)

**[🔵 LOW]** According to the code quality guidelines, the use of `any` type should be avoided. If its use is absolutely necessary, please add a comment explaining why. Otherwise, consider using a more specific type or an interface to improve type safety.

**Suggested change:**
```suggestion
          .map((w: { type?: string }) => (typeof w?.type === "string" ? w.type.toLowerCase() : ""))
```

#### 3. `packages/opencode/src/altimate/tools/sql-execute.ts` (L95-L98)

**[🔵 LOW]** The expression `args.warehouse ?? registered[0]?.name` is evaluated on every iteration of the `.find()` method. Although the array is small, it's better practice to evaluate this target name once before the loop to make the code cleaner and avoid repeated evaluation.

**Suggested change:**
```suggestion
        try {
          const registered = Registry.list().warehouses
          const targetName = args.warehouse ?? registered[0]?.name
          return registered.find((w) => w.name === targetName)
        } catch {
```

#### 4. `packages/opencode/src/altimate/tools/sql-execute.ts` (L108-L112)

**[🔵 LOW]** Since both metadata properties depend on the exact same condition (`warehouseEntry?.type`), they can be combined into a single object spread. This reduces redundancy and makes the code slightly more concise.

**Suggested change:**
```suggestion
          // altimate_change start — de.* attributes lifted onto span by tracer
          "de.warehouse.rows_returned": result.row_count,
          ...(warehouseEntry?.type && { 
            "de.warehouse.system": warehouseEntry.type,
            "de.sql.dialect": warehouseEntry.type 
          }),
          // altimate_change end
```

#### 5. `packages/opencode/src/altimate/tools/altimate-core-column-lineage.ts` (L49)

**[🔵 LOW]** Avoid using the `any` type to comply with TypeScript best practices. Since `edge` properties are accessed dynamically via string index, `unknown` is safe and appropriate here.

**Suggested change:**
```suggestion
        const extractTable = (edge: Record<string, unknown>, side: "source" | "target"): string | undefined => {
```

#### 6. `packages/opencode/src/altimate/tools/altimate-core-column-lineage.ts` (L64-L68)

**[🟠 MEDIUM]** There are two issues here:
1. When falling back to extracting the column from the `endpoint` string, it currently returns the entire string (e.g., `schema.table.column`). To make this consistent with the `direct` extraction (which returns just the column name) and `extractTable` behavior, it should extract only the column part by stripping the table prefix.
2. Replace `Record<string, any>` with `Record<string, unknown>` to comply with the TypeScript rule against using `any`.

**Suggested change:**
```suggestion
        const extractColumn = (edge: Record<string, unknown>, side: "source" | "target"): string | undefined => {
          const direct = edge[`${side}_column`] ?? edge[`${side}Column`]
          if (typeof direct === "string" && direct) return direct
          const endpoint = edge[side]
          if (typeof endpoint === "string") {
            return endpoint.includes(".") ? endpoint.split(".").pop() : endpoint
          }
```

#### 7. `packages/opencode/src/altimate/tools/altimate-core-column-lineage.ts` (L77)

**[🔵 LOW]** Replace `any` with `unknown` to adhere to the rule against using `any` types.

**Suggested change:**
```suggestion
        for (const edge of (data.column_lineage ?? []) as Record<string, unknown>[]) {
```

#### 8. `packages/opencode/src/altimate/tools/altimate-core-column-lineage.ts` (L87-L93)

**[🟠 MEDIUM]** Hardcoding telemetry keys violates the 'No Hardcoding' standard and can lead to inconsistencies if the attribute names change in the future. These keys are already defined as constants. Please import `DE_SQL` from `../observability/de-attributes` and use the exported constants instead.

**Suggested change:**
```suggestion
        if (inputTables.size > 0) lineageAttrs[DE_SQL.LINEAGE_INPUT_TABLES] = [...inputTables].slice(0, 50)
        // Keep output_table scalar — Codex chunk-3 review #5: don't switch attribute
        // type to array when there are multiple outputs. Omit the attribute instead.
        if (outputs.size === 1) lineageAttrs[DE_SQL.LINEAGE_OUTPUT_TABLE] = [...outputs][0]
        if (colsRead.size > 0) lineageAttrs[DE_SQL.LINEAGE_COLUMNS_READ] = [...colsRead].slice(0, 100)
        if (colsWritten.size > 0) lineageAttrs[DE_SQL.LINEAGE_COLUMNS_WRITTEN] = [...colsWritten].slice(0, 100)
        if (args.dialect) lineageAttrs[DE_SQL.DIALECT] = args.dialect
```

#### 9. `packages/opencode/src/altimate/observability/tracing.ts`

**[🟠 MEDIUM]** There are two improvements that can be made here:
1. **Inaccurate byte size computation**: `serialized.length` measures string length (UTF-16 code units), not actual UTF-8 bytes. If multi-byte characters are present, the true payload byte size could exceed the intended limits. Using `Buffer.byteLength(serialized)` evaluates the actual byte length accurately.
2. **Code duplication**: The identical serialization and size-limiting logic is repeated for both Layer 1 and Layer 2. Extracting this into a reusable `addAttribute` helper improves maintainability and keeps the code DRY.

**Suggested change:**
```suggestion
      const addAttribute = (k: string, v: unknown) => {
        if (v === undefined || k in spanAttributes) return
        try {
          const serialized = JSON.stringify(v)
          if (serialized === undefined) return
          // Use Buffer.byteLength to accurately count UTF-8 bytes
          const byteLen = Buffer.byteLength(serialized)
          if (byteLen > ATTR_VALUE_MAX_BYTES) return
          if (totalBytes + byteLen > ATTR_TOTAL_MAX_BYTES) return
          
          // Store original value (matches setSpanAttributes() for
          // consistent overwrite semantics if both paths target the same key).
          spanAttributes[k] = v
          totalBytes += byteLen
        } catch {
          // Bad metadata value must never break the tracer
        }
      }

      // Layer 1: tool-provided structured metadata (high fidelity — driver
      // values, parser output). Filtered to the de.* prefix.
      const rawMetadata = state.metadata
      if (rawMetadata && typeof rawMetadata === "object") {
        for (const [k, v] of Object.entries(rawMetadata)) {
          if (typeof k === "string" && k.startsWith("de.")) {
            addAttribute(k, v)
          }
        }
      }

      // Layer 2: derived classification from (name, input, output). Best-effort
      // procedural — taxonomy lookup, bash intent, dbt layer from path, etc.
      // Tool-provided metadata (Layer 1) wins on conflicts.
      try {
        const derived = annotateToolSpan(toolName, safeInput, isError ? errorStr : outputStr)
        for (const [k, v] of Object.entries(derived)) {
          addAttribute(k, v)
        }
      } catch {
        // Annotator must never break the tracer
      }
```

#### 10. `packages/opencode/src/altimate/observability/annotator.ts` (L179-L180)

**[🟠 MEDIUM]** This dynamic regular expression is instantiated on every call to `classifyBash`, which incurs unnecessary allocation and compilation overhead. Since `dbtVerbs` is a constant string, consider extracting this `RegExp` declaration outside the function body so it is only compiled once, improving performance.

#### 11. `packages/opencode/src/altimate/observability/annotator.ts` (L310-L312)

**[🟠 MEDIUM]** For `bash` tools, `extractInputTables` is called on the sliced 8000-character string `sql`. This contradicts the strategy used for `sql_execute` tools (where the full query `q` is parsed to avoid losing table references at the tail). To maintain consistency and ensure all lineage is captured, you should pass the full extracted string (`sqlMatch[1]`) to `extractInputTables`.

**Suggested change:**
```suggestion
        const sql = sqlMatch[1].slice(0, 8000)\n        out[DE.SQL.QUERY_TEXT] = sql\n        const tables = extractInputTables(sqlMatch[1])
```

#### 12. `packages/opencode/src/altimate/observability/annotator.ts` (L442-L443)

**[🔵 LOW]** According to the coding standards, the use of `!=` is prohibited. Since `dbtPresent` and `manifestPresent` are defined as `boolean | undefined`, you should use strict inequality `!== undefined` instead of `!= null`.

**Suggested change:**
```suggestion
      if (env.dbtPresent !== undefined) out[DE.ENV.DBT_PRESENT] = env.dbtPresent\n      if (env.manifestPresent !== undefined) out[DE.ENV.DBT_MANIFEST_PRESENT] = env.manifestPresent
```

</details>

packages/opencode/src/altimate/tools/project-scan.ts
```diff
@@ -929,6 +929,24 @@ export const ProjectScanTool = Tool.define("project_scan", {
 
     const degradedSuffix = degradedList.length > 0 ? ` (${degradedList.length} degraded)` : ""
 
+    // altimate_change start — trace augmentation: lift environment signals onto
+    // the de.* metadata channel so the session-level rollup gets authoritative
+    // values instead of regex-parsing output text.
+    const deWhTypes = [
+      ...new Set(
+        (schemaCache?.warehouses ?? [])
+          .map((w: any) => (typeof w?.type === "string" ? w.type.toLowerCase() : ""))
```
dev-punia-altimate, line 941: **[🔵 LOW]** According to the code quality guidelines, the use of `any` type should be avoided. If its use is absolutely necessary, please add a comment explaining why. Otherwise, consider using a more specific type or an interface to improve type safety.

**Suggested change:**
```suggestion
          .map((w: { type?: string }) => (typeof w?.type === "string" ? w.type.toLowerCase() : ""))
```

packages/opencode/src/altimate/tools/sql-execute.ts
```diff
@@ -87,9 +87,30 @@ export const SqlExecuteTool = Tool.define("sql_execute", {
...
       // altimate_change end
+      // altimate_change start — trace augmentation: surface driver-reported
+      // values via the de.* metadata channel. Tools never import observability;
+      // they just set keys on the returned metadata object and the tracer's
+      // logToolCall hook lifts `de.*` keys onto the tool span's attributes.
+      const warehouseEntry = (() => {
+        try {
+          const registered = Registry.list().warehouses
+          return registered.find((w) => w.name === (args.warehouse ?? registered[0]?.name))
+        } catch {
```
dev-punia-altimate, line 101: **[🔵 LOW]** The expression `args.warehouse ?? registered[0]?.name` is evaluated on every iteration of the `.find()` method. Although the array is small, it's better practice to evaluate this target name once before the loop to make the code cleaner and avoid repeated evaluation.

**Suggested change:**
```suggestion
        try {
          const registered = Registry.list().warehouses
          const targetName = args.warehouse ?? registered[0]?.name
          return registered.find((w) => w.name === targetName)
        } catch {
```

packages/opencode/src/altimate/tools/sql-execute.ts
```diff
@@ -87,9 +87,30 @@ export const SqlExecuteTool = Tool.define("sql_execute", {
...
         title: `SQL: ${args.query.slice(0, 60)}${args.query.length > 60 ? "..." : ""}`,
-        metadata: { rowCount: result.row_count, truncated: result.truncated },
+        metadata: {
+          rowCount: result.row_count,
+          truncated: result.truncated,
+          // altimate_change start — de.* attributes lifted onto span by tracer
+          "de.warehouse.rows_returned": result.row_count,
+          ...(warehouseEntry?.type && { "de.warehouse.system": warehouseEntry.type }),
+          ...(warehouseEntry?.type && { "de.sql.dialect": warehouseEntry.type }),
+          // altimate_change end
```
dev-punia-altimate, line 115: **[🔵 LOW]** Since both metadata properties depend on the exact same condition (`warehouseEntry?.type`), they can be combined into a single object spread. This reduces redundancy and makes the code slightly more concise.

**Suggested change:**
```suggestion
          // altimate_change start — de.* attributes lifted onto span by tracer
          "de.warehouse.rows_returned": result.row_count,
          ...(warehouseEntry?.type && { 
            "de.warehouse.system": warehouseEntry.type,
            "de.sql.dialect": warehouseEntry.type 
          }),
          // altimate_change end
```

packages/opencode/src/altimate/tools/altimate-core-column-lineage.ts
```diff
@@ -30,9 +30,78 @@ export const AltimateCoreColumnLineageTool = Tool.define("altimate_core_column_l
...
+      // preserves quoted/case-sensitive identifiers. Fall back to splitting
+      // a dotted endpoint string only when the structured fields are absent.
+      const lineageAttrs: Record<string, unknown> = {}
+      if (!isFailure) {
+        const inputTables = new Set<string>()
+        const outputs = new Set<string>()
+        const colsRead = new Set<string>()
+        const colsWritten = new Set<string>()
+
+        const extractTable = (edge: Record<string, any>, side: "source" | "target"): string | undefined => {
```
dev-punia-altimate, line 50: **[🔵 LOW]** Avoid using the `any` type to comply with TypeScript best practices. Since `edge` properties are accessed dynamically via string index, `unknown` is safe and appropriate here.

**Suggested change:**
```suggestion
        const extractTable = (edge: Record<string, unknown>, side: "source" | "target"): string | undefined => {
```

packages/opencode/src/altimate/tools/altimate-core-column-lineage.ts
```diff
@@ -30,9 +30,78 @@ export const AltimateCoreColumnLineageTool = Tool.define("altimate_core_column_l
...
+            // Strip the trailing column segment; preserve original case
+            return endpoint.split(".").slice(0, -1).join(".") || undefined
+          }
+          return undefined
+        }
+        const extractColumn = (edge: Record<string, any>, side: "source" | "target"): string | undefined => {
+          const direct = edge[`${side}_column`] ?? edge[`${side}Column`]
+          if (typeof direct === "string" && direct) return direct
+          const endpoint = edge[side]
+          if (typeof endpoint === "string") return endpoint
```
dev-punia-altimate: **[🟠 MEDIUM]** There are two issues here:
1. When falling back to extracting the column from the `endpoint` string, it currently returns the entire string (e.g., `schema.table.column`). To make this consistent with the `direct` extraction (which returns just the column name) and `extractTable` behavior, it should extract only the column part by stripping the table prefix.
2. Replace `Record<string, any>` with `Record<string, unknown>` to comply with the TypeScript rule against using `any`.

**Suggested change:**
```suggestion
        const extractColumn = (edge: Record<string, unknown>, side: "source" | "target"): string | undefined => {
          const direct = edge[`${side}_column`] ?? edge[`${side}Column`]
          if (typeof direct === "string" && direct) return direct
          const endpoint = edge[side]
          if (typeof endpoint === "string") {
            return endpoint.includes(".") ? endpoint.split(".").pop() : endpoint
          }
```

packages/opencode/src/altimate/tools/altimate-core-column-lineage.ts
```diff
@@ -30,9 +30,78 @@ export const AltimateCoreColumnLineageTool = Tool.define("altimate_core_column_l
...
+          if (typeof endpoint === "object" && endpoint !== null) {
+            const obj = endpoint as Record<string, unknown>
+            const c = obj.column ?? obj.name
+            if (typeof c === "string" && c) return c
+          }
+          return undefined
+        }
+
+        for (const edge of (data.column_lineage ?? []) as Record<string, any>[]) {
```
dev-punia-altimate: **[🔵 LOW]** Replace `any` with `unknown` to adhere to the rule against using `any` types.

**Suggested change:**
```suggestion
        for (const edge of (data.column_lineage ?? []) as Record<string, unknown>[]) {
```

packages/opencode/src/altimate/tools/altimate-core-column-lineage.ts
```diff
@@ -30,9 +30,78 @@ export const AltimateCoreColumnLineageTool = Tool.define("altimate_core_column_l
...
+          if (srcCol) colsRead.add(srcCol)
+          if (tgtCol) colsWritten.add(tgtCol)
+        }
+        if (inputTables.size > 0) lineageAttrs["de.sql.lineage.input_tables"] = [...inputTables].slice(0, 50)
+        // Keep output_table scalar — Codex chunk-3 review #5: don't switch attribute
+        // type to array when there are multiple outputs. Omit the attribute instead.
+        if (outputs.size === 1) lineageAttrs["de.sql.lineage.output_table"] = [...outputs][0]
+        if (colsRead.size > 0) lineageAttrs["de.sql.lineage.columns_read"] = [...colsRead].slice(0, 100)
+        if (colsWritten.size > 0) lineageAttrs["de.sql.lineage.columns_written"] = [...colsWritten].slice(0, 100)
+        if (args.dialect) lineageAttrs["de.sql.dialect"] = args.dialect
```
dev-punia-altimate: **[🟠 MEDIUM]** Hardcoding telemetry keys violates the 'No Hardcoding' standard and can lead to inconsistencies if the attribute names change in the future. These keys are already defined as constants. Please import `DE_SQL` from `../observability/de-attributes` and use the exported constants instead.

**Suggested change:**
```suggestion
        if (inputTables.size > 0) lineageAttrs[DE_SQL.LINEAGE_INPUT_TABLES] = [...inputTables].slice(0, 50)
        // Keep output_table scalar — Codex chunk-3 review #5: don't switch attribute
        // type to array when there are multiple outputs. Omit the attribute instead.
        if (outputs.size === 1) lineageAttrs[DE_SQL.LINEAGE_OUTPUT_TABLE] = [...outputs][0]
        if (colsRead.size > 0) lineageAttrs[DE_SQL.LINEAGE_COLUMNS_READ] = [...colsRead].slice(0, 100)
        if (colsWritten.size > 0) lineageAttrs[DE_SQL.LINEAGE_COLUMNS_WRITTEN] = [...colsWritten].slice(0, 100)
        if (args.dialect) lineageAttrs[DE_SQL.DIALECT] = args.dialect
```

packages/opencode/src/altimate/observability/annotator.ts
```diff
@@ -0,0 +1,531 @@
...
+function classifyBash(command: string | undefined): BashClassification | undefined {
+  if (typeof command !== "string" || !command) return undefined
+
+  // Strip leading "cd <dir> &&" so we classify the actual work.
+  const stripped = command.replace(/^\s*cd\s+\S+\s*&&\s*/, "").trim()
+  if (!stripped) return undefined
+
+  // dbt CLI: detect verb after `dbt`. Broad list of subcommands.
+  const dbtVerbs = "build|run|test|seed|snapshot|compile|deps|run-operation|debug|parse|docs|clean|list|ls|source|init|show|retry|freshness"
+  const dbtMatch = stripped.match(new RegExp(`\\bdbt\\s+(${dbtVerbs})\\b`, "i"))
```
dev-punia-altimate: **[🟠 MEDIUM]** This dynamic regular expression is instantiated on every call to `classifyBash`, which incurs unnecessary allocation and compilation overhead. Since `dbtVerbs` is a constant string, consider extracting this `RegExp` declaration outside the function body so it is only compiled once, improving performance.

packages/opencode/src/altimate/observability/annotator.ts
```diff
@@ -0,0 +1,531 @@
...
+
+    // SQL extracted from bash (overlay onto the bash classification)
+    if (toolName === "bash" && typeof inp.command === "string") {
+      const cmd = inp.command
+      // Find a sql_execute-shaped SELECT/INSERT/CTE inside the bash command
+      const sqlMatch = cmd.match(/(?:^|[\s'"`(])((?:WITH\s+\w[\s\S]{0,200}?AS\s*\(|SELECT|INSERT\s+INTO|UPDATE\s|DELETE\s+FROM|CREATE\s+TABLE|ALTER\s+TABLE)[\s\S]+)/i)
+      if (sqlMatch) {
+        const sql = sqlMatch[1].slice(0, 8000)
+        out[DE.SQL.QUERY_TEXT] = sql
+        const tables = extractInputTables(sql)
```
dev-punia-altimate: **[🟠 MEDIUM]** For `bash` tools, `extractInputTables` is called on the sliced 8000-character string `sql`. This contradicts the strategy used for `sql_execute` tools (where the full query `q` is parsed to avoid losing table references at the tail). To maintain consistency and ensure all lineage is captured, you should pass the full extracted string (`sqlMatch[1]`) to `extractInputTables`.

**Suggested change:**
```suggestion
        const sql = sqlMatch[1].slice(0, 8000)\n        out[DE.SQL.QUERY_TEXT] = sql\n        const tables = extractInputTables(sqlMatch[1])
```

packages/opencode/src/altimate/observability/annotator.ts
```diff
@@ -0,0 +1,531 @@
...
+    }
+    if (filesRead.size > 0) out[DE.ARTIFACTS.FILES_READ] = [...filesRead].slice(0, 100)
+    if (filesEdited.size > 0) out[DE.ARTIFACTS.FILES_EDITED] = [...filesEdited].slice(0, 100)
+
+    // Environment capabilities (deterministic if project_scan ran — parse its output text)
+    const projectScanSpan = toolSpans.find((s) => s.name === "project_scan")
+    if (projectScanSpan) {
+      const env = detectEnvFromProjectScan(projectScanSpan)
+      if (env.dbtPresent != null) out[DE.ENV.DBT_PRESENT] = env.dbtPresent
+      if (env.manifestPresent != null) out[DE.ENV.DBT_MANIFEST_PRESENT] = env.manifestPresent
```
dev-punia-altimate: **[🔵 LOW]** According to the coding standards, the use of `!=` is prohibited. Since `dbtPresent` and `manifestPresent` are defined as `boolean | undefined`, you should use strict inequality `!== undefined` instead of `!= null`.

**Suggested change:**
```suggestion
      if (env.dbtPresent !== undefined) out[DE.ENV.DBT_PRESENT] = env.dbtPresent\n      if (env.manifestPresent !== undefined) out[DE.ENV.DBT_MANIFEST_PRESENT] = env.manifestPresent
```

dev-punia-altimate:

No supported files changed.

anandgupta42:

**Panel:** Claude + 5 OpenRouter models (GPT-5.4, Gemini 3.1 Pro, Kimi K2.5, MiniMax M2.7, GLM-5). _(3 configured models — qwen/deepseek/mimo — were unavailable on the provider and excluded.)_

**Verdict: Request changes.** The architecture is genuinely good and the change is well-scoped, but the classifiers that produce the telemetry are over-eager in ways that will **contaminate the very data this PR exists to populate**, and there are no tests on any of the new logic.

Two findings that other reviewers raised were checked against the code and **dismissed as already-handled**: the `state.metadata` access (the PR adds the `metadata?: Record<string, unknown>` slot, so it typechecks) and the root-span merge (the code mirrors `sessionAttrs` onto **both** the snapshot root and the live root with identical `{...derived, ...existing}` precedence — the crash-path gap is covered). Good.

### MAJOR — `classifyWorkflow` mislabels generic sessions as dbt (Logic Error) — `annotator.ts` (workflow fallback) — _GPT-5.4_
The fallback marks any prompt containing `fix|debug|error` + a `read`/`edit` span as `dbt_troubleshoot`, and `create|add|build|model|refactor` + `write`/`edit` as `dbt_develop` — with **no dbt evidence required**. A plain TypeScript debugging session becomes a "dbt" workflow, polluting the workflow distribution. Gate the fallback on real dbt signal:
```ts
const hasDbtEvidence =
  spans.some((s) => s.attributes?.[DE.DBT.LAYER] !== undefined) ||
  spans.some((s) => ["dbt","altimate_dbt"].includes(s.attributes?.[DE.TOOL.BASH_INTENT] as string)) ||
  skills.some((s) => s.startsWith("dbt-"))
if (hasDbtEvidence && /\b(fix|debug|troubleshoot|broken|failing|error)\b/.test(p) && (toolNames.includes("read") || toolNames.includes("edit")))
  return { type: "dbt_troubleshoot", confidence: 0.5 }
```

### MAJOR — bash SQL extraction is too loose and self-violates "prefer absent over wrong" (Bug) — `annotator.ts` (`extractInputTables` / bash branch) — _GPT-5.4, Gemini 3.1, MiniMax_
`de.sql.query_text` + derived lineage are computed for **every** bash span via a permissive regex. `python script.py "select one"`, heredocs, or log text containing `SELECT`/`WITH` produce false SQL annotations and bogus lineage. Separately, `SQL_TABLE_PATTERN` mis-captures quoted identifiers — Gemini verified `FROM "my table"` extracts `"my` (and backtick/BigQuery identifiers are mishandled). Restrict extraction to known SQL invocation contexts (`psql`, `clickhouse-client`, `duckdb`, `python -c` with SQL) or require a stronger signal; if low-confidence, emit only `de.tool.bash_intent=sql` and skip query text/lineage.

### MAJOR — no tests for any new logic (Testing) — _unanimous (all 5 models)_
The PR adds a 531-line classifier, a new merge-precedence path, crash-path session rollup, and 4 tool emitters with **zero** unit/integration tests. Failures here are silent mis-annotation, not crashes, so they won't surface without tests. Minimum coverage: `annotateToolSpan`/`annotateSession`/`classifyBash`/`dbtLayerFromPath`/`classifyWorkflow`; the metadata-channel lift (tool `de.*` → span attributes incl. the 10 KB/32 KB caps); and merge precedence (tool metadata > derived > explicit `setSpanAttributes(...,"session")`). A contract test for "tool completion carries `metadata`" also de-risks the schema-drift concern GPT-5.4 raised (today it's a healthy-looking trace that silently drops all `de.*` if message-v2 drifts).

### MINOR — `classifyBash` returns `{ intent: "other" }` instead of `undefined` — `annotator.ts` (end of `classifyBash`) — _GLM-5_
Every unmatched bash span gets `de.tool.bash_intent: "other"`, directly contradicting the module's "return undefined rather than emit wrong attributes" docstring and adding noise to a large fraction of spans. Return `undefined` for the no-match case.

### MINOR — `TOOL_TAXONOMY` is hand-maintained with no drift guard — `annotator.ts` (taxonomy table) — _GLM-5_
70+ entries kept in sync by hand; newly added tools silently fall to `category: "generic"`. Consider a CI assertion that every registered `Tool.define()` name has a taxonomy entry (or is explicitly opted out).

### NIT — _MiniMax_
- `skill` subcategory is set from the taxonomy then immediately overwritten to `skill.${name}` — the first assignment is dead.
- `WAREHOUSE_KEYWORDS`/`TOOL_KEYWORDS` match first-substring-wins; a future `snowflake_dw` keyword would be shadowed by `snowflake`. Document or anchor.

- The **metadata channel** (tools set `de.*` keys on their returned metadata; the tracer lifts them; tools never import observability) is a clean seam with no circular deps — praised by every reviewer.
- Best-effort, never-throw error handling on all annotator/lift paths — correct posture for telemetry.
- 10 KB per-value / 32 KB total byte caps prevent runaway span payloads.
- Layer 1 (tool) > Layer 2 (derived) precedence is documented and consistently implemented across per-tool and session merges.
- Strong manual verification: live E2E + retroactive pass over 101 historical traces (3010/3016 spans annotated). Converting that retroactive harness into a committed fixture test would knock out the MAJOR testing gap directly.

**Bottom line:** ship-worthy design; tighten the three over-eager classifiers and add tests before merge, so the dataset doesn't get polluted on day one.

sahrizvi:
Pushed 8 follow-up commits addressing the 8 substantive review findings (skipped the 5 style/low ones per request). Each is its own commit so reviewers can step through them.

## Commit-to-issue mapping

| Commit | Issue | Reviewer |
|---|---|---|
| `fb07061` | #2 Array.isArray guard for `column_lineage` + per-element isRecord narrowing | coderabbit, cubic, Gemini |
| `37306de` | #4 `altimate-dbt` misclassified as `dbt` (regex ordering) | cubic |
| `1de2e64` | #3 Attribute byte caps use UTF-8 byte length, not JS string length | cubic |
| `310ef3d` | #1 `de.env.dbt_manifest_present` derived from `dbtProject.manifestPath` (file existence), not RPC success | cubic, coderabbit |
| `d87be06` | #5 Import `DE_*` constants in all 4 tool opt-ins; drop hardcoded `de.*` strings | Gemini |
| `e303bc1` | #6 Bash SQL lineage scans full query, not 8 KB-truncated slice | Gemini |
| `b44c03a` | #7 Hoist `classifyBash` regexes to module scope | Gemini |
| `f75079c` | #8 `extractColumn` strips table prefix in endpoint-string fallback for consistency with structured path | Gemini |

## Pre-existing issues (flagged but not fixed in this PR)

Codex caught two pre-existing instances of the same Array.isArray bug pattern in `altimate-core-column-lineage.ts` that I did NOT touch in this PR:

1. **`formatColumnLineage` at line 141** — `for (const edge of data.column_lineage)` has the identical unguarded iteration that would throw `TypeError` on a non-array truthy `column_lineage`.
2. **`edgeCount` at line 29** — `data.column_lineage?.length ?? 0` reports misleading counts for non-array truthy values (e.g., a string's `.length`).

Both are real bugs in the same file. Per scope discipline I'm leaving them for a follow-up PR rather than expanding this one.

## Style/low findings deliberately skipped per request

- `var groups` → `const groups` (`viewer.ts:519`)
- `any` → `unknown` in 3 places
- Hoist `args.warehouse ?? registered[0]?.name` out of `.find()`
- Collapse two object spreads gated by the same condition (`sql-execute.ts:112`)
- `!= null` → `!== undefined` (`annotator.ts:443`)

These are project-style nits the linter should be enforcing; happy to address in a follow-up if you'd prefer them in this PR.

## Verification

- Typecheck clean on all 8 changed files (`bun run typecheck` — pre-existing zod/effect/datamate noise unrelated)
- Per-commit smoke tests where applicable:
  - Issue #4: 6-case classification test (altimate-dbt vs dbt) — 6/6 pass
  - Issue #6: synthetic bash with table ref past byte 8000 — table now captured
  - Issue #7: same 6-case test still passes after regex hoisting

sahrizvi:
Pushed 8 more follow-up commits addressing the substantive findings from a multi-model consensus review (4 reviewer models). Each is one commit per issue.

## Major correctness fixes (4)

| Commit | Issue |
|---|---|
| `c09893d` | `classifyBash` regexes were `\b`-anchored, so `dbt` matched inside string literals — `python -c "import dbt"`, `echo "dbt run"`, `grep dbt file` misclassified as `de.tool.bash_intent="dbt"`. Now anchored to command boundaries (start, `;`, `\|`, `&`). 13/13 smoke cases pass including 5 negative cases. |
| `6766457` | `annotateSession` was regex-parsing `project_scan` output text to derive `de.env.*`, ignoring the authoritative Layer-1 attributes that `project-scan.ts` already writes. Also used `find()` so the first scan won instead of the latest. Now reads `de.env.*` from the latest `project_scan` span's attributes; falls back to text parse only when a key is absent. |
| `8ec4b99` | `extractColumn` only handled `direct` as string; `extractTable` handled it as string OR `{ table, name }` object. Added the matching object-form branch so columns aren't silently dropped while their table is captured. |
| `3cda78f` | `de.outcome.executed` regex duplicated `classifyBash`'s dbt detection with a stale verb subset, and re-parsed raw `input.command` (still vulnerable to the substring bug the M1 commit just fixed). Now reads cached `de.tool.bash_intent` and `de.dbt.command` attributes — single source of truth. Intentionally keeps the 5-verb DML scope (build/run/test/seed/snapshot) for "executed". |

## Minor consensus fixes (4)

| Commit | Issue |
|---|---|
| `9220058` | Annotator was capping `de.sql.query_text` at 8000 chars; tracer caps at 10000 UTF-8 bytes. CJK SQL of 8000 chars = 24000 bytes → silently dropped. Removed annotator-side char cap; tracer is now the single byte-limit authority. |
| `81ae7b9` | SQL-text/lineage extraction covered only 8 `sql_*` tools, missing 13 `altimate_core_*` wrappers that consume SQL via a `sql` parameter (`altimate_core_rewrite`, `altimate_core_validate`, etc.). Refactored to a module-scope `SQL_QUERY_TOOLS` Set + added `sql_diff` special handling for its `original`/`modified` pair. |
| `6b3cac5` | `de.outcome.change_applied` was set even for errored write/edit spans. The attribute says "applied," not "attempted" — filtered on `s.status === "ok"`. |
| `5ea44cd` | `dbt_troubleshoot` heuristic checked `read`/`edit` but not `write`. A session that applied a fix via `write` was silently penalized. One-word fix to match `dbt_develop`'s symmetric check. |

## Out of scope (intentionally deferred)

- `setSpanAttributes` (pre-existing setter) lacks byte cap → broader refactor.
- Greedy `[\s\S]+` SQL bash-extraction regex → best-effort, 8 KB output cap limits damage.
- `WAREHOUSE_KEYWORDS` first-match → auto-mitigated by the M2 fix.
- Missing unit tests for `annotateToolSpan` / `annotateSession` / crash-path → testing follow-up.
- Style nits (`!= null`, redundant `typeof k` check, `var groups` in viewer).
- 5 pre-existing bugs in adjacent code (`buildTraceFile` perf, `formatColumnLineage` Array.isArray, `edgeCount` count, `formatLineageEndpoint` validation, `snapshot()` blocking serialization) — separate cleanup PR.

## Verification

- All 8 commits typecheck clean
- Per-commit smoke tests where applicable:
  - M1 classifyBash: 13/13 (8 positive + 5 negative — `python -c "import dbt"` etc.)
  - M2 project_scan: 3/3 (Layer-1 wins, text fallback, latest-scan-wins)
  - M3 extractColumn: covered via diff inspection (mirrors existing extractTable)
  - M4 ranDbt: 7/7 (3 positive + 4 negative including the M1 cases)
  - m5 byte caps: behavior change verified inline (ASCII 8001-10000 chars now persist)
  - m6 SQL tools: 4/4 (altimate_core_rewrite, altimate_core_validate, sql_diff, sql_autocomplete)
  - m9 dbt_troubleshoot: covered via diff (one-token fix)
  - m7 change_applied: covered via diff (one-condition fix)

Same per-issue cadence as the previous round; happy to address the deferred items in follow-up PRs if you'd like.

dev-punia-altimate:

- connection_refused
- timeout
- permission_denied [1.00ms]
- parse_error
- network_error
- auth_failure
- rate_limit
- internal_error
- empty_error
- connection_refused
- timeout
- permission_denied
- parse_error [1.00ms]
- network_error
- auth_failure

### Next Step

cc @sahrizvi

PR #940, opened 2026-06-14
fix: trigger auto-update check on headless `serve` startup

ralphstodomingo:
PINEAPPLE

## Summary

Triggers a best-effort auto-update check when headless `altimate serve` starts, so extension-launched servers follow the same upgrade path as TUI.

Updated implementation details to match the final code:

- Schedules a one-time startup check (~1s after server listen) via a dedicated helper.
- Runs the check in `Instance.provide({ directory: process.cwd(), init: InstanceBootstrap, fn: ... })` (no `bootstrap()` wrapper), avoiding unintended `Instance.dispose()` teardown during in-process server runtime.
- Keeps all update policy inside `upgrade()` (autoupdate gating, install method handling, downgrade guard, telemetry).
- Uses explicit error handling/logging for both upgrade execution failures and outer orchestration failures.
- Extracts logic into `packages/opencode/src/cli/cmd/serve-upgrade-check.ts`.
- Adds dependency injection for collaborators in the helper to avoid process-global module mocking in tests.

## Test Plan

- Ran targeted tests for startup upgrade check behavior:
  - delayed scheduling
  - `process.cwd()` instance usage
  - non-throwing behavior when upgrade/provide fail
- Verified combined CLI test coverage after mock isolation fix:
  - `packages/opencode/test/cli` suite passing (`512 pass, 0 fail`).
- Verified CI TypeScript failure caused by leaked `mock.module` state is resolved by DI-based test rewrite.

## Checklist

- [x] Tests added/updated
- [ ] Documentation updated (if needed)
- [ ] CHANGELOG updated (if user-facing)

### DIFF
--- a/packages/opencode/src/cli/cmd/serve-upgrade-check.ts
+++ b/packages/opencode/src/cli/cmd/serve-upgrade-check.ts
@@ -44,18 +44,20 @@ const defaultDeps: StartupUpgradeDeps = {
 }
 
 /**
- * Runs a single best-effort upgrade check. Resolves, never rejects: `run()`
- * (upgrade) swallows its own errors via the inner catch; the outer guard only
- * fires for an `Instance.provide` failure. A flaky network/registry therefore
- * can't take the server down.
+ * Runs a single best-effort upgrade check. Resolves, never rejects, via two
+ * layers: the inner `.catch` swallows any error from `run()` (`upgrade()` can
+ * throw, e.g. from `Config.global()` / `Installation.method()`), and the outer
+ * try/catch guards an `Instance.provide` (bootstrap) failure. A flaky
+ * network/registry therefore can't take the server down. Errors are passed as
+ * `Error` objects so `Log` formats the message + `cause` chain.
  */
 export async function runStartupUpgradeCheck(deps: StartupUpgradeDeps = defaultDeps): Promise<void> {
   try {
     await deps.provide(process.cwd(), () =>
-      deps.run().catch((err) => log.error("startup upgrade check failed", { error: String(err) })),
+      deps.run().catch((err) => log.error("startup upgrade check failed", { error: err })),
     )
   } catch (err) {
-    log.error("startup upgrade instance failed", { error: String(err) })
+    log.error("startup upgrade instance failed", { error: err })
   }
 }
 
--- a/packages/opencode/src/cli/cmd/serve-upgrade-check.ts
+++ b/packages/opencode/src/cli/cmd/serve-upgrade-check.ts
@@ -10,33 +10,50 @@ const log = Log.create({ service: "serve" })
 export const STARTUP_UPGRADE_DELAY_MS = 1000
 
 /**
- * Runs a single best-effort upgrade check, mirroring the TUI worker
- * (`cli/cmd/tui/worker.ts` → `checkUpgrade`): provide an `Instance` context for
- * `upgrade()` but — exactly like the worker — do NOT dispose it.
+ * Collaborators for {@link runStartupUpgradeCheck}, injectable for tests.
  *
- * `upgrade()` reads only global config + `Installation`, but `Bus.publish`
- * (its update notifications) needs an ambient instance. We use the same
- * `process.cwd()` key the server's default-directory requests use
- * (`server/server.ts:196`), so we reuse/seed that shared cached instance rather
- * than a throwaway one, and Bus notifications still reach default-directory SSE
- * subscribers.
- *
- * Crucially we never dispose: an earlier version wrapped this in `bootstrap()`,
- * whose `finally` → `Instance.dispose()` tears down the entire `process.cwd()`
- * bucket — including state created by concurrent server requests that defaulted
- * to that directory (use-after-dispose / needless churn). The worker avoids
- * this by running in a separate thread; in-process we avoid it by not disposing.
- *
- * Resolves, never rejects: `upgrade()` swallows its own errors via the inner
- * catch; the outer guard only fires for an `Instance.provide` failure.
+ * Injected rather than module-mocked on purpose: bun's `mock.module` is
+ * process-global, so mocking `../upgrade` / `../../project/instance` here would
+ * clobber those modules for every other test in the run (e.g. blow away
+ * `cli/upgrade`'s `compareVersions`/`isValidVersion` exports).
  */
-export async function runStartupUpgradeCheck(): Promise<void> {
+export interface StartupUpgradeDeps {
+  /**
+   * Runs `fn` inside an ambient `Instance` for `directory` and, like the TUI
+   * worker, does NOT dispose it (see note on the default below).
+   */
+  provide: (directory: string, fn: () => Promise<unknown>) => Promise<unknown>
+  /** The upgrade check itself. */
+  run: () => Promise<unknown>
+}
+
+const defaultDeps: StartupUpgradeDeps = {
+  // Mirror the TUI worker (cli/cmd/tui/worker.ts → checkUpgrade): provide an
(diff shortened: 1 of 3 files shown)


anandgupta42:
## Code Review — `/code-review` (Claude)

**Verdict: Request changes** — one real concurrency/lifecycle risk to resolve first. The change is otherwise clean, well-commented, and correctly delegates all update policy to `upgrade()`.

### MAJOR — the transient `Instance` disposes state the live server shares (Bug / Logic Error) — `serve.ts:59-61`

This runs **in-process** with the HTTP server, unlike the TUI path it mirrors (`worker.checkUpgrade` runs in a separate worker thread, so its `Instance.dispose()` is isolated). Here it is not:

- The server resolves each request's directory as `c.req.query("directory") || header || process.cwd()` and wraps the handler in `Instance.provide({ directory, … })` (`server/server.ts:196,210`). Requests that don't pass a directory default to **`process.cwd()`**.
- `Instance`/`State` cache per-directory state in a process-global map keyed by the directory string (`project/state.ts:10` `recordsByKey`).
- `bootstrap(process.cwd(), …)` runs `upgrade()` and then, in its `finally`, calls `Instance.dispose()` → `State.dispose(process.cwd())` + `disposeInstance(process.cwd())` (`project/instance.ts:141-144`), which disposes the **entire `process.cwd()` bucket** — including state created by server requests that defaulted to that directory.

Net effect: ~1s after the server starts listening, the upgrade check tears down the default-directory instance state. Best case it's lazily recreated on the next request (needless churn of config/storage/connections); worst case it races an in-flight request on `process.cwd()` → use-after-dispose.

**Suggested fix:** `upgrade()` only needs global config + `Installation` (both global, not directory-scoped), so give the check its own isolated instance key that the server never uses, instead of `process.cwd()`:

```ts
import { tmpdir } from "os"
import { join } from "path"
// a sentinel dir the request path will never resolve to
const UPGRADE_INSTANCE_DIR = join(tmpdir(), "altimate-code-upgrade-check")
setTimeout(() => {
  bootstrap(UPGRADE_INSTANCE_DIR, () =>
    upgrade().catch((err) => log.error("startup upgrade check failed", { error: String(err) })),
  ).catch((err) => log.error("startup upgrade bootstrap failed", { error: String(err) }))
}, 1000).unref?.()
```

Please confirm `upgrade()` has no directory-scoped dependency (it reads `Config.global()` and `Installation.*`, which are global) — if so, the sentinel-dir fix fully removes the collision.

### MINOR — magic `1000` delay — `serve.ts:62`
The 1s delay is an unexplained constant. A named const (`const STARTUP_UPGRADE_DELAY_MS = 1000`) with a one-line "let the listener settle before the best-effort check" makes intent explicit.

### NIT — double-logging path
`upgrade()` is already `.catch`-wrapped, so the inner catch handles upgrade failures and the outer `.catch` only ever fires for `bootstrap`/`dispose` errors. That's fine and intentional, but worth a comment so a future reader doesn't think one is dead code.

### Positives
- Correct marker hygiene (`altimate_change start/end`) on every added block.
- Faithfully mirrors the existing TUI trigger and keeps **all** policy inside `upgrade()` (autoupdate gate, downgrade guard, method detection, notify) — no policy duplication.
- Fire-and-forget with `.unref?.()` and logged-not-thrown errors: a flaky network/registry can't take the server down. Good instinct.

### Missing tests
Hard to unit-test a detached `setTimeout`, but consider extracting the trigger body into a small named function so a test can assert it calls `upgrade()` and never rejects. At minimum, a manual note in the PR on how you verified the extension fleet now updates (since that's the whole motivation) would help.

ralphstodomingo:
Thanks — the MAJOR is a real bug, confirmed all three claims against the code (`bootstrap()` `finally` → `Instance.dispose()` at `cli/bootstrap.ts:12`; server default dir = `process.cwd()` at `server/server.ts:196`; per-directory global cache in `project/instance.ts`). Fixed in 49f64ff.

**On the fix — I went a slightly different route than the sentinel dir, and I think it's strictly better; flagging so you can sanity-check:**

The root issue isn't *which* directory we dispose — it's that we dispose at all while in-process with the server. The TUI path I was mirroring (`tui/worker.ts` `checkUpgrade`) uses **raw `Instance.provide` and never disposes**; my mistake was reusing `bootstrap()`, which adds the `finally` dispose. So the faithful mirror is just `Instance.provide({ directory: process.cwd(), init: InstanceBootstrap, fn: upgrade })` with no dispose.

Why this over the sentinel `tmpdir()` key:
- **No teardown at all** — addresses the root concern more directly than relocating the dispose to a throwaway bucket.
- **No throwaway instance on a non-existent dir** — `InstanceBootstrap` spins up LSP / FileWatcher / Vcs / Snapshot against `Instance.directory`; pointing that at a bogus tmpdir path is wasteful and a bit risky. Reusing the `process.cwd()` instance the server already seeds avoids that.
- **Bus notifications still land** — `Bus.publish` fans out to the instance PubSub *and* `GlobalBus` tagged with the directory; keeping `process.cwd()` means `UpdateAvailable`/`Updated` still reach default-directory SSE subscribers, which a sentinel key would orphan.

Confirmed `upgrade()` has no directory-scoped dependency: it reads `Config.global()` (process-global `lazy()`) + `Installation.*` (global); the instance exists only so `Bus.publish` has an ambient context.

**MINOR / NIT / testability** — all addressed: extracted the trigger into `serve-upgrade-check.ts` with named `STARTUP_UPGRADE_DELAY_MS` (+ the "let the listener settle" rationale), a comment explaining the inner-vs-outer catch split, and unit tests asserting it runs `upgrade()` exactly once in the `process.cwd()` instance and never rejects when `upgrade()` throws. 114 existing upgrade/installation tests still green.

**Verification note** (re: "how did you confirm the fleet updates"): the serve-side trigger can't unstick the *existing* 0.7.3 fleet on its own — they won't have this code until they upgrade once (catch-22). That half is handled from the extension in AltimateAI/vscode-altimate-mcp-server#369, which version-checks and re-installs on activation. End-to-end, a stuck machine gets pulled forward by #369; #940 keeps every future serve self-updating.

ralphstodomingo:
Pushed fixes for the CI failure and the remaining review comments.

**CI red (TypeScript job) — fixed (`d0b7b6e`).** The failure was `SyntaxError: Export named 'isValidVersion' not found in module cli/upgrade.ts`, caused by my test: bun's `mock.module` is **process-global**, so mocking `../upgrade` clobbered `compareVersions`/`isValidVersion` for `upgrade-decision.test.ts` in the combined run (passed locally only because I ran the file in isolation). Exactly the leak @Copilot flagged on the test. Rewrote the test to use **dependency injection** — `runStartupUpgradeCheck` takes an optional `StartupUpgradeDeps` ({ provide, run }) defaulting to the real collaborators; the test injects fakes, no global module state touched. Verified in the combined suite + full `test/cli`: **512 pass, 0 fail**.

**Copilot nits — fixed (`e6c26a1`):**
- **Unused `Installation` import** — removed. Also removed `Workspace` and `Project` in the same file (same pre-existing dead-import issue, just not flagged).
- **`String(err)` loses stack/cause** — correct; `Log.build` special-cases `Error` via `formatError` (walks `.cause`). Switched both `log.error` calls to pass the raw `error: err`.
- **Inaccurate "swallows its own errors" comment** — reworded to attribute the never-rejects guarantee to the explicit inner `.catch` wrapper + outer try/catch, not to `upgrade()`.

**Dispose bug (MAJOR / coderabbit / cubic)** — already resolved in `49f64ff`: no longer wraps in `bootstrap()` (whose `finally` disposes), now mirrors the TUI worker with raw `Instance.provide` on `process.cwd()` and no dispose. Rationale for choosing that over the sentinel-dir suggestion is in my earlier reply.

PR #942, opened 2026-06-15
feat(install): verify release archive checksums in both installers

mdesmet:
## What

Follow-up to #930 (raised in review by @coderabbitai and the consensus panel): publish a checksums file with releases and verify downloaded archives in **both** the curl/bash and PowerShell installers.

**Stacked on `feat/windows-powershell-installer`** (base of this PR) because `install.ps1` only exists on that branch. Merge #930 first, then this — or rebase onto `main` after #930 lands.

## Changes

- **`.github/workflows/release.yml`** — generate `checksums.txt` (`sha256sum *.tar.gz *.zip`) and publish it as a release asset alongside the archives.
- **`install` (bash)** — `verify_checksum()` fetches `checksums.txt`, looks up the archive's expected hash, and compares (`sha256sum` or `shasum -a 256`) before extracting.
- **`install.ps1`** — `Test-Checksum` fetches `checksums.txt` and compares `Get-FileHash -Algorithm SHA256` before `Expand-Archive`. Replaces the deferral note from #930.
- **Behavior:** hard-fail on mismatch; **soft-skip** (with a notice) when `checksums.txt` is absent (releases predating this change) or unreachable, so existing version-pinned installs keep working.
- **Tests** — `checksum-verification.test.ts` asserts the release publishes the file and both installers fetch + compare + hard-fail on mismatch.

## Verification

- `bash -n install` clean; `install.ps1` parses clean and the Pester suite (6/6) still passes on PowerShell 7.6.2.
- TS: `checksum-verification.test.ts` green (57 pass across the install/branding set).
- The mismatch/skip paths are content-asserted here and will get live exercise on the first release that ships `checksums.txt`.

---
## Summary by cubic
Adds SHA256 verification to both installers and ships a `checksums.txt` with each release. PowerShell pins the archive and checksum to the same tag; both installers hard-fail on mismatch and soft-skip when checksums are missing or unreachable.

- **New Features**
  - Release workflow generates and uploads `checksums.txt` for all archives.
  - Bash `install` and PowerShell `install.ps1` fetch `checksums.txt` and verify before extraction; cross-platform SHA tools. PowerShell pins archive and `checksums.txt` to the same resolved tag to avoid a latest/ race.

- **Bug Fixes**
  - PowerShell: decode Byte[] `Invoke-WebRequest` bodies on PS 5.1; ASCII-only so the script parses on Windows PowerShell 5.1.
  - Bash: derive `checksums.txt` from the archive URL; safe cleanup on checksum mismatch with guards against "." and "/".
  - Tests: added `packages/opencode/test/install/checksum-verification.test.ts`; updated `packages/opencode/test/release-validation/windows-installer-930-codex.test.ts` to assert shared `$base` URLs and that verification runs before `Expand-Archive`; Pester coverage for `Test-Checksum` (String/Byte[]/mismatch).

<sup>Written for commit 1ecf6aeca35173574a66d89999d726c4a8aedfcc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/942?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Release artifacts now include a generated SHA-256 `checksums.txt` to verify downloaded archives.
  * Bash installer validates the archive against `checksums.txt` before extraction, and fails fast on mismatches.
  * Windows installer validates the zip against `checksums.txt` before extraction, with improved handling for different checksum file encodings.
  * Installer downloads are now pinned to the same resolved release tag to prevent mixed assets.

* **Tests**
  * Added checks to ensure the release pipeline generates and uploads `checksums.txt`.
  * Added automated coverage for checksum verification in both bash and PowerShell, including mismatch failures and Windows decoding behavior.

### DIFF
--- a/install
+++ b/install
@@ -397,8 +397,13 @@ verify_checksum() {
         print_message error "  actual:   $actual"
         # Clean up via the file's own directory rather than the caller's $tmp_dir,
         # so this stays self-contained and doesn't depend on a dynamically-scoped
-        # local from download_and_install.
-        rm -rf "$(dirname "$file")"
+        # local from download_and_install. Guard against a pathological $file
+        # (empty or root-level) that would make dirname resolve to "." or "/".
+        local cleanup_dir
+        cleanup_dir=$(dirname "$file")
+        if [ -n "$cleanup_dir" ] && [ "$cleanup_dir" != "." ] && [ "$cleanup_dir" != "/" ]; then
+            rm -rf "$cleanup_dir"
+        fi
         exit 1
     fi
     print_message info "${MUTED}Verified ${NC}$name${MUTED} (sha256)${NC}"
--- a/install
+++ b/install
@@ -356,6 +356,51 @@ download_with_progress() {
     return $ret
 }
 
+# Verify the downloaded archive against the release's checksums.txt.
+# Hard-fails on a real mismatch; soft-skips when checksums.txt can't be fetched
+# (older release, network blip) or has no entry, so pinned installs of
+# pre-checksums releases keep working.
+verify_checksum() {
+    local file="$1"
+    local name="$2"
+    # $url ends in /$filename — strip it to get the release base, append checksums.txt.
+    local checksums_url="${url%/*}/checksums.txt"
+
+    local sums
+    if ! sums=$(curl --fail -sL "$checksums_url" 2>/dev/null); then
+        print_message info "${MUTED}Skipping integrity check — checksums.txt not published for this release${NC}"
+        return 0
+    fi
+
+    # checksums.txt is sha256sum format: "<hash>  <filename>" (sha256sum may
+    # prefix the name with '*' in binary mode — tolerate it).
+    local expected
+    expected=$(printf '%s\n' "$sums" | awk -v f="$name" '{ n=$2; sub(/^\*/,"",n); if (n==f) { print $1; exit } }')
+    if [ -z "$expected" ]; then
+        print_message info "${MUTED}Skipping integrity check — no checksum entry for $name${NC}"
+        return 0
+    fi
+
+    local actual
+    if command -v sha256sum >/dev/null 2>&1; then
+        actual=$(sha256sum "$file" | cut -d' ' -f1)
+    elif command -v shasum >/dev/null 2>&1; then
+        actual=$(shasum -a 256 "$file" | cut -d' ' -f1)
+    else
+        print_message info "${MUTED}Skipping integrity check — no sha256 tool available${NC}"
+        return 0
+    fi
+
+    if [ "$actual" != "$expected" ]; then
+        print_message error "Checksum mismatch for $name"
+        print_message error "  expected: $expected"
+        print_message error "  actual:   $actual"
+        rm -rf "$tmp_dir"
+        exit 1
+    fi
+    print_message info "${MUTED}Verified ${NC}$name${MUTED} (sha256)${NC}"
+}
+
 download_and_install() {
     print_message info "\n${MUTED}Installing ${NC}altimate ${MUTED}version: ${NC}$specific_version"
     local tmp_dir="${TMPDIR:-/tmp}/altimate_install_$$"
@@ -366,6 +411,8 @@ download_and_install() {
         # --fail so 404s exit non-zero instead of writing the error page to $tmp_dir/$filename.
         curl --fail -# -L -o "$tmp_dir/$filename" "$url"
     fi
+
+    verify_checksum "$tmp_dir/$filename" "$filename"
 
     # Extract only the expected binary member rather than the whole archive.
     # The current build only puts a single file in each archive, but listing
--- a/install
+++ b/install
@@ -395,7 +395,10 @@ verify_checksum() {
         print_message error "Checksum mismatch for $name"
         print_message error "  expected: $expected"
         print_message error "  actual:   $actual"
-        rm -rf "$tmp_dir"
+        # Clean up via the file's own directory rather than the caller's $tmp_dir,
+        # so this stays self-contained and doesn't depend on a dynamically-scoped
+        # local from download_and_install.
+        rm -rf "$(dirname "$file")"
         exit 1
     fi
     print_message info "${MUTED}Verified ${NC}$name${MUTED} (sha256)${NC}"
--- a/install.ps1
+++ b/install.ps1
@@ -86,7 +86,17 @@ function Test-Checksum {
 
   $sums = $null
   try {
-    $sums = (Invoke-WebRequest -Uri $ChecksumsUrl -UseBasicParsing).Content
+    $resp = Invoke-WebRequest -Uri $ChecksumsUrl -UseBasicParsing
+    # On Windows PowerShell 5.1, .Content is a Byte[] (not a String) whenever the
+    # response isn't a text-recognized content-type — and GitHub serves release
+    # assets as application/octet-stream. A raw Byte[] coerces to a "49 50 51 …"
+    # decimal string when split, so verification would silently soft-skip on the
+    # default Windows shell. Decode the bytes explicitly to recover real text.
+    if ($resp.Content -is [byte[]]) {
+      $sums = [System.Text.Encoding]::UTF8.GetString($resp.Content)
+    } else {
+      $sums = $resp.Content
+    }
   } catch {
     Write-Muted "Skipping integrity check — checksums.txt not published for this release"
     return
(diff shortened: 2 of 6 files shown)


install.ps1
```diff
@@ -77,6 +77,36 @@ public static extern IntPtr SendMessageTimeout(IntPtr hWnd, uint Msg, UIntPtr wP
...
+# Verify a downloaded archive against the release's checksums.txt.
+# Hard-fails (throws) on a real mismatch. Soft-skips when checksums.txt can't be
+# fetched (older release, network blip) or has no entry for this file, so pinned
+# installs of pre-checksums releases keep working.
+function Test-Checksum {
+  param([string]$Path, [string]$Name, [string]$ChecksumsUrl)
+
+  $sums = $null
+  try {
+    $sums = (Invoke-WebRequest -Uri $ChecksumsUrl -UseBasicParsing).Content
```
sahrizvi: **P1 — verification is effectively dead on Windows PowerShell 5.1.**

GitHub release-assets are served with `Content-Type: application/octet-stream` (verified with `curl -I` against an existing release-asset checksums file). On Windows PowerShell 5.1, `Invoke-WebRequest -UseBasicParsing` returns `.Content` as `System.Byte[]` (not `String`) whenever the content-type isn't text-recognized. PS 5.1 is the default shell on Windows 10 and is preinstalled alongside PS 7 on Windows 11.

Downstream effect:
- `$sums = (... ).Content` → byte array, not text
- `$sums -split "`n"` → PowerShell coerces the byte[] to a `"49 50 51 …"` decimal string before splitting → no `\n` boundaries → one element
- `Where-Object { $_ -match … }` → no match
- Falls into the "no checksum entry for X" soft-skip branch every time

So on the dominant Windows shell, every install soft-skips verification and the user sees a misleading "no checksum entry" notice instead of either a real check or a clear "unsupported" error.

Fix — decode bytes explicitly:
```powershell
$resp = Invoke-WebRequest -Uri $ChecksumsUrl -UseBasicParsing
$sums = if ($resp.Content -is [byte[]]) {
    [System.Text.Encoding]::UTF8.GetString($resp.Content)
} else {
    $resp.Content
}
```
Worth a Pester test that feeds a known-good fixture through `Test-Checksum` so this regression can't come back.

sahrizvi (commented):
Local review of #942. One real issue inline (the PS 5.1 byte[] case on `install.ps1:89` — verification is effectively dead on the default Windows shell). Notes from triage:

**False positives I verified before discarding:**
- `tmp_dir` allegedly unbound under `set -u` inside `verify_checksum` (cubic + OCR): bash `local` is dynamically scoped — `tmp_dir` set in `download_and_install` *is* visible to the function it calls. Verified with a 4-line test.
- `\r\n` line endings allegedly corrupting the extracted hash (OCR): the regex's `\s*$` consumes the trailing `\r`, and `-split '\s+'` puts the `\r` in the separator, not in `[0]`. The hash comes out clean.

**Out of scope:** the catch-all-errors behavior in `Test-Checksum`'s try/catch (also flagged by OCR) is a deliberate design choice per the PR description (soft-skip on *any* fetch failure for backwards-compat with pre-checksums releases). Worth a note as a follow-up if this ever moves to GPG-signed releases, but not a defect against this PR's stated contract.

install.ps1
ralphstodomingo: Fixed in \`cecfe069\` — \`Test-Checksum\` now decodes a \`Byte[]\` body explicitly (\`[System.Text.Encoding]::UTF8.GetString\` when \`$resp.Content -is [byte[]]\`), so verification works under Windows PowerShell 5.1 where GitHub's \`application/octet-stream\` assets come back as bytes. Added a Pester test in \`test/windows/install.Tests.ps1\` that runs the real \`Test-Checksum\` against a known-good fixture as **both** String and Byte[], plus a mismatch case — confirmed it fails on the old code and passes on the fix.

While verifying on **real Windows PowerShell 5.1** (5.1.26100) I found the script didn't even *parse* there: it had a few non-ASCII chars (em dash / ellipsis / arrow) and no BOM, so 5.1 read it as ANSI and the tokenizer broke. Fixed that too in this PR (\`d9fe5be7\`, ASCII-only) — it's a pre-existing issue CI missed because the Pester job runs under pwsh 7. End-to-end install (resolve → download → soft-skip checksum → extract → place binary) now succeeds on real PS 5.1.

install
```diff
@@ -356,6 +356,54 @@ download_with_progress() {
...
+
+    if [ "$actual" != "$expected" ]; then
+        print_message error "Checksum mismatch for $name"
+        print_message error "  expected: $expected"
+        print_message error "  actual:   $actual"
+        # Clean up via the file's own directory rather than the caller's $tmp_dir,
+        # so this stays self-contained and doesn't depend on a dynamically-scoped
+        # local from download_and_install.
+        rm -rf "$(dirname "$file")"
+        exit 1
```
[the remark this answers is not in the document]
ralphstodomingo: Done in \`ab14d6bc\` — guarded so we only \`rm -rf\` when \`dirname\` resolves to a real subdirectory, never \`.\` or \`/\`.

ralphstodomingo:
## Combined end-to-end verification (both installer PRs together)

#942 and #946 are sibling PRs that both touch `install` / `install.ps1`. Both were rebased onto `main` (now that #930 has merged) and verified **together** as the combined installer that will ship.

### Merge safety
- Test-merged in **both orders** (#942→#946 and #946→#942) onto current `main`: **zero conflicts** in either order.
- `install.ps1` composes cleanly (the changes touch disjoint regions); `install` (bash) changes are kept in disjoint regions too. The merged result was re-tested after every change.

### Real Windows verification
Ran the **merged** `install.ps1` on a real **Windows PowerShell 5.1** host (5.1.26100, Desktop), the default shell on Windows 10 / clean Windows 11:

- 🐛 **Found & fixed a pre-existing parse failure on PS 5.1.** `install.ps1` had a few non-ASCII characters (em dash / ellipsis / arrow) and no BOM, so PS 5.1 read it as ANSI and the whole script failed to tokenize. CI never caught it because the Pester job runs under `pwsh` (PowerShell 7, UTF-8). Both PRs are now ASCII-only and parse cleanly on 5.1. (The characters predate these PRs.)
- ✅ **Full install E2E on real PS 5.1:** resolve latest (`0.8.7`) → download the real `altimate-windows-x64.zip` (~268 MB) from the pinned release URL → checksum step (soft-skips correctly, since `v0.8.7` predates the new `checksums.txt` asset) → `Expand-Archive` → binary placed (verified as a valid 281 MB PE executable).

### #942 — checksum verification
- The real `Test-Checksum` was exercised on PS 5.1 against `checksums.txt` served **both** as a `String` and as a `Byte[]` (the latter is what Windows PowerShell 5.1 actually returns for GitHub's `application/octet-stream` assets), plus a mismatch case.
- Confirmed the `Byte[]` test **fails on the old code** (silent soft-skip) and **passes with the decode fix** — i.e. it's a real regression guard, now also pinned by a Pester test.
- Archive + `checksums.txt` are pinned to one resolved `$base`, so they can't come from two different releases.

### #946 — resilient latest-version fetch
- Reproduced the `set -euo pipefail` abort: without the trailing `|| true`, a failing `curl --fail` aborts the script at attempt 1 (exit 22); with it, all three retries run and the graceful-degrade path prints and continues.
- Confirmed on real PowerShell that resetting `$specificVersion = $null` makes the "already installed" check fall through (`$null -eq ""` is `$false`) instead of falsely short-circuiting on `"" -eq ""` for a missing/corrupt binary.

### Cross-PR composition (verified on real PowerShell)
With both changes combined: when the version can't be resolved, #946 sets `$specificVersion = $null`, which flows into #942's URL selection and correctly falls back to the mutable `latest/` path; when a version **is** resolved, both the archive and checksums pin to it.

### CI
Green on both: TypeScript, Windows Installer (Pester), and the `#930` release-validation suite (updated in each PR to track the URL-construction and version-fetch behavior changes).

dev-punia-altimate:

- connection_refused
- timeout
- permission_denied
- parse_error
- oom [1.00ms]
- network_error
- auth_failure
- rate_limit
- internal_error
- empty_error
- connection_refused
- timeout
- permission_denied
- parse_error
- network_error [1.00ms]

### Next Step

cc @mdesmet

sahrizvi (approved):
All review concerns adressed and tested E2E, so LGTM.

PR #946, opened 2026-06-15
fix(install): resilient latest-version fetch (both installers)

mdesmet:
## What

Follow-up to #930 (reported by @ralphstodomingo): a transient **504 from `api.github.com/.../releases/latest`** — or the **60/hr/IP unauthenticated rate limit** — aborted the whole install with `Failed to fetch version information`, even though the download uses `releases/latest/download/<file>` (GitHub resolves "latest" server-side, **no API call needed**). The API response only feeds the version-string display and the already-installed short-circuit.

This issue exists in **both** installers (`install` bash:206-213 and `install.ps1`), so both are fixed here for parity.

**Stacked on `feat/windows-powershell-installer`** (sibling to #942) because `install.ps1` isn't on `main` yet.

## Changes (latest path, both installers)

- **Retry** the `releases/latest` API call up to 3× with linear backoff (bash uses `curl --fail` so a 504 retries instead of parsing an error body).
- On continued failure, **degrade gracefully**: print a muted notice and install latest anyway (banner shows `latest`).
- **Only short-circuit as "already installed" on a real version match** — never treat empty==empty (unresolved version + unreadable binary) as installed.
- **Pinned `-Version`/`--version` unchanged**: a genuine 404 still hard-fails.

Note (from the same review comment): the `exit 0`-closes-an-interactive-window quirk under `irm | iex` is PowerShell-only and only affects pasting directly into an interactive shell (not the documented `powershell -c "…"` form); `curl | bash` exits in a subshell. Left as-is — low severity, no parity gap to fix.

## Verification

- `bash -n install` clean; `install.ps1` parses clean and the Pester suite (6/6) still passes on PowerShell 7.6.2.
- `version-fetch-resilience.test.ts` pins retry + graceful-degrade in both installers (57 pass across the install/branding set).
- The degrade path isn't exercised end-to-end (forcing an API failure then proceeds to a 268 MB download); covered by content assertions + parse + no-regression.

---
## Summary by cubic
Makes the “latest” install path resilient in both installers and fixes Windows parsing on PowerShell 5.1. Installs no longer fail when the GitHub releases API blips; they retry and proceed with “latest”.

- **Bug Fixes**
  - Latest path (both): retry `releases/latest` up to 3 times with backoff and a 10s timeout; on failure, continue and show “latest”. Pinned `--version`/`-Version` installs unchanged.
  - Correct “already installed” check: only short-circuit on a real version match; in PowerShell reset `$specificVersion` to `$null` on degrade to avoid empty==empty.
  - Windows: make `install.ps1` ASCII-only so it parses on Windows PowerShell 5.1.
  - Tests: add `packages/opencode/test/install/version-fetch-resilience.test.ts` to pin retry, timeout, and graceful-degrade; update `packages/opencode/test/release-validation/windows-installer-930-codex.test.ts` to expect degrade (not hard-fail) on latest-version fetch.

<sup>Written for commit e10e32293440d0cf32d296bb7798823271a97741. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/946?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Release Notes

* **Bug Fixes**
  * Improved “latest release” resolution in both installers: transient GitHub API failures are retried and no longer stop the installation process.
  * Prevented incorrect “already installed” skipping when version resolution fails, and updated install status to show **latest** when the resolved tag isn’t available.

* **Tests**
  * Updated and added automated coverage to confirm both bash and PowerShell installers degrade gracefully during GitHub API failures and behave correctly afterward.

### DIFF
--- a/install
+++ b/install
@@ -205,11 +205,20 @@ else
 
     if [ -z "$requested_version" ]; then
         url="https://github.com/AltimateAI/altimate-code/releases/latest/download/$filename"
-        specific_version=$(curl -s https://api.github.com/repos/AltimateAI/altimate-code/releases/latest | sed -n 's/.*"tag_name": *"v\([^"]*\)".*/\1/p')
-
-        if [[ $? -ne 0 || -z "$specific_version" ]]; then
-            echo -e "${RED}Failed to fetch version information${NC}"
-            exit 1
+        # The download above resolves "latest" server-side, so this API call only
+        # feeds the version display and the already-installed short-circuit. A
+        # transient api.github.com blip or the unauthenticated rate limit
+        # (60/hr/IP) must NOT abort the install — retry a few times with --fail
+        # (so a 504 retries instead of parsing an error body), then proceed
+        # without the version string.
+        specific_version=""
+        for attempt in 1 2 3; do
+            specific_version=$(curl -fsSL https://api.github.com/repos/AltimateAI/altimate-code/releases/latest 2>/dev/null | sed -n 's/.*"tag_name": *"v\([^"]*\)".*/\1/p')
+            [ -n "$specific_version" ] && break
+            [ "$attempt" -lt 3 ] && sleep "$attempt"
+        done
+        if [ -z "$specific_version" ]; then
+            echo -e "${MUTED}Could not resolve the latest version from GitHub (API unavailable) — installing the latest release anyway.${NC}"
         fi
     else
         # Strip leading 'v' if present
@@ -255,11 +264,14 @@ check_version() {
     if [ -n "$probe" ]; then
         installed_version=$("$probe" --version 2>/dev/null || echo "")
 
-        if [[ "$installed_version" != "$specific_version" ]]; then
-            print_message info "${MUTED}Installed version: ${NC}$installed_version."
-        else
+        # Only short-circuit on a real version match. When the latest version
+        # couldn't be resolved (API unavailable → specific_version empty), never
+        # treat an empty==empty as "already installed" — fall through and reinstall.
+        if [ -n "$specific_version" ] && [[ "$installed_version" == "$specific_version" ]]; then
             print_message info "${MUTED}Version ${NC}$specific_version${MUTED} already installed${NC}"
             exit 0
+        elif [ -n "$installed_version" ]; then
+            print_message info "${MUTED}Installed version: ${NC}$installed_version."
         fi
     fi
 }
@@ -357,7 +369,7 @@ download_with_progress() {
 }
 
 download_and_install() {
-    print_message info "\n${MUTED}Installing ${NC}altimate ${MUTED}version: ${NC}$specific_version"
+    print_message info "\n${MUTED}Installing ${NC}altimate ${MUTED}version: ${NC}${specific_version:-latest}"
     local tmp_dir="${TMPDIR:-/tmp}/altimate_install_$$"
     mkdir -p "$tmp_dir"
 
--- a/install
+++ b/install
@@ -211,9 +211,15 @@ else
         # (60/hr/IP) must NOT abort the install — retry a few times with --fail
         # (so a 504 retries instead of parsing an error body), then proceed
         # without the version string.
+        #
+        # --max-time 10 bounds a dead-air socket (curl's default has no transfer
+        # cap), and the trailing `|| true` is load-bearing: under `set -euo
+        # pipefail`, a failing `curl --fail` propagates through the pipeline and
+        # the assignment, so `set -e` would abort the script before the loop can
+        # retry or degrade. `|| true` lets the failure resolve to an empty string.
         specific_version=""
         for attempt in 1 2 3; do
-            specific_version=$(curl -fsSL https://api.github.com/repos/AltimateAI/altimate-code/releases/latest 2>/dev/null | sed -n 's/.*"tag_name": *"v\([^"]*\)".*/\1/p')
+            specific_version=$(curl -fsSL --max-time 10 https://api.github.com/repos/AltimateAI/altimate-code/releases/latest 2>/dev/null | sed -n 's/.*"tag_name": *"v\([^"]*\)".*/\1/p' || true)
             [ -n "$specific_version" ] && break
             [ "$attempt" -lt 3 ] && sleep "$attempt"
         done
--- a/install.ps1
+++ b/install.ps1
@@ -120,7 +120,10 @@ if ([string]::IsNullOrWhiteSpace($Version)) {
   $specificVersion = ""
   for ($attempt = 1; $attempt -le 3; $attempt++) {
     try {
-      $rel = Invoke-RestMethod -Uri "https://api.github.com/repos/AltimateAI/altimate-code/releases/latest" -Headers @{ "User-Agent" = "altimate-install" }
+      # -TimeoutSec 10 bounds a stuck socket: Invoke-RestMethod defaults to 100s
+      # on PS 5.1 and is effectively unbounded on PS 7+, so without it three
+      # back-to-back retries on dead air could freeze for minutes.
+      $rel = Invoke-RestMethod -Uri "https://api.github.com/repos/AltimateAI/altimate-code/releases/latest" -Headers @{ "User-Agent" = "altimate-install" } -TimeoutSec 10
       $specificVersion = ($rel.tag_name -replace '^v', '')
       if (-not [string]::IsNullOrWhiteSpace($specificVersion)) { break }
     } catch {}
@@ -128,6 +131,13 @@ if ([string]::IsNullOrWhiteSpace($Version)) {
   }
   if ([string]::IsNullOrWhiteSpace($specificVersion)) {
     Write-Muted "Could not resolve the latest version from GitHub (API unavailable) — installing the latest release anyway."
+    # Reset to $null (not ""): the already-installed short-circuit below compares
+    # $installedVersion -eq $specificVersion. If the version probe of a missing or
+    # corrupt binary also yields "", an "" -eq "" match would falsely report
+    # "already installed" and skip the reinstall. $null -eq "" is $false, so the
+    # comparison correctly falls through; the banner still shows "latest" because
+    # if ($specificVersion) treats $null as falsy.
+    $specificVersion = $null
   }
 } else {
   $useLatest = $false
--- a/install.ps1
+++ b/install.ps1
@@ -3,7 +3,7 @@
 #
 # Mirrors ./install (the bash installer for macOS/Linux): it downloads the
 # Bun-compiled standalone executable (altimate.exe) from GitHub releases and
-# drops it in %USERPROFILE%\.altimate\bin — it does NOT depend on npm/Node.
+# drops it in %USERPROFILE%\.altimate\bin - it does NOT depend on npm/Node.
 #
 # Usage:
 #   powershell -c "irm https://www.altimate.sh/install.ps1 | iex"
@@ -64,8 +64,8 @@ if ($Help) {
   exit 0
 }
 
-# A single P/Invoke type carries both native calls we need — the AVX2 CPU probe
-# (kernel32) and the PATH-change broadcast (user32) — so we Add-Type once instead
+# A single P/Invoke type carries both native calls we need - the AVX2 CPU probe
+# (kernel32) and the PATH-change broadcast (user32) - so we Add-Type once instead
 # of compiling a throwaway type per call site.
 function Initialize-Native {
   if (-not ("Win32.AltimateNative" -as [type])) {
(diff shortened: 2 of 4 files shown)


install.ps1
```diff
@@ -112,16 +112,22 @@ function Test-Avx2 {
...
+      $rel = Invoke-RestMethod -Uri "https://api.github.com/repos/AltimateAI/altimate-code/releases/latest" -Headers @{ "User-Agent" = "altimate-install" }
+      $specificVersion = ($rel.tag_name -replace '^v', '')
+      if (-not [string]::IsNullOrWhiteSpace($specificVersion)) { break }
+    } catch {}
+    if ($attempt -lt 3) { Start-Sleep -Seconds $attempt }
   }
   if ([string]::IsNullOrWhiteSpace($specificVersion)) {
-    Write-Err "Failed to fetch version information"
-    exit 1
+    Write-Muted "Could not resolve the latest version from GitHub (API unavailable) — installing the latest release anyway."
```
dev-punia-altimate: **[🟠 MEDIUM]** By changing the API failure behavior from `exit 1` to a fallback mechanism, `$specificVersion` can now safely be an empty string `""`.

However, this introduces a subtle bug further down in the script when it checks if the app is already installed:
```powershell
  try { $installedVersion = (& $probe --version 2>$null | Select-Object -First 1).ToString().Trim() } catch {}
  if ($installedVersion -eq $specificVersion) {
    Write-Muted "Version $specificVersion already installed"
    exit 0
  }
```
If the GitHub API fails (`$specificVersion = ""`) AND the local executable is missing or corrupted such that the `--version` command fails (`$installedVersion` defaults to `""`), the condition `"" -eq ""` will evaluate to `$true`. The installer will output `Version  already installed` and incorrectly short-circuit with `exit 0` instead of downloading and replacing the broken installation.

To avoid this, you can explicitly reset `$specificVersion = $null` here when the API fails, which ensures the downstream string-comparison evaluates to `$false`.

**Suggested change:**
```suggestion
    Write-Muted "Could not resolve the latest version from GitHub (API unavailable) — installing the latest release anyway."
    $specificVersion = $null
```

dev-punia-altimate (commented):

- 0 without a line anchor

#### 1. `install.ps1` (L130)

**[🟠 MEDIUM]** By changing the API failure behavior from `exit 1` to a fallback mechanism, `$specificVersion` can now safely be an empty string `""`.

However, this introduces a subtle bug further down in the script when it checks if the app is already installed:
```powershell
  try { $installedVersion = (& $probe --version 2>$null | Select-Object -First 1).ToString().Trim() } catch {}
  if ($installedVersion -eq $specificVersion) {
    Write-Muted "Version $specificVersion already installed"
    exit 0
  }
```
If the GitHub API fails (`$specificVersion = ""`) AND the local executable is missing or corrupted such that the `--version` command fails (`$installedVersion` defaults to `""`), the condition `"" -eq ""` will evaluate to `$true`. The installer will output `Version  already installed` and incorrectly short-circuit with `exit 0` instead of downloading and replacing the broken installation.

To avoid this, you can explicitly reset `$specificVersion = $null` here when the API fails, which ensures the downstream string-comparison evaluates to `$false`.

**Suggested change:**
```suggestion
    Write-Muted "Could not resolve the latest version from GitHub (API unavailable) — installing the latest release anyway."
    $specificVersion = $null
```

</details>

dev-punia-altimate:

No supported files changed.

install
```diff
@@ -205,11 +205,20 @@ else
...
-            exit 1
+        # The download above resolves "latest" server-side, so this API call only
+        # feeds the version display and the already-installed short-circuit. A
+        # transient api.github.com blip or the unauthenticated rate limit
+        # (60/hr/IP) must NOT abort the install — retry a few times with --fail
+        # (so a 504 retries instead of parsing an error body), then proceed
+        # without the version string.
+        specific_version=""
+        for attempt in 1 2 3; do
+            specific_version=$(curl -fsSL https://api.github.com/repos/AltimateAI/altimate-code/releases/latest 2>/dev/null | sed -n 's/.*"tag_name": *"v\([^"]*\)".*/\1/p')
```
sahrizvi: One subtle interaction to flag here — under the script's `set -euo pipefail` (line 2), the retry loop may not iterate the way it reads.

When `curl --fail` is paired with `pipefail` + `set -e`, a non-zero curl exit (e.g. a real 504 or rate-limit 403) propagates through the pipeline → the command substitution → the assignment, and `set -e` exits the script before `[ -n "$specific_version" ] && break` is evaluated. The loop iterates at most once and the muted "installing latest anyway" notice doesn't get a chance to print.

I ran a quick repro against a local 504-serving stub using these exact lines and the script aborted at attempt 1 with curl's exit 22, never reaching the degrade path.

A one-line fix that lets the inner pipeline failure be absorbed cleanly:

```suggestion
            specific_version=$(curl -fsSL https://api.github.com/repos/AltimateAI/altimate-code/releases/latest 2>/dev/null | sed -n 's/.*"tag_name": *"v\([^"]*\)".*/\1/p' || true)
```

With `|| true` appended, all three attempts iterate and the degrade banner prints on persistent failure.

One related note for a possible follow-up: `packages/opencode/test/install/version-fetch-resilience.test.ts` substring-matches the script source, so `expect(BASH).toContain("for attempt in 1 2 3")` passes regardless of whether the loop body actually runs. A small behavioural test that runs `install` against a stubbed 504 server and asserts the degrade banner appears would catch this kind of regression in CI.

install.ps1
sahrizvi: A subtle interaction worth guarding against on the PS side: if this branch fires (API blip → `$specificVersion = ""`) **and** the installed binary's `--version` probe also fails (corrupt/missing binary → catch absorbs → `$installedVersion = ""`), the downstream check at install.ps1:161 lands on `"" -eq ""` and the installer exits 0 with `"Version  already installed"` instead of reinstalling.

For reference, the equivalent bash check at install:270 has the matching `[ -n "$specific_version" ]` guard — this is just to bring the PS side in line with it. (The Gemini reviewer flagged a similar concern on 2026-06-15 with a producer-side patch; same idea here:)

```suggestion
    Write-Muted "Could not resolve the latest version from GitHub (API unavailable) — installing the latest release anyway."
    $specificVersion = $null
```

`$null -eq ""` evaluates `$false` in PowerShell, so the comparison at line 161 falls through cleanly. The `"Installing $App version: ..."` banner at line 186 still renders "latest" because `if ($specificVersion) { ... }` treats both `""` and `$null` as falsy.

install.ps1
```diff
@@ -112,16 +112,22 @@ function Test-Avx2 {
...
-    exit 1
+  # The download below resolves "latest" server-side (releases/latest/download),
+  # so this API call only feeds the version-string display and the
+  # already-installed short-circuit. A transient api.github.com blip or the
+  # unauthenticated rate limit (60/hr/IP) must NOT abort the install — retry a
+  # few times, then proceed without the version string.
+  $specificVersion = ""
+  for ($attempt = 1; $attempt -le 3; $attempt++) {
+    try {
+      $rel = Invoke-RestMethod -Uri "https://api.github.com/repos/AltimateAI/altimate-code/releases/latest" -Headers @{ "User-Agent" = "altimate-install" }
```
sahrizvi: Worth considering — neither retry sets an explicit HTTP timeout.

`Invoke-RestMethod` defaults to 100s on PS 5.1 and effectively unbounded on PS 7+, so a stuck/dead-air socket can hold each attempt for a long time. Across three back-to-back attempts that can add up to multi-minute apparent freezes during what's normally a quick `irm | iex`.

A short timeout here keeps the worst case bounded:

```suggestion
      $rel = Invoke-RestMethod -Uri "https://api.github.com/repos/AltimateAI/altimate-code/releases/latest" -Headers @{ "User-Agent" = "altimate-install" } -TimeoutSec 10
```

Same gap on the bash side at install:216 if you want to keep parity — `curl -fsSL` has a 60s connect-timeout default and no transfer cap, so `--max-time 10` would be the matching knob there.

sahrizvi (changes requested):
Three findings inline below — one Critical (the retry loop is currently a no-op under `set -euo pipefail`), and two High (install.ps1 short-circuit on empty==empty, and no HTTP timeout on either retry). Repro details and one-line suggested fixes inline.

install
ralphstodomingo: Good catch — fixed in \`d23fc1a5\`. Appended \`|| true\` to the \`curl … | sed …\` assignment so a failing \`curl --fail\` can't abort the script under \`set -euo pipefail\` before the retry loop can iterate/degrade. Reproduced exactly as you described — exit 22 at attempt 1 without it; all 3 attempts run and the degrade banner prints with it. Also added \`--max-time 10\` to bound a dead-air socket (parity with the PS `-TimeoutSec` below). The branch is now rebased onto `main` (since #930 merged).

install.ps1
ralphstodomingo: Fixed in \`d23fc1a5\` — the degrade branch now sets \`$specificVersion = $null\` (not \`""\`), so the already-installed check \`$installedVersion -eq $specificVersion\` can't false-match \`"" -eq ""\` when a missing/corrupt binary's probe also yields \`""\`. \`$null -eq ""\` is \`$false\` so it falls through and reinstalls; the banner still shows \`latest\` because \`if ($specificVersion)\` treats \`$null\` as falsy. Verified the comparison on real PowerShell.

ralphstodomingo: Fixed in \`d23fc1a5\` — set \`$specificVersion = $null\` on the degrade path, bringing the PS side in line with the bash \`[ -n "$specific_version" ]\` guard. \`$null -eq ""\` is \`$false\` so the comparison falls through cleanly and the banner still renders \`latest\`. Verified on real PowerShell.

install.ps1
ralphstodomingo: Done in \`d23fc1a5\` — added \`-TimeoutSec 10\` to \`Invoke-RestMethod\` and \`--max-time 10\` to the bash \`curl\` for parity, bounding the worst-case dead-air freeze across the three retries.

ralphstodomingo:
## Combined end-to-end verification (both installer PRs together)

#942 and #946 are sibling PRs that both touch `install` / `install.ps1`. Both were rebased onto `main` (now that #930 has merged) and verified **together** as the combined installer that will ship.

### Merge safety
- Test-merged in **both orders** (#942→#946 and #946→#942) onto current `main`: **zero conflicts** in either order.
- `install.ps1` composes cleanly (the changes touch disjoint regions); `install` (bash) changes are kept in disjoint regions too. The merged result was re-tested after every change.

### Real Windows verification
Ran the **merged** `install.ps1` on a real **Windows PowerShell 5.1** host (5.1.26100, Desktop), the default shell on Windows 10 / clean Windows 11:

- 🐛 **Found & fixed a pre-existing parse failure on PS 5.1.** `install.ps1` had a few non-ASCII characters (em dash / ellipsis / arrow) and no BOM, so PS 5.1 read it as ANSI and the whole script failed to tokenize. CI never caught it because the Pester job runs under `pwsh` (PowerShell 7, UTF-8). Both PRs are now ASCII-only and parse cleanly on 5.1. (The characters predate these PRs.)
- ✅ **Full install E2E on real PS 5.1:** resolve latest (`0.8.7`) → download the real `altimate-windows-x64.zip` (~268 MB) from the pinned release URL → checksum step (soft-skips correctly, since `v0.8.7` predates the new `checksums.txt` asset) → `Expand-Archive` → binary placed (verified as a valid 281 MB PE executable).

### #942 — checksum verification
- The real `Test-Checksum` was exercised on PS 5.1 against `checksums.txt` served **both** as a `String` and as a `Byte[]` (the latter is what Windows PowerShell 5.1 actually returns for GitHub's `application/octet-stream` assets), plus a mismatch case.
- Confirmed the `Byte[]` test **fails on the old code** (silent soft-skip) and **passes with the decode fix** — i.e. it's a real regression guard, now also pinned by a Pester test.
- Archive + `checksums.txt` are pinned to one resolved `$base`, so they can't come from two different releases.

### #946 — resilient latest-version fetch
- Reproduced the `set -euo pipefail` abort: without the trailing `|| true`, a failing `curl --fail` aborts the script at attempt 1 (exit 22); with it, all three retries run and the graceful-degrade path prints and continues.
- Confirmed on real PowerShell that resetting `$specificVersion = $null` makes the "already installed" check fall through (`$null -eq ""` is `$false`) instead of falsely short-circuiting on `"" -eq ""` for a missing/corrupt binary.

### Cross-PR composition (verified on real PowerShell)
With both changes combined: when the version can't be resolved, #946 sets `$specificVersion = $null`, which flows into #942's URL selection and correctly falls back to the mutable `latest/` path; when a version **is** resolved, both the archive and checksums pin to it.

### CI
Green on both: TypeScript, Windows Installer (Pester), and the `#930` release-validation suite (updated in each PR to track the URL-construction and version-fetch behavior changes).

dev-punia-altimate:

- connection_refused
- timeout
- permission_denied
- parse_error
- network_error [1.00ms]
- auth_failure
- rate_limit
- internal_error
- empty_error
- connection_refused
- timeout
- permission_denied
- parse_error
- oom [1.00ms]
- network_error

### Next Step

cc @mdesmet

sahrizvi (approved):
E2E tested, doesnt't touch core altimate-code, so LGTM.

PR #979, opened 2026-06-30
feat: add `demo-creator` skill

anandgupta42:
## What

Adds **`.claude/skills/demo-creator/`** — a skill that turns a one-line problem statement
into a REAL recorded terminal demo + a blog showing *how* to do it with `altimate-code` and
*why* that's worth it.

## Pipeline

deep-research the real pain → critique → design → record a genuine `altimate-code run` →
extract what the session teaches → visually inspect the frames (and the published page) →
write the blog → critique loop, then publish as a Claude artifact.

## Design notes

- **Baseline is optional.** `showcase` mode (default) shows `altimate-code` alone and lets
  the specific thing it does carry the value; `comparison` mode adds a `claude -p` control
  only when the contrast is the point *and* `altimate-code` actually wins.
- **Two integrity laws:** recordings reconcile with the session trace / `--format json`
  (no faking); value is *shown*, not asserted.
- **Learns from each run:** `extract_learnings.sh` mines the session into `learnings.md`
  (commands run, value-moment evidence, reusable lessons) — feeding the blog and a
  cross-session lesson ledger.
- **Publishes a self-contained artifact:** `build_artifact.sh` inlines clips as base64 and
  renders headless so the page is visually inspected before publishing.
- Reuses the `blog-writer`, `viral-tech-blog`, and `humanizer` skills.

## Validation

Dogfooded end-to-end on a PySpark SQL-port demo (`claude-code` vs `altimate-code`): both
translate correctly, but `altimate-code` runs the reference SQL and `diff`s its output to
*prove* equivalence while `claude-code` only asserts it. Demo artifacts are intentionally
kept out of this PR — this PR is the skill only.

## Contents

`SKILL.md`, `references/` (value-wedge, recording, quality-bar), `scripts/`
(record, inspect, baseline_vs_altimate, setup_pyspark_fixture, extract_learnings,
build_artifact, lib), `templates/` (blog, replicate, vhs tape).

---
## Summary by cubic
Adds `demo-creator`, a skill that turns a one‑line brief into a real recorded terminal demo and a blog showing how to do it with `altimate-code` and why it matters. This update also hardens scripts and docs to enforce fair comparisons and fail‑safe authenticity.

- **New Features**
  - Two modes: `showcase` (default) and optional `comparison` vs `claude-code`; only compare when `altimate-code` actually wins.
  - Real-run recording and inspection: `record.sh` (`asciinema` + `agg`) and `inspect.sh` for frames and trace cross‑check; optional `vhs` fallback.
  - Evidence and publishing: `extract_learnings.sh` → `learnings.md`; `build_artifact.sh` inlines media and headless‑renders before publishing; templates and a PySpark fixture helper.

- **Bug Fixes**
  - Path/input safety and artifacts: `lib.sh` enforces slug‑only topics; `build_artifact.sh` uses `realpath`/`commonpath`, discovers Chrome/Chromium, and hard‑fails `--render` if missing.
  - Fair comparisons: `baseline_vs_altimate.sh` warns on missing `--reset-from`/`--model`, fixes `OPENCODE_PERMISSION`, guards Bash 3.2 array expansion, and fails when the `altimate-code` run fails.
  - Recording/inspection: `record.sh` propagates command exits via `asciinema --return`; `inspect.sh` counts frames robustly and checks `.baseline` clips against the baseline JSON.
  - Fixture setup: `setup_pyspark_fixture.sh` detects existing JDKs (8/11/17/21) across macOS/Linux before Homebrew, validates prerequisites, and smoke‑tests Spark.
  - Docs/templates/miner: `SKILL.md` clarifies same‑run trace requirement; `blog.md.tmpl` tags the baseline section as comparison‑only; `extract_learnings.sh` shows the real source filename.

<sup>Written for commit 13772ba13d70b9d19135a34641531eb61ea6ddf3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/979?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added a guided workflow for creating polished terminal demo content and companion blog posts, including recording, inspection, and packaging into shareable artifacts.
  * Introduced quality-bar and value-angle reference docs to standardize authenticity, evidence, and presentation.
  * Added tools to record runs, extract learnings, inspect recorded clips, and generate baseline-vs-altimate comparisons.
  * Added self-contained HTML artifact generation with embedded assets and optional rendering.
  * Added templates for blogs, demo tapes, and “replicate it yourself” instructions, plus PySpark fixture setup support.

### DIFF
--- a/.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh
+++ b/.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+# Run the SAME task twice and capture each run for comparison:
+#   baseline  = claude-code (the base agent: `claude -p`)         [default]
+#   altimate  = altimate-code (the fork with data tooling)
+# Same prompt, same fixture, same underlying model -> the only difference is altimate-code's
+# additions. This is the honest "why use us over plain Claude Code" control.
+#
+# It captures the json event stream for each run (authenticity proof). It does NOT render
+# GIFs — use record.sh to capture the watchable clip of whichever run(s) you show.
+#
+# Usage:
+#   baseline_vs_altimate.sh <topic> <angle> --cwd DIR --prompt "..." \
+#       [--reset-from PRISTINE_DIR] [--max-turns N] \
+#       [--model PROVIDER/MODEL]      # altimate-code model (e.g. anthropic/claude-...)
+#       [--baseline-model ALIAS]      # claude-code model alias (default: sonnet)
+#       [--baseline-mode claude|disable] [--baseline-deny TOOL,TOOL] [--baseline-agent NAME]
+set -euo pipefail
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"; source "$DIR/lib.sh"
+
+TOPIC=""; ANGLE=""; CWD="$PWD"; PROMPT=""; MODEL=""; BASE_MODEL="sonnet"; MAXT="40"
+RESET_FROM=""; BASE_MODE="claude"; BASE_DENY=""; BASE_AGENT=""
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --cwd) CWD="$2"; shift 2;;
+    --prompt) PROMPT="$2"; shift 2;;
+    --model) MODEL="$2"; shift 2;;
+    --baseline-model) BASE_MODEL="$2"; shift 2;;
+    --reset-from) RESET_FROM="$2"; shift 2;;
+    --max-turns) MAXT="$2"; shift 2;;
+    --baseline-mode) BASE_MODE="$2"; shift 2;;     # claude (default) | disable
+    --baseline-deny) BASE_DENY="$2"; shift 2;;     # only for --baseline-mode disable
+    --baseline-agent) BASE_AGENT="$2"; shift 2;;   # only for --baseline-mode disable
+    -*) echo "unknown opt: $1" >&2; exit 2;;
+    *) if [ -z "$TOPIC" ]; then TOPIC="$1"; elif [ -z "$ANGLE" ]; then ANGLE="$1"; fi; shift;;
+  esac
+done
+[ -n "$TOPIC" ] && [ -n "$ANGLE" ] && [ -n "$PROMPT" ] || {
+  echo "usage: baseline_vs_altimate.sh <topic> <angle> --cwd DIR --prompt '...' [...]" >&2; exit 2; }
+need altimate-code
+
+DD="$(demo_dir "$TOPIC")"; RUNS="$DD/runs"; mkdir -p "$RUNS"
+
+reset_fixture() {
+  [ -n "$RESET_FROM" ] || return 0
+  [ -d "$RESET_FROM" ] || { echo "ERROR: --reset-from '$RESET_FROM' is not a dir" >&2; return 1; }
+  log "  resetting fixture: rsync $RESET_FROM/ -> $CWD/"
+  rsync -a --delete --exclude '.git' "$RESET_FROM"/ "$CWD"/
+}
+
+# ---------------- baseline ----------------
+log "=== BASELINE run ($BASE_MODE) ==="
+reset_fixture
+if [ "$BASE_MODE" = "claude" ]; then
+  need claude
+  ( cd "$CWD" && claude -p --dangerously-skip-permissions --max-turns "$MAXT" \
+      --output-format json --model "$BASE_MODEL" "$PROMPT" ) \
+      | tee "$RUNS/$ANGLE.baseline.json" || true
+else
+  # In-fork isolation: run altimate-code but strip the capability under test.
+  B_ARGS=(); [ -n "$BASE_AGENT" ] && B_ARGS+=(--agent "$BASE_AGENT")
+  M_ARGS=(); [ -n "$MODEL" ] && M_ARGS+=(--model "$MODEL")
+  if [ -n "$BASE_DENY" ]; then
+    export OPENCODE_PERMISSION="{\"tools\":{$(echo "$BASE_DENY" | awk -F, '{for(i=1;i<=NF;i++){printf "%s\"%s\":\"deny\"", (i>1?",":""), $i}}')}}"
+    log "  baseline OPENCODE_PERMISSION=$OPENCODE_PERMISSION"
+  fi
+  ( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
+      "${M_ARGS[@]}" "${B_ARGS[@]}" "$PROMPT" ) | tee "$RUNS/$ANGLE.baseline.json" || true
+  unset OPENCODE_PERMISSION
+fi
+
+# ---------------- altimate ----------------
+log "=== ALTIMATE run ==="
+reset_fixture
+M_ARGS=(); [ -n "$MODEL" ] && M_ARGS+=(--model "$MODEL")
+( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
+    "${M_ARGS[@]}" "$PROMPT" ) | tee "$RUNS/$ANGLE.json" || true
+
+log "json captured: $RUNS/$ANGLE.baseline.json  +  $RUNS/$ANGLE.json"
+log "fair-comparison check: baseline model='$BASE_MODEL', altimate model='${MODEL:-<default>}' — keep these equivalent."
+log "trace ids: 'altimate-code trace list' (altimate run is traced; the claude run leaves its own transcript)."
--- a/.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh
+++ b/.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh
@@ -40,6 +40,19 @@ need altimate-code
 
 DD="$(demo_dir "$TOPIC")"; RUNS="$DD/runs"; mkdir -p "$RUNS"
 
+# The comparison is only fair if BOTH runs start from the identical fixture. Without a
+# pristine reset, the baseline run's edits leak into the altimate run's starting state.
+if [ -z "$RESET_FROM" ]; then
+  log "WARNING: no --reset-from given. The baseline run can mutate $CWD, so the altimate run"
+  log "         would start from contaminated state. Pass --reset-from <pristine> for a valid comparison."
+fi
+# Model control: baseline defaults to '$BASE_MODEL'; if altimate --model is unset it uses the
+# configured default, which may differ. Warn so a win/loss can't come from model selection.
+if [ -z "$MODEL" ]; then
+  log "WARNING: no --model for the altimate run; it will use the configured default. Pass --model"
+  log "         so it matches --baseline-model ('$BASE_MODEL') and the only variable is altimate-code."
+fi
+
 reset_fixture() {
   [ -n "$RESET_FROM" ] || return 0
   [ -d "$RESET_FROM" ] || { echo "ERROR: --reset-from '$RESET_FROM' is not a dir" >&2; return 1; }
@@ -48,6 +61,7 @@ reset_fixture() {
 }
 
 # ---------------- baseline ----------------
+# The baseline is EXPECTED to sometimes struggle, so its failure doesn't fail the script.
 log "=== BASELINE run ($BASE_MODE) ==="
 reset_fixture
 if [ "$BASE_MODE" = "claude" ]; then
@@ -56,24 +70,30 @@ if [ "$BASE_MODE" = "claude" ]; then
       --output-format json --model "$BASE_MODEL" "$PROMPT" ) \
       | tee "$RUNS/$ANGLE.baseline.json" || true
 else
-  # In-fork isolation: run altimate-code but strip the capability under test.
+  # In-fork isolation: run altimate-code but strip the capability under test. Permission keys
+  # are top-level tool names (see config.ts permission schema), e.g. {"bash":"deny"}.
   B_ARGS=(); [ -n "$BASE_AGENT" ] && B_ARGS+=(--agent "$BASE_AGENT")
   M_ARGS=(); [ -n "$MODEL" ] && M_ARGS+=(--model "$MODEL")
   if [ -n "$BASE_DENY" ]; then
-    export OPENCODE_PERMISSION="{\"tools\":{$(echo "$BASE_DENY" | awk -F, '{for(i=1;i<=NF;i++){printf "%s\"%s\":\"deny\"", (i>1?",":""), $i}}')}}"
+    export OPENCODE_PERMISSION="{$(echo "$BASE_DENY" | awk -F, '{for(i=1;i<=NF;i++){printf "%s\"%s\":\"deny\"", (i>1?",":""), $i}}')}"
     log "  baseline OPENCODE_PERMISSION=$OPENCODE_PERMISSION"
   fi
   ( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
-      "${M_ARGS[@]}" "${B_ARGS[@]}" "$PROMPT" ) | tee "$RUNS/$ANGLE.baseline.json" || true
+      ${M_ARGS[@]+"${M_ARGS[@]}"} ${B_ARGS[@]+"${B_ARGS[@]}"} "$PROMPT" ) | tee "$RUNS/$ANGLE.baseline.json" || true
   unset OPENCODE_PERMISSION
 fi
 
 # ---------------- altimate ----------------
+# This run is the evidence — if it doesn't complete, the comparison is invalid, so fail loudly.
 log "=== ALTIMATE run ==="
 reset_fixture
 M_ARGS=(); [ -n "$MODEL" ] && M_ARGS+=(--model "$MODEL")
-( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
-    "${M_ARGS[@]}" "$PROMPT" ) | tee "$RUNS/$ANGLE.json" || true
+set -o pipefail
+if ! ( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
+    ${M_ARGS[@]+"${M_ARGS[@]}"} "$PROMPT" ) | tee "$RUNS/$ANGLE.json"; then
+  echo "ERROR: the altimate-code run failed — comparison evidence is invalid. Not reporting success." >&2
+  exit 1
+fi
 
 log "json captured: $RUNS/$ANGLE.baseline.json  +  $RUNS/$ANGLE.json"
 log "fair-comparison check: baseline model='$BASE_MODEL', altimate model='${MODEL:-<default>}' — keep these equivalent."
--- a/.claude/skills/demo-creator/scripts/build_artifact.sh
+++ b/.claude/skills/demo-creator/scripts/build_artifact.sh
@@ -31,9 +31,12 @@ tpl, base, out = sys.argv[1:4]
 html = open(tpl, encoding="utf-8").read()
 mimetypes.add_type("image/svg+xml", ".svg")
 inlined = [0]; missing = []
+base_real = os.path.realpath(base)
 def datauri(rel):
-    p = os.path.normpath(os.path.join(base, rel))
-    if not (p.startswith(os.path.abspath(base)) and os.path.isfile(p)):
+    # realpath resolves symlinks; commonpath confines to base (a plain startswith prefix
+    # would let a sibling dir sharing base's name — e.g. base/../base-secret — escape).
+    p = os.path.realpath(os.path.join(base, rel))
+    if not (os.path.isfile(p) and os.path.commonpath([p, base_real]) == base_real):
         missing.append(rel); return None
     mime = mimetypes.guess_type(p)[0] or "application/octet-stream"
     inlined[0] += 1
@@ -52,16 +55,27 @@ if missing: print("WARNING: could not inline (will break under CSP): %s" % ", ".
 PY
 
 if [ "$RENDER" = "1" ]; then
-  CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
+  # Discover a Chromium-family browser on PATH or in the usual macOS bundle locations.
+  CHROME=""
+  for c in "${CHROME_BIN:-}" google-chrome google-chrome-stable chromium chromium-browser chrome \
+           "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
+           "/Applications/Chromium.app/Contents/MacOS/Chromium" \
+           "/Applications/Microsoft Edge.app/Contents/MacOS/Microsoft Edge"; do
+    [ -n "$c" ] || continue
+    if command -v "$c" >/dev/null 2>&1; then CHROME="$(command -v "$c")"; break; fi
+    [ -x "$c" ] && { CHROME="$c"; break; }
+  done
   SHOT="$DD/clips/frames/_artifact_render.png"; mkdir -p "$(dirname "$SHOT")"
-  if [ -x "$CHROME" ]; then
-    "$CHROME" --headless=new --disable-gpu --hide-scrollbars --force-device-scale-factor=1 \
-      --window-size="$W,$H" --screenshot="$SHOT" "file://$OUTPATH" 2>/dev/null
-    log "rendered for inspection -> $SHOT  (Read it; check empty grid cells, mostly-empty GIFs, no h-scroll)"
-    echo "$SHOT"
-  else
-    log "Chrome not found; render skipped. Open $OUTPATH manually to inspect before publishing."
+  if [ -z "$CHROME" ]; then
+    # Render-inspection is mandatory before publishing, so a missing browser is a hard error.
+    echo "ERROR: --render requested but no Chrome/Chromium found. Install one or set CHROME_BIN." >&2
+    echo "       (Do NOT publish an artifact you haven't visually inspected.)" >&2
+    exit 1
   fi
+  "$CHROME" --headless=new --disable-gpu --hide-scrollbars --force-device-scale-factor=1 \
+    --window-size="$W,$H" --screenshot="$SHOT" "file://$OUTPATH" 2>/dev/null
+  log "rendered for inspection -> $SHOT  (Read it; check empty grid cells, mostly-empty GIFs, no h-scroll)"
+  echo "$SHOT"
 fi
 log "artifact ready: $OUTPATH  (publish with the Artifact tool; redeploy to the same path on edits)"
 echo "$OUTPATH"
--- a/.claude/skills/demo-creator/scripts/build_artifact.sh
+++ b/.claude/skills/demo-creator/scripts/build_artifact.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+# Build a SELF-CONTAINED HTML artifact from a page template by inlining every relative
+# image src (clips/*.gif, frames/*.png, …) as a base64 data URI. Claude artifacts run under
+# a strict CSP that blocks external hosts, so assets MUST be embedded. Optionally renders the
+# result headless so you can visually inspect it BEFORE publishing (do not skip that).
+#
+# Usage:
+#   build_artifact.sh <topic> <template.html> [--out demo.html] [--render] [--width 820] [--height 5200]
+# Then publish the written file with the Artifact tool (redeploy to the same path/URL on edits).
+set -euo pipefail
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"; source "$DIR/lib.sh"
+
+[ $# -ge 2 ] || { echo "usage: build_artifact.sh <topic> <template.html> [--out NAME] [--render] [--width W] [--height H]" >&2; exit 2; }
+TOPIC="$1"; TPL="$2"; shift 2
+OUT="demo.html"; RENDER=0; W=820; H=5200
+while [ $# -gt 0 ]; do case "$1" in
+  --out) OUT="$2"; shift 2;;
+  --render) RENDER=1; shift;;
+  --width) W="$2"; shift 2;;
+  --height) H="$2"; shift 2;;
+  *) echo "unknown opt: $1" >&2; exit 2;;
+esac; done
+
+[ -f "$TPL" ] || { echo "ERROR: template '$TPL' not found" >&2; exit 1; }
+DD="$(demo_dir "$TOPIC")"; OUTPATH="$DD/$OUT"
+log "building self-contained artifact: $TPL -> $OUTPATH (base dir $DD)"
+
+python3 - "$TPL" "$DD" "$OUTPATH" <<'PY'
+import base64, mimetypes, os, re, sys
+tpl, base, out = sys.argv[1:4]
+html = open(tpl, encoding="utf-8").read()
+mimetypes.add_type("image/svg+xml", ".svg")
+inlined = [0]; missing = []
+def datauri(rel):
+    p = os.path.normpath(os.path.join(base, rel))
+    if not (p.startswith(os.path.abspath(base)) and os.path.isfile(p)):
+        missing.append(rel); return None
+    mime = mimetypes.guess_type(p)[0] or "application/octet-stream"
+    inlined[0] += 1
+    return "data:%s;base64,%s" % (mime, base64.b64encode(open(p,"rb").read()).decode())
+def repl(m):
+    pre, q, url, post = m.group(1), m.group(2), m.group(3), m.group(4)
+    if re.match(r'^(https?:|data:|//)', url): return m.group(0)
+    d = datauri(url)
+    return m.group(0) if d is None else "%s%s%s%s" % (pre, q, d, post)
+# src="..." and CSS url(...) for local image paths
+html = re.sub(r'(src=)(["\'])([^"\']+\.(?:gif|png|jpe?g|svg|webp))(["\'])', repl, html, flags=re.I)
+html = re.sub(r'(url\()(["\']?)([^)"\']+\.(?:gif|png|jpe?g|svg|webp))(["\']?\))', repl, html, flags=re.I)
+open(out, "w", encoding="utf-8").write(html)
+print("inlined %d asset(s); %.2f MB" % (inlined[0], len(html)/1e6))
+if missing: print("WARNING: could not inline (will break under CSP): %s" % ", ".join(missing), file=sys.stderr)
+PY
+
+if [ "$RENDER" = "1" ]; then
+  CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
+  SHOT="$DD/clips/frames/_artifact_render.png"; mkdir -p "$(dirname "$SHOT")"
+  if [ -x "$CHROME" ]; then
+    "$CHROME" --headless=new --disable-gpu --hide-scrollbars --force-device-scale-factor=1 \
+      --window-size="$W,$H" --screenshot="$SHOT" "file://$OUTPATH" 2>/dev/null
+    log "rendered for inspection -> $SHOT  (Read it; check empty grid cells, mostly-empty GIFs, no h-scroll)"
+    echo "$SHOT"
+  else
+    log "Chrome not found; render skipped. Open $OUTPATH manually to inspect before publishing."
+  fi
+fi
+log "artifact ready: $OUTPATH  (publish with the Artifact tool; redeploy to the same path on edits)"
+echo "$OUTPATH"
--- a/.claude/skills/demo-creator/scripts/extract_learnings.sh
+++ b/.claude/skills/demo-creator/scripts/extract_learnings.sh
@@ -1,0 +1,96 @@
+#!/usr/bin/env bash
+# Mine a recorded altimate-code session into learnings.md: what the run actually DID
+# (tool/command sequence), the verification/value markers it hit, and raw material for the
+# "what it did" beat in the blog + the cross-session learning ledger.
+#
+# It only extracts evidence from the real run — it does not invent takeaways. The agent
+# reads learnings.md and writes the one-line "why this matters" + any reusable lesson.
+#
+# Usage: extract_learnings.sh <topic> <angle> [--baseline]   (--baseline mines the claude run too)
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"; source "$DIR/lib.sh"
+# Best-effort miner: grep no-matches and head-closed pipes are expected, so don't let
+# lib.sh's `set -euo pipefail` abort us on them.
+set +e +o pipefail
+
+[ $# -ge 2 ] || { echo "usage: extract_learnings.sh <topic> <angle> [--baseline]" >&2; exit 2; }
+TOPIC="$1"; ANGLE="$2"; WITH_BASE=0
+[ "${3:-}" = "--baseline" ] && WITH_BASE=1
+
+DD="$(demo_dir "$TOPIC")"
+JSON="$DD/runs/$ANGLE.json"
+LOG="$DD/runs/$ANGLE.log"; [ -f "$LOG" ] || LOG="$DD/runs/altimate.log"
+OUT="$DD/learnings.md"
+
+# Pull the source the run actually produced. Prefer json; fall back to the captured log.
+SRC=""
+[ -f "$JSON" ] && SRC="$JSON"
+[ -z "$SRC" ] && [ -f "$LOG" ] && SRC="$LOG"
+[ -n "$SRC" ] || { echo "ERROR: no run json/log for $TOPIC/$ANGLE (runs/$ANGLE.json or runs/altimate.log)" >&2; exit 1; }
+log "mining $SRC -> $OUT"
+
+# Strip ANSI escape codes so the mined evidence is clean (formatted logs are full of them).
+CLEAN="$(mktemp)"; trap 'rm -f "$CLEAN"' EXIT
+perl -pe 's/\x1b\[[0-9;]*[a-zA-Z]//g' "$SRC" > "$CLEAN" 2>/dev/null || cp "$SRC" "$CLEAN"
+SRC="$CLEAN"
+
+# Tool/command sequence the agent executed (bash tool-calls + tool names).
+extract_tools() {  # extract_tools <file>
+  grep -oE '"(tool|name)"[[:space:]]*:[[:space:]]*"[^"]+"' "$1" 2>/dev/null | sed -E 's/.*"([^"]+)"$/\1/' \
+    | grep -viE '^(text|message|assistant|user|step)$' || true
+}
+# Shell commands captured in the recorded output (works on the formatted log too).
+extract_cmds() {  # extract_cmds <file>
+  grep -oE '(\$ |bash |python |diff |altimate-code |duckdb |run_sql)[^"]*' "$1" 2>/dev/null \
+    | sed -E 's/\\u[0-9a-fA-F]{4}//g' | grep -vE '^\s*$' | head -40 || true
+}
+# Verification / value-moment markers — evidence the valuable thing happened on camera.
+VALUE_MARKERS='run_sql|diff |byte-for-byte|identical|zero diff|verify|expected_sql|equival|sql_analyze|altimate_core|\bcheck\b|lineage|cost|row count|assert'
+
+TOOLS="$(extract_tools "$SRC" | sort | uniq -c | sort -rn | head -25)"
+CMDS="$(extract_cmds "$SRC")"
+VALUE_HITS="$(grep -niE "$VALUE_MARKERS" "$SRC" 2>/dev/null | sed -E 's/\\u[0-9a-fA-F]{4}//g' | head -25)"
+TRACE_ID="$(grep -oE 'ses_[A-Za-z0-9]+' "$SRC" 2>/dev/null | head -1)"
+
+{
+  echo "# Learnings — $TOPIC / $ANGLE"
+  echo
+  echo "_Mined from \`$(basename "$SRC")\`$( [ -n "$TRACE_ID" ] && echo " · trace \`$TRACE_ID\`")._"
+  echo "_All items below are extracted from the REAL run. The 'why it matters' and reusable"
+  echo "lessons are filled in by the author after reading this — do not invent value the run"
+  echo "did not show (law #2)._"
+  echo
+  echo "## What altimate-code actually ran"
+  echo '```'
+  [ -n "$CMDS" ] && echo "$CMDS" || echo "(no shell commands parsed — inspect $SRC directly)"
+  echo '```'
+  echo
+  echo "## Tools / events observed (count)"
+  echo '```'
+  [ -n "$TOOLS" ] && echo "$TOOLS" || echo "(none parsed)"
+  echo '```'
+  echo
+  echo "## Value-moment evidence (verification / capability markers)"
+  if [ -n "$VALUE_HITS" ]; then echo '```'; echo "$VALUE_HITS"; echo '```'; else
+    echo "**NONE FOUND.** The run shows no verification/capability marker — per law #2 you"
+    echo "do not yet have a value moment for this angle. Change the task or report honestly."
+  fi
+  echo
+  echo "## Why it matters  <!-- author fills in: one line, grounded in the evidence above -->"
+  echo
+  echo "## Reusable lesson  <!-- author: product bug / env gotcha / prompt that triggers the value / fixture pattern."
+  echo "                         If it should persist across sessions, also save to memory + MANIFEST ledger. -->"
+} > "$OUT"
+
+if [ "$WITH_BASE" = "1" ] && [ -f "$DD/runs/$ANGLE.baseline.json" ]; then
+  {
+    echo
+    echo "## Baseline (claude-code) — value markers, for the contrast"
+    BH="$(grep -niE "$VALUE_MARKERS" "$DD/runs/$ANGLE.baseline.json" 2>/dev/null | head -20)"
+    if [ -n "$BH" ]; then echo '```'; echo "$BH"; echo '```'; else
+      echo "(baseline shows none of: $VALUE_MARKERS — this absence IS the contrast.)"
+    fi
+  } >> "$OUT"
+fi
+
+log "wrote $OUT"
+echo "$OUT"
--- a/.claude/skills/demo-creator/scripts/extract_learnings.sh
+++ b/.claude/skills/demo-creator/scripts/extract_learnings.sh
@@ -22,13 +22,18 @@ LOG="$DD/runs/$ANGLE.log"; [ -f "$LOG" ] || LOG="$DD/runs/altimate.log"
 OUT="$DD/learnings.md"
 
 # Pull the source the run actually produced. Prefer json; fall back to the captured log.
+CAST="$DD/clips/$ANGLE.cast"
 SRC=""
 [ -f "$JSON" ] && SRC="$JSON"
 [ -z "$SRC" ] && [ -f "$LOG" ] && SRC="$LOG"
-[ -n "$SRC" ] || { echo "ERROR: no run json/log for $TOPIC/$ANGLE (runs/$ANGLE.json or runs/altimate.log)" >&2; exit 1; }
+# Fall back to the recorded asciinema cast — it IS the run record (JSONL; output text is greppable).
+[ -z "$SRC" ] && [ -f "$CAST" ] && SRC="$CAST"
+[ -n "$SRC" ] || { echo "ERROR: no run source for $TOPIC/$ANGLE (runs/$ANGLE.json, runs/altimate.log, or clips/$ANGLE.cast)" >&2; exit 1; }
 log "mining $SRC -> $OUT"
 
-# Strip ANSI escape codes so the mined evidence is clean (formatted logs are full of them).
+# Remember the real source for display; mine from an ANSI-stripped copy (formatted logs are
+# full of escape codes). SRC below points at the temp copy, SRC_NAME at the real file.
+SRC_NAME="$(basename "$SRC")"
 CLEAN="$(mktemp)"; trap 'rm -f "$CLEAN"' EXIT
 perl -pe 's/\x1b\[[0-9;]*[a-zA-Z]//g' "$SRC" > "$CLEAN" 2>/dev/null || cp "$SRC" "$CLEAN"
 SRC="$CLEAN"
@@ -54,7 +59,7 @@ TRACE_ID="$(grep -oE 'ses_[A-Za-z0-9]+' "$SRC" 2>/dev/null | head -1)"
 {
   echo "# Learnings — $TOPIC / $ANGLE"
   echo
-  echo "_Mined from \`$(basename "$SRC")\`$( [ -n "$TRACE_ID" ] && echo " · trace \`$TRACE_ID\`")._"
+  echo "_Mined from \`$SRC_NAME\`$( [ -n "$TRACE_ID" ] && echo " · trace \`$TRACE_ID\`")._"
   echo "_All items below are extracted from the REAL run. The 'why it matters' and reusable"
   echo "lessons are filled in by the author after reading this — do not invent value the run"
   echo "did not show (law #2)._"
--- a/.claude/skills/demo-creator/scripts/inspect.sh
+++ b/.claude/skills/demo-creator/scripts/inspect.sh
@@ -25,11 +25,17 @@ mkdir -p "$FRAMES"; rm -f "$FRAMES"/f_*.png
 
 log "extracting frames @ ${FPS}fps -> $FRAMES"
 ffmpeg -y -loglevel error -i "$GIF" -vf "fps=$FPS" "$FRAMES/f_%03d.png"
-COUNT=$(ls "$FRAMES"/f_*.png 2>/dev/null | wc -l | tr -d ' ')
+# Count without letting a no-match ls abort us under `set -e`/pipefail (lib.sh sets both).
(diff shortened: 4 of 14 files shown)


dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 18 finding(s)

- 18 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `.claude/skills/demo-creator/scripts/inspect.sh` (L11-L17)

**[🟠 MEDIUM]** Because `set -euo pipefail` (specifically `set -u`) is enabled, if the script is invoked with `--fps` or `--prefix` as the very last argument without providing a corresponding value, evaluating `$2` will throw an unbound variable error and crash the script ungracefully (e.g., `./inspect.sh: line 13: $2: unbound variable`).

You can add a check to explicitly ensure that the argument value is present, making the script more robust.

**Suggested change:**
```suggestion
while [ $# -gt 0 ]; do
  case "$1" in
    --fps)
      [ $# -ge 2 ] || { echo "--fps requires a value" >&2; exit 2; }
      FPS="$2"; shift 2;;
    --prefix)
      [ $# -ge 2 ] || { echo "--prefix requires a value" >&2; exit 2; }
      PREFIX="$2"; shift 2;;
    *) echo "unknown opt: $1" >&2; exit 2;;
  esac
done
```

#### 2. `.claude/skills/demo-creator/scripts/inspect.sh` (L28)

**[🟠 MEDIUM]** Due to `set -euo pipefail`, if `ls` fails to find any matching files (e.g., if `ffmpeg` produces no output due to an empty or corrupted GIF), it will exit with a non-zero status. This will cause the entire pipeline to fail and the script will silently abort at this line without printing any error message since `stderr` is redirected to `/dev/null`.

Additionally, using shell glob expansion with `ls` can result in an `Argument list too long` error if a massive number of frames is extracted.

Using `find` avoids pipeline failures when no files exist (since `find` safely returns 0) and protects against the argument list limit.

**Suggested change:**
```suggestion
COUNT=$(find "$FRAMES" -maxdepth 1 -name "f_*.png" 2>/dev/null | wc -l | tr -d ' ')
```

#### 3. `.claude/skills/demo-creator/scripts/extract_learnings.sh` (L33-L34)

**[🔵 LOW]** Since `SRC` is reassigned to the temporary file `$CLEAN`, `$(basename "$SRC")` later in the script will output the temporary filename (e.g., `tmp.xxxxxx`) instead of the original log or JSON filename.

Consider storing the original filename before reassignment to preserve traceability.

**Suggested change:**
```suggestion
perl -pe 's/\x1b\[[0-9;]*[a-zA-Z]//g' "$SRC" > "$CLEAN" 2>/dev/null || cp "$SRC" "$CLEAN"
ORIG_SRC="$SRC"
SRC="$CLEAN"
```

#### 4. `.claude/skills/demo-creator/scripts/extract_learnings.sh` (L55-L57)

**[🔵 LOW]** Use the original source file name here to accurately report where the learnings were mined from.

**Suggested change:**
```suggestion
  echo "# Learnings — $TOPIC / $ANGLE"
  echo
  echo "_Mined from \`$(basename "$ORIG_SRC")\`$( [ -n "$TRACE_ID" ] && echo " · trace \`$TRACE_ID\`")._"
```

#### 5. `.claude/skills/demo-creator/scripts/extract_learnings.sh` (L42-L45)

**[🟠 MEDIUM]** The regex `[^"]*` will truncate the extracted command at the first double quote. While this correctly stops at the end of a JSON string property, it will prematurely truncate shell commands that themselves contain double quotes (e.g., `python -c "print('test')"` or escaped quotes like `\"`). 

Since this is explicitly a "best-effort miner", this might be an acceptable limitation, but if extracting complete command arguments is important, consider a more robust matching strategy or using a JSON parser like `jq` when `SRC` is a JSON file.

#### 6. `.claude/skills/demo-creator/scripts/record.sh` (L49)

**[🟠 MEDIUM]** The use of `bash -lc` initiates a login shell, which will source user profile scripts (e.g., `~/.bash_profile`, `~/.profile`). This could unexpectedly output text (like a Message of the Day, or initialization logs) that gets captured in the asciinema recording, potentially corrupting the demonstration. 

Unless login shell behaviors are strictly required (like loading custom `$PATH` from profile scripts), it is safer to use a non-login shell with `bash -c`.

**Suggested change:**
```suggestion
    --command "bash -c $(printf '%q' "$INNER")" "$CAST" )
```

#### 7. `.claude/skills/demo-creator/scripts/record.sh` (L14-L22)

**[🟠 MEDIUM]** There is no boundary check for the option arguments in this `while` loop. Because the script uses `set -u`, providing a flag (e.g., `--cwd` or `--cols`) as the very last argument without its value will trigger a confusing `unbound variable` error on `$2` and abruptly exit. 

To improve the user experience and provide clearer error messages, consider verifying that `$2` exists before assignment, or using `${2:-}`. For example:

**Suggested change:**
```suggestion
while [ $# -gt 0 ]; do
  case "$1" in
    --cwd) CWD="${2?Missing argument for --cwd}"; shift 2;;
    --cols) COLS="${2?Missing argument for --cols}"; shift 2;;
    --rows) ROWS="${2?Missing argument for --rows}"; shift 2;;
    --font) FONT="${2?Missing argument for --font}"; shift 2;;
    --idle) IDLE="${2?Missing argument for --idle}"; shift 2;;
    --out-prefix) PREFIX="${2?Missing argument for --out-prefix}"; shift 2;;
    --banner) BANNER="${2?Missing argument for --banner}"; shift 2;;   # shown on screen before the run (headless hides the invocation)
```

#### 8. `.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh` (L16-L21)

**[🔴 HIGH]** **Platform dependence and rigid `JAVA_HOME` handling.** The script hardcodes Apple Silicon macOS paths (`/opt/homebrew/...`) for JDK discovery and strictly relies on `brew` to install openjdk if a suitable version is not found. This will cause failures on Linux, Windows, CI containers, and Intel Macs (`/usr/local/opt/...`). Additionally, it completely ignores any existing, valid `JAVA_HOME` environment variable provided by the user.

Consider including the user's `$JAVA_HOME` in the candidate list and generalizing the discovery logic.

**Suggested change:**
```suggestion
ENV_JAVA_HOME="${JAVA_HOME:-}"
JAVA_HOME=""
for cand in \
  "$ENV_JAVA_HOME" \
  "/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home" \
  "/opt/homebrew/opt/openjdk@11/libexec/openjdk.jdk/Contents/Home" \
  "/opt/homebrew/opt/openjdk@21/libexec/openjdk.jdk/Contents/Home"; do
  [ -n "$cand" ] || continue
  [ -x "$cand/bin/java" ] || continue
```

#### 9. `.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh` (L15)

**[🟠 MEDIUM]** **Incorrect Java major version parsing for Java 8.** The `java_major` function uses a regex that incorrectly extracts `1` instead of `8` for Java 8 (e.g., `java version "1.8.0_xxx"`). As a result, the subsequent `case` statement will erroneously reject valid Java 8 installations and force an unnecessary JDK download.

You can adjust the regex to account for the `1.x` pattern used by older Java versions.

**Suggested change:**
```suggestion
java_major() { "$1/bin/java" -version 2>&1 | sed -nE 's/.*version "(1\.)?([0-9]+).*/\2/p' | head -1; }
```

#### 10. `.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh` (L39-L41)

**[🟠 MEDIUM]** **Missing prerequisite validation in the fallback path.** When the `uv` tool is unavailable, the script silently falls back to executing `python3 -m venv` without verifying if the `python3` command actually exists on the host. This will result in a messy 'command not found' failure mid-execution.

You can use the standard `need` helper from `lib.sh` to validate `python3` presence.

**Suggested change:**
```suggestion
else
  need python3
  python3 -m venv "$VENV"; "$VENV/bin/pip" install -q --upgrade pip "$PYSPARK_SPEC"
fi
```

#### 11. `.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh` (L66-L67)

**[🔴 HIGH]** Expanding empty arrays (`"${M_ARGS[@]}"` and `"${B_ARGS[@]}"`) with `set -u` enabled will trigger an "unbound variable" error and crash the script on Bash versions older than 4.4 (which includes the default Bash 3.2 on macOS).

Consider using fallback syntax `"${M_ARGS[@]+"${M_ARGS[@]}"}"` or checking for length before expansion to ensure cross-compatibility.

**Suggested change:**
```suggestion
  ( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
      ${M_ARGS[@]+"${M_ARGS[@]}"} ${B_ARGS[@]+"${B_ARGS[@]}"} "$PROMPT" ) | tee "$RUNS/$ANGLE.baseline.json" || true
```

#### 12. `.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh` (L62-L68)

**[🟠 MEDIUM]** The `OPENCODE_PERMISSION` variable is conditionally exported in the main script environment and then unconditionally cleared via `unset OPENCODE_PERMISSION`. 

If a user had this variable configured in their environment prior to running the script, it will be unintentionally wiped out for the remainder of the script (e.g., during the subsequent `ALTIMATE run`). 

To avoid side effects, consider moving the conditional `export` logic inside the `( ... )` subshell so it naturally scopes to the baseline run without affecting the parent environment.

**Suggested change:**
```suggestion
  (
    if [ -n "$BASE_DENY" ]; then
      export OPENCODE_PERMISSION="{\"tools\":{$(echo "$BASE_DENY" | awk -F, '{for(i=1;i<=NF;i++){printf "%s\"%s\":\"deny\"", (i>1?",":""), $i}}')}}"
      log "  baseline OPENCODE_PERMISSION=$OPENCODE_PERMISSION"
    fi
    cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
      ${M_ARGS[@]+"${M_ARGS[@]}"} ${B_ARGS[@]+"${B_ARGS[@]}"} "$PROMPT" 
  ) | tee "$RUNS/$ANGLE.baseline.json" || true
```

#### 13. `.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh` (L23-L33)

**[🟠 MEDIUM]** Argument parsing directly accesses `$2` and shifts by 2 for flags that expect values. 

Since `set -u` is enabled, if any of these flags are passed as the very last argument without a following value, the script will crash abruptly with a Bash "unbound variable" error rather than printing a helpful validation message. 

Consider checking `$# -ge 2` before processing these flags, or using standard fallback syntax like `${2:-}`.

**Suggested change:**
```suggestion
  case "$1" in
    --cwd) [ $# -ge 2 ] || { echo "ERROR: --cwd requires a value" >&2; exit 2; }; CWD="$2"; shift 2;;
    --prompt) [ $# -ge 2 ] || { echo "ERROR: --prompt requires a value" >&2; exit 2; }; PROMPT="$2"; shift 2;;
    --model) [ $# -ge 2 ] || { echo "ERROR: --model requires a value" >&2; exit 2; }; MODEL="$2"; shift 2;;
    --baseline-model) [ $# -ge 2 ] || { echo "ERROR: --baseline-model requires a value" >&2; exit 2; }; BASE_MODEL="$2"; shift 2;;
    --reset-from) [ $# -ge 2 ] || { echo "ERROR: --reset-from requires a value" >&2; exit 2; }; RESET_FROM="$2"; shift 2;;
    --max-turns) [ $# -ge 2 ] || { echo "ERROR: --max-turns requires a value" >&2; exit 2; }; MAXT="$2"; shift 2;;
    --baseline-mode) [ $# -ge 2 ] || { echo "ERROR: --baseline-mode requires a value" >&2; exit 2; }; BASE_MODE="$2"; shift 2;;
    --baseline-deny) [ $# -ge 2 ] || { echo "ERROR: --baseline-deny requires a value" >&2; exit 2; }; BASE_DENY="$2"; shift 2;;
    --baseline-agent) [ $# -ge 2 ] || { echo "ERROR: --baseline-agent requires a value" >&2; exit 2; }; BASE_AGENT="$2"; shift 2;;
    -*) echo "unknown opt: $1" >&2; exit 2;;
```

#### 14. `.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh` (L75-L76)

**[🔴 HIGH]** This array expansion will also trigger an "unbound variable" error under Bash 3.2 when the array is empty. Consider applying the same fallback syntax here.

**Suggested change:**
```suggestion
( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
    ${M_ARGS[@]+"${M_ARGS[@]}"} "$PROMPT" ) | tee "$RUNS/$ANGLE.json" || true
```

#### 15. `.claude/skills/demo-creator/scripts/build_artifact.sh` (L16-L22)

**[🟠 MEDIUM]** Since `set -u` is enabled at the top of the script, referencing `$2` without checking if it exists will throw an abrupt bash "unbound variable" error if an option like `--out` is passed as the very last argument. Additionally, `shift 2` will fail if there are less than 2 arguments left. It's safer to explicitly check `$#` to provide a clear error message.

**Suggested change:**
```suggestion
while [ $# -gt 0 ]; do case "$1" in
  --out) [ $# -ge 2 ] || { echo "missing value for $1" >&2; exit 2; }; OUT="$2"; shift 2;;
  --render) RENDER=1; shift;;
  --width) [ $# -ge 2 ] || { echo "missing value for $1" >&2; exit 2; }; W="$2"; shift 2;;
  --height) [ $# -ge 2 ] || { echo "missing value for $1" >&2; exit 2; }; H="$2"; shift 2;;
  *) echo "unknown opt: $1" >&2; exit 2;;
esac; done
```

#### 16. `.claude/skills/demo-creator/scripts/build_artifact.sh` (L34-L36)

**[🟠 MEDIUM]** `os.path.normpath` does not resolve relative paths to absolute ones. If `base` happens to be a relative path (e.g., if invoked dynamically differently than current shell expectation), `p` will also be relative. Comparing a relative `p` against the absolute path from `os.path.abspath(base)` via `startswith()` will always evaluate to `False`, failing to inline valid images. Consider using `os.path.abspath` to correctly resolve `p`.

**Suggested change:**
```suggestion
def datauri(rel):
    p = os.path.abspath(os.path.join(base, rel))
    if not (p.startswith(os.path.abspath(base)) and os.path.isfile(p)):
```

#### 17. `.claude/skills/demo-creator/scripts/build_artifact.sh` (L51)

**[🟠 MEDIUM]** The bash script runs with `set -e`, but the embedded Python script exits with a `0` status code even when local assets cannot be found. Continuing the build process with missing assets will silently produce artifacts that break under strict CSP. Exiting with a non-zero status code here would correctly halt the build process.

**Suggested change:**
```suggestion
if missing:
    print("ERROR: could not inline (will break under CSP): %s" % ", ".join(missing), file=sys.stderr)
    sys.exit(1)
```

#### 18. `.claude/skills/demo-creator/scripts/build_artifact.sh` (L55-L56)

**[🔴 HIGH]** The Chrome executable path is hardcoded to the macOS-specific application path. If this script is executed in Linux or Windows environments, the `--render` step will silently fail to find Chrome and skip rendering. It's recommended to add a cross-platform fallback to look up valid executables (e.g., `google-chrome` or `chromium`) in `$PATH`.

**Suggested change:**
```suggestion
  CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
  if [ ! -x "$CHROME" ]; then
    CHROME="$(command -v google-chrome || command -v chromium || echo "$CHROME")"
  fi
  SHOT="$DD/clips/frames/_artifact_render.png"; mkdir -p "$(dirname "$SHOT")"
```

</details>

.claude/skills/demo-creator/scripts/inspect.sh
```diff
@@ -0,0 +1,43 @@
...
+[ $# -ge 2 ] || { echo "usage: inspect.sh <topic> <angle> [--fps N] [--prefix NAME]" >&2; exit 2; }
+TOPIC="$1"; ANGLE="$2"; shift 2
+FPS=1; PREFIX="$ANGLE"
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --fps) FPS="$2"; shift 2;;
+    --prefix) PREFIX="$2"; shift 2;;
+    *) echo "unknown opt: $1" >&2; exit 2;;
+  esac
+done
```
dev-punia-altimate, line 17: **[🟠 MEDIUM]** Because `set -euo pipefail` (specifically `set -u`) is enabled, if the script is invoked with `--fps` or `--prefix` as the very last argument without providing a corresponding value, evaluating `$2` will throw an unbound variable error and crash the script ungracefully (e.g., `./inspect.sh: line 13: $2: unbound variable`).

You can add a check to explicitly ensure that the argument value is present, making the script more robust.

**Suggested change:**
```suggestion
while [ $# -gt 0 ]; do
  case "$1" in
    --fps)
      [ $# -ge 2 ] || { echo "--fps requires a value" >&2; exit 2; }
      FPS="$2"; shift 2;;
    --prefix)
      [ $# -ge 2 ] || { echo "--prefix requires a value" >&2; exit 2; }
      PREFIX="$2"; shift 2;;
    *) echo "unknown opt: $1" >&2; exit 2;;
  esac
done
```

.claude/skills/demo-creator/scripts/inspect.sh
```diff
@@ -0,0 +1,43 @@
...
+
+DD="$(demo_dir "$TOPIC")"
+GIF="$DD/clips/$PREFIX.gif"
+FRAMES="$DD/clips/frames/$PREFIX"
+[ -f "$GIF" ] || { echo "ERROR: no GIF at $GIF (record it first)" >&2; exit 1; }
+mkdir -p "$FRAMES"; rm -f "$FRAMES"/f_*.png
+
+log "extracting frames @ ${FPS}fps -> $FRAMES"
+ffmpeg -y -loglevel error -i "$GIF" -vf "fps=$FPS" "$FRAMES/f_%03d.png"
+COUNT=$(ls "$FRAMES"/f_*.png 2>/dev/null | wc -l | tr -d ' ')
```
dev-punia-altimate: **[🟠 MEDIUM]** Due to `set -euo pipefail`, if `ls` fails to find any matching files (e.g., if `ffmpeg` produces no output due to an empty or corrupted GIF), it will exit with a non-zero status. This will cause the entire pipeline to fail and the script will silently abort at this line without printing any error message since `stderr` is redirected to `/dev/null`.

Additionally, using shell glob expansion with `ls` can result in an `Argument list too long` error if a massive number of frames is extracted.

Using `find` avoids pipeline failures when no files exist (since `find` safely returns 0) and protects against the argument list limit.

**Suggested change:**
```suggestion
COUNT=$(find "$FRAMES" -maxdepth 1 -name "f_*.png" 2>/dev/null | wc -l | tr -d ' ')
```

.claude/skills/demo-creator/scripts/extract_learnings.sh
```diff
@@ -0,0 +1,96 @@
...
+SRC=""
+[ -f "$JSON" ] && SRC="$JSON"
+[ -z "$SRC" ] && [ -f "$LOG" ] && SRC="$LOG"
+[ -n "$SRC" ] || { echo "ERROR: no run json/log for $TOPIC/$ANGLE (runs/$ANGLE.json or runs/altimate.log)" >&2; exit 1; }
+log "mining $SRC -> $OUT"
+
+# Strip ANSI escape codes so the mined evidence is clean (formatted logs are full of them).
+CLEAN="$(mktemp)"; trap 'rm -f "$CLEAN"' EXIT
+perl -pe 's/\x1b\[[0-9;]*[a-zA-Z]//g' "$SRC" > "$CLEAN" 2>/dev/null || cp "$SRC" "$CLEAN"
+SRC="$CLEAN"
```
dev-punia-altimate, line 39: **[🔵 LOW]** Since `SRC` is reassigned to the temporary file `$CLEAN`, `$(basename "$SRC")` later in the script will output the temporary filename (e.g., `tmp.xxxxxx`) instead of the original log or JSON filename.

Consider storing the original filename before reassignment to preserve traceability.

**Suggested change:**
```suggestion
perl -pe 's/\x1b\[[0-9;]*[a-zA-Z]//g' "$SRC" > "$CLEAN" 2>/dev/null || cp "$SRC" "$CLEAN"
ORIG_SRC="$SRC"
SRC="$CLEAN"
```

.claude/skills/demo-creator/scripts/extract_learnings.sh
```diff
@@ -0,0 +1,96 @@
...
+
+TOOLS="$(extract_tools "$SRC" | sort | uniq -c | sort -rn | head -25)"
+CMDS="$(extract_cmds "$SRC")"
+VALUE_HITS="$(grep -niE "$VALUE_MARKERS" "$SRC" 2>/dev/null | sed -E 's/\\u[0-9a-fA-F]{4}//g' | head -25)"
+TRACE_ID="$(grep -oE 'ses_[A-Za-z0-9]+' "$SRC" 2>/dev/null | head -1)"
+
+{
+  echo "# Learnings — $TOPIC / $ANGLE"
+  echo
+  echo "_Mined from \`$(basename "$SRC")\`$( [ -n "$TRACE_ID" ] && echo " · trace \`$TRACE_ID\`")._"
```
dev-punia-altimate: **[🔵 LOW]** Use the original source file name here to accurately report where the learnings were mined from.

**Suggested change:**
```suggestion
  echo "# Learnings — $TOPIC / $ANGLE"
  echo
  echo "_Mined from \`$(basename "$ORIG_SRC")\`$( [ -n "$TRACE_ID" ] && echo " · trace \`$TRACE_ID\`")._"
```

.claude/skills/demo-creator/scripts/extract_learnings.sh
```diff
@@ -0,0 +1,96 @@
...
+# Tool/command sequence the agent executed (bash tool-calls + tool names).
+extract_tools() {  # extract_tools <file>
+  grep -oE '"(tool|name)"[[:space:]]*:[[:space:]]*"[^"]+"' "$1" 2>/dev/null | sed -E 's/.*"([^"]+)"$/\1/' \
+    | grep -viE '^(text|message|assistant|user|step)$' || true
+}
+# Shell commands captured in the recorded output (works on the formatted log too).
+extract_cmds() {  # extract_cmds <file>
+  grep -oE '(\$ |bash |python |diff |altimate-code |duckdb |run_sql)[^"]*' "$1" 2>/dev/null \
+    | sed -E 's/\\u[0-9a-fA-F]{4}//g' | grep -vE '^\s*$' | head -40 || true
+}
```
dev-punia-altimate, line 50: **[🟠 MEDIUM]** The regex `[^"]*` will truncate the extracted command at the first double quote. While this correctly stops at the end of a JSON string property, it will prematurely truncate shell commands that themselves contain double quotes (e.g., `python -c "print('test')"` or escaped quotes like `\"`). 

Since this is explicitly a "best-effort miner", this might be an acceptable limitation, but if extracting complete command arguments is important, consider a more robust matching strategy or using a JSON parser like `jq` when `SRC` is a JSON file.

.claude/skills/demo-creator/scripts/record.sh
```diff
@@ -0,0 +1,56 @@
...
+# Optional on-screen banner so viewers see what was asked (headless mode hides the prompt).
+# The real command is unchanged — the banner only echoes context, it does not fake output.
+INNER="$CMD"
+if [ -n "$BANNER" ]; then
+  INNER="printf '\033[1;36m\$ %s\033[0m\n\n' $(printf '%q' "$BANNER"); sleep 1; $CMD"
+fi
+
+# Record. asciinema (v3) runs the command in a real pty and stops when it exits.
+( cd "$CWD" && asciinema rec --overwrite --window-size "${COLS}x${ROWS}" \
+    --command "bash -lc $(printf '%q' "$INNER")" "$CAST" )
```
dev-punia-altimate: **[🟠 MEDIUM]** The use of `bash -lc` initiates a login shell, which will source user profile scripts (e.g., `~/.bash_profile`, `~/.profile`). This could unexpectedly output text (like a Message of the Day, or initialization logs) that gets captured in the asciinema recording, potentially corrupting the demonstration. 

Unless login shell behaviors are strictly required (like loading custom `$PATH` from profile scripts), it is safer to use a non-login shell with `bash -c`.

**Suggested change:**
```suggestion
    --command "bash -c $(printf '%q' "$INNER")" "$CAST" )
```

.claude/skills/demo-creator/scripts/record.sh
```diff
@@ -0,0 +1,56 @@
...
+CWD="$PWD"; COLS=100; ROWS=30; FONT=22; IDLE=2; PREFIX="$ANGLE"; BANNER=""
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --cwd) CWD="$2"; shift 2;;
+    --cols) COLS="$2"; shift 2;;
+    --rows) ROWS="$2"; shift 2;;
+    --font) FONT="$2"; shift 2;;
+    --idle) IDLE="$2"; shift 2;;
+    --out-prefix) PREFIX="$2"; shift 2;;
+    --banner) BANNER="$2"; shift 2;;   # shown on screen before the run (headless hides the invocation)
```
dev-punia-altimate, line 22: **[🟠 MEDIUM]** There is no boundary check for the option arguments in this `while` loop. Because the script uses `set -u`, providing a flag (e.g., `--cwd` or `--cols`) as the very last argument without its value will trigger a confusing `unbound variable` error on `$2` and abruptly exit. 

To improve the user experience and provide clearer error messages, consider verifying that `$2` exists before assignment, or using `${2:-}`. For example:

**Suggested change:**
```suggestion
while [ $# -gt 0 ]; do
  case "$1" in
    --cwd) CWD="${2?Missing argument for --cwd}"; shift 2;;
    --cols) COLS="${2?Missing argument for --cols}"; shift 2;;
    --rows) ROWS="${2?Missing argument for --rows}"; shift 2;;
    --font) FONT="${2?Missing argument for --font}"; shift 2;;
    --idle) IDLE="${2?Missing argument for --idle}"; shift 2;;
    --out-prefix) PREFIX="${2?Missing argument for --out-prefix}"; shift 2;;
    --banner) BANNER="${2?Missing argument for --banner}"; shift 2;;   # shown on screen before the run (headless hides the invocation)
```

.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh
```diff
@@ -0,0 +1,62 @@
...
+# --- JDK (Spark needs a JVM). We pin PySpark to the 3.5 line, which supports Java 8/11/17.
+#     Many brew `openjdk@N` aliases actually resolve to the latest JDK (e.g. 23), which Spark
+#     rejects, so we VALIDATE the reported major version and accept only 8/11/17 (then 21). ---
+java_major() { "$1/bin/java" -version 2>&1 | sed -nE 's/.*version "([0-9]+).*/\1/p' | head -1; }
+JAVA_HOME=""
+for cand in \
+  "/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home" \
+  "/opt/homebrew/opt/openjdk@11/libexec/openjdk.jdk/Contents/Home" \
+  "/opt/homebrew/opt/openjdk@21/libexec/openjdk.jdk/Contents/Home"; do
+  [ -x "$cand/bin/java" ] || continue
```
dev-punia-altimate: **[🔴 HIGH]** **Platform dependence and rigid `JAVA_HOME` handling.** The script hardcodes Apple Silicon macOS paths (`/opt/homebrew/...`) for JDK discovery and strictly relies on `brew` to install openjdk if a suitable version is not found. This will cause failures on Linux, Windows, CI containers, and Intel Macs (`/usr/local/opt/...`). Additionally, it completely ignores any existing, valid `JAVA_HOME` environment variable provided by the user.

Consider including the user's `$JAVA_HOME` in the candidate list and generalizing the discovery logic.

**Suggested change:**
```suggestion
ENV_JAVA_HOME="${JAVA_HOME:-}"
JAVA_HOME=""
for cand in \
  "$ENV_JAVA_HOME" \
  "/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home" \
  "/opt/homebrew/opt/openjdk@11/libexec/openjdk.jdk/Contents/Home" \
  "/opt/homebrew/opt/openjdk@21/libexec/openjdk.jdk/Contents/Home"; do
  [ -n "$cand" ] || continue
  [ -x "$cand/bin/java" ] || continue
```

.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh
```diff
@@ -0,0 +1,62 @@
...
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"; source "$DIR/lib.sh"
+
+FIXTURE="${1:?usage: setup_pyspark_fixture.sh <fixture-dir>}"
+mkdir -p "$FIXTURE"; FIXTURE="$(cd "$FIXTURE" && pwd)"
+log "fixture dir: $FIXTURE"
+
+# --- JDK (Spark needs a JVM). We pin PySpark to the 3.5 line, which supports Java 8/11/17.
+#     Many brew `openjdk@N` aliases actually resolve to the latest JDK (e.g. 23), which Spark
+#     rejects, so we VALIDATE the reported major version and accept only 8/11/17 (then 21). ---
+java_major() { "$1/bin/java" -version 2>&1 | sed -nE 's/.*version "([0-9]+).*/\1/p' | head -1; }
```
dev-punia-altimate: **[🟠 MEDIUM]** **Incorrect Java major version parsing for Java 8.** The `java_major` function uses a regex that incorrectly extracts `1` instead of `8` for Java 8 (e.g., `java version "1.8.0_xxx"`). As a result, the subsequent `case` statement will erroneously reject valid Java 8 installations and force an unnecessary JDK download.

You can adjust the regex to account for the `1.x` pattern used by older Java versions.

**Suggested change:**
```suggestion
java_major() { "$1/bin/java" -version 2>&1 | sed -nE 's/.*version "(1\.)?([0-9]+).*/\2/p' | head -1; }
```

.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# --- Python venv + pyspark (3.5 line) via uv (fast). Fallback to python3 -m venv. ---
+VENV="$FIXTURE/.venv"
+PYSPARK_SPEC='pyspark>=3.5,<4'
+if command -v uv >/dev/null 2>&1; then
+  ( cd "$FIXTURE" && uv venv "$VENV" >/dev/null 2>&1 || true; \
+    VIRTUAL_ENV="$VENV" uv pip install --python "$VENV/bin/python" "$PYSPARK_SPEC" )
+else
+  python3 -m venv "$VENV"; "$VENV/bin/pip" install -q --upgrade pip "$PYSPARK_SPEC"
+fi
```
dev-punia-altimate, line 70: **[🟠 MEDIUM]** **Missing prerequisite validation in the fallback path.** When the `uv` tool is unavailable, the script silently falls back to executing `python3 -m venv` without verifying if the `python3` command actually exists on the host. This will result in a messy 'command not found' failure mid-execution.

You can use the standard `need` helper from `lib.sh` to validate `python3` presence.

**Suggested change:**
```suggestion
else
  need python3
  python3 -m venv "$VENV"; "$VENV/bin/pip" install -q --upgrade pip "$PYSPARK_SPEC"
fi
```

.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh
```diff
@@ -0,0 +1,80 @@
...
+else
+  # In-fork isolation: run altimate-code but strip the capability under test.
+  B_ARGS=(); [ -n "$BASE_AGENT" ] && B_ARGS+=(--agent "$BASE_AGENT")
+  M_ARGS=(); [ -n "$MODEL" ] && M_ARGS+=(--model "$MODEL")
+  if [ -n "$BASE_DENY" ]; then
+    export OPENCODE_PERMISSION="{\"tools\":{$(echo "$BASE_DENY" | awk -F, '{for(i=1;i<=NF;i++){printf "%s\"%s\":\"deny\"", (i>1?",":""), $i}}')}}"
+    log "  baseline OPENCODE_PERMISSION=$OPENCODE_PERMISSION"
+  fi
+  ( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
+      "${M_ARGS[@]}" "${B_ARGS[@]}" "$PROMPT" ) | tee "$RUNS/$ANGLE.baseline.json" || true
```
dev-punia-altimate: **[🔴 HIGH]** Expanding empty arrays (`"${M_ARGS[@]}"` and `"${B_ARGS[@]}"`) with `set -u` enabled will trigger an "unbound variable" error and crash the script on Bash versions older than 4.4 (which includes the default Bash 3.2 on macOS).

Consider using fallback syntax `"${M_ARGS[@]+"${M_ARGS[@]}"}"` or checking for length before expansion to ensure cross-compatibility.

**Suggested change:**
```suggestion
  ( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
      ${M_ARGS[@]+"${M_ARGS[@]}"} ${B_ARGS[@]+"${B_ARGS[@]}"} "$PROMPT" ) | tee "$RUNS/$ANGLE.baseline.json" || true
```

.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh
```diff
@@ -0,0 +1,80 @@
...
+  unset OPENCODE_PERMISSION
```
dev-punia-altimate, line 83: **[🟠 MEDIUM]** The `OPENCODE_PERMISSION` variable is conditionally exported in the main script environment and then unconditionally cleared via `unset OPENCODE_PERMISSION`. 

If a user had this variable configured in their environment prior to running the script, it will be unintentionally wiped out for the remainder of the script (e.g., during the subsequent `ALTIMATE run`). 

To avoid side effects, consider moving the conditional `export` logic inside the `( ... )` subshell so it naturally scopes to the baseline run without affecting the parent environment.

**Suggested change:**
```suggestion
  (
    if [ -n "$BASE_DENY" ]; then
      export OPENCODE_PERMISSION="{\"tools\":{$(echo "$BASE_DENY" | awk -F, '{for(i=1;i<=NF;i++){printf "%s\"%s\":\"deny\"", (i>1?",":""), $i}}')}}"
      log "  baseline OPENCODE_PERMISSION=$OPENCODE_PERMISSION"
    fi
    cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
      ${M_ARGS[@]+"${M_ARGS[@]}"} ${B_ARGS[@]+"${B_ARGS[@]}"} "$PROMPT" 
  ) | tee "$RUNS/$ANGLE.baseline.json" || true
```

.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh
```diff
@@ -0,0 +1,80 @@
...
+    --cwd) CWD="$2"; shift 2;;
+    --prompt) PROMPT="$2"; shift 2;;
+    --model) MODEL="$2"; shift 2;;
+    --baseline-model) BASE_MODEL="$2"; shift 2;;
+    --reset-from) RESET_FROM="$2"; shift 2;;
+    --max-turns) MAXT="$2"; shift 2;;
+    --baseline-mode) BASE_MODE="$2"; shift 2;;     # claude (default) | disable
+    --baseline-deny) BASE_DENY="$2"; shift 2;;     # only for --baseline-mode disable
+    --baseline-agent) BASE_AGENT="$2"; shift 2;;   # only for --baseline-mode disable
+    -*) echo "unknown opt: $1" >&2; exit 2;;
```
dev-punia-altimate, line 33: **[🟠 MEDIUM]** Argument parsing directly accesses `$2` and shifts by 2 for flags that expect values. 

Since `set -u` is enabled, if any of these flags are passed as the very last argument without a following value, the script will crash abruptly with a Bash "unbound variable" error rather than printing a helpful validation message. 

Consider checking `$# -ge 2` before processing these flags, or using standard fallback syntax like `${2:-}`.

**Suggested change:**
```suggestion
  case "$1" in
    --cwd) [ $# -ge 2 ] || { echo "ERROR: --cwd requires a value" >&2; exit 2; }; CWD="$2"; shift 2;;
    --prompt) [ $# -ge 2 ] || { echo "ERROR: --prompt requires a value" >&2; exit 2; }; PROMPT="$2"; shift 2;;
    --model) [ $# -ge 2 ] || { echo "ERROR: --model requires a value" >&2; exit 2; }; MODEL="$2"; shift 2;;
    --baseline-model) [ $# -ge 2 ] || { echo "ERROR: --baseline-model requires a value" >&2; exit 2; }; BASE_MODEL="$2"; shift 2;;
    --reset-from) [ $# -ge 2 ] || { echo "ERROR: --reset-from requires a value" >&2; exit 2; }; RESET_FROM="$2"; shift 2;;
    --max-turns) [ $# -ge 2 ] || { echo "ERROR: --max-turns requires a value" >&2; exit 2; }; MAXT="$2"; shift 2;;
    --baseline-mode) [ $# -ge 2 ] || { echo "ERROR: --baseline-mode requires a value" >&2; exit 2; }; BASE_MODE="$2"; shift 2;;
    --baseline-deny) [ $# -ge 2 ] || { echo "ERROR: --baseline-deny requires a value" >&2; exit 2; }; BASE_DENY="$2"; shift 2;;
    --baseline-agent) [ $# -ge 2 ] || { echo "ERROR: --baseline-agent requires a value" >&2; exit 2; }; BASE_AGENT="$2"; shift 2;;
    -*) echo "unknown opt: $1" >&2; exit 2;;
```

.claude/skills/demo-creator/scripts/baseline_vs_altimate.sh
```diff
@@ -0,0 +1,80 @@
...
+      "${M_ARGS[@]}" "${B_ARGS[@]}" "$PROMPT" ) | tee "$RUNS/$ANGLE.baseline.json" || true
+  unset OPENCODE_PERMISSION
+fi
+
+# ---------------- altimate ----------------
+log "=== ALTIMATE run ==="
+reset_fixture
+M_ARGS=(); [ -n "$MODEL" ] && M_ARGS+=(--model "$MODEL")
+( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
+    "${M_ARGS[@]}" "$PROMPT" ) | tee "$RUNS/$ANGLE.json" || true
```
dev-punia-altimate: **[🔴 HIGH]** This array expansion will also trigger an "unbound variable" error under Bash 3.2 when the array is empty. Consider applying the same fallback syntax here.

**Suggested change:**
```suggestion
( cd "$CWD" && altimate-code run --yolo --max-turns "$MAXT" --format json --trace \
    ${M_ARGS[@]+"${M_ARGS[@]}"} "$PROMPT" ) | tee "$RUNS/$ANGLE.json" || true
```

.claude/skills/demo-creator/scripts/build_artifact.sh
```diff
@@ -0,0 +1,67 @@
...
+[ $# -ge 2 ] || { echo "usage: build_artifact.sh <topic> <template.html> [--out NAME] [--render] [--width W] [--height H]" >&2; exit 2; }
+TOPIC="$1"; TPL="$2"; shift 2
+OUT="demo.html"; RENDER=0; W=820; H=5200
+while [ $# -gt 0 ]; do case "$1" in
+  --out) OUT="$2"; shift 2;;
+  --render) RENDER=1; shift;;
+  --width) W="$2"; shift 2;;
+  --height) H="$2"; shift 2;;
+  *) echo "unknown opt: $1" >&2; exit 2;;
+esac; done
```
dev-punia-altimate, line 22: **[🟠 MEDIUM]** Since `set -u` is enabled at the top of the script, referencing `$2` without checking if it exists will throw an abrupt bash "unbound variable" error if an option like `--out` is passed as the very last argument. Additionally, `shift 2` will fail if there are less than 2 arguments left. It's safer to explicitly check `$#` to provide a clear error message.

**Suggested change:**
```suggestion
while [ $# -gt 0 ]; do case "$1" in
  --out) [ $# -ge 2 ] || { echo "missing value for $1" >&2; exit 2; }; OUT="$2"; shift 2;;
  --render) RENDER=1; shift;;
  --width) [ $# -ge 2 ] || { echo "missing value for $1" >&2; exit 2; }; W="$2"; shift 2;;
  --height) [ $# -ge 2 ] || { echo "missing value for $1" >&2; exit 2; }; H="$2"; shift 2;;
  *) echo "unknown opt: $1" >&2; exit 2;;
esac; done
```

.claude/skills/demo-creator/scripts/build_artifact.sh
```diff
@@ -0,0 +1,67 @@
...
+
+python3 - "$TPL" "$DD" "$OUTPATH" <<'PY'
+import base64, mimetypes, os, re, sys
+tpl, base, out = sys.argv[1:4]
+html = open(tpl, encoding="utf-8").read()
+mimetypes.add_type("image/svg+xml", ".svg")
+inlined = [0]; missing = []
+def datauri(rel):
+    p = os.path.normpath(os.path.join(base, rel))
+    if not (p.startswith(os.path.abspath(base)) and os.path.isfile(p)):
```
dev-punia-altimate: **[🟠 MEDIUM]** `os.path.normpath` does not resolve relative paths to absolute ones. If `base` happens to be a relative path (e.g., if invoked dynamically differently than current shell expectation), `p` will also be relative. Comparing a relative `p` against the absolute path from `os.path.abspath(base)` via `startswith()` will always evaluate to `False`, failing to inline valid images. Consider using `os.path.abspath` to correctly resolve `p`.

**Suggested change:**
```suggestion
def datauri(rel):
    p = os.path.abspath(os.path.join(base, rel))
    if not (p.startswith(os.path.abspath(base)) and os.path.isfile(p)):
```

.claude/skills/demo-creator/scripts/build_artifact.sh
```diff
@@ -0,0 +1,67 @@
...
+    pre, q, url, post = m.group(1), m.group(2), m.group(3), m.group(4)
+    if re.match(r'^(https?:|data:|//)', url): return m.group(0)
+    d = datauri(url)
+    return m.group(0) if d is None else "%s%s%s%s" % (pre, q, d, post)
+# src="..." and CSS url(...) for local image paths
+html = re.sub(r'(src=)(["\'])([^"\']+\.(?:gif|png|jpe?g|svg|webp))(["\'])', repl, html, flags=re.I)
+html = re.sub(r'(url\()(["\']?)([^)"\']+\.(?:gif|png|jpe?g|svg|webp))(["\']?\))', repl, html, flags=re.I)
+open(out, "w", encoding="utf-8").write(html)
+print("inlined %d asset(s); %.2f MB" % (inlined[0], len(html)/1e6))
+if missing: print("WARNING: could not inline (will break under CSP): %s" % ", ".join(missing), file=sys.stderr)
```
dev-punia-altimate, line 54: **[🟠 MEDIUM]** The bash script runs with `set -e`, but the embedded Python script exits with a `0` status code even when local assets cannot be found. Continuing the build process with missing assets will silently produce artifacts that break under strict CSP. Exiting with a non-zero status code here would correctly halt the build process.

**Suggested change:**
```suggestion
if missing:
    print("ERROR: could not inline (will break under CSP): %s" % ", ".join(missing), file=sys.stderr)
    sys.exit(1)
```

.claude/skills/demo-creator/scripts/build_artifact.sh
```diff
@@ -0,0 +1,67 @@
...
+PY
+
+if [ "$RENDER" = "1" ]; then
+  CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
+  SHOT="$DD/clips/frames/_artifact_render.png"; mkdir -p "$(dirname "$SHOT")"
```
dev-punia-altimate: **[🔴 HIGH]** The Chrome executable path is hardcoded to the macOS-specific application path. If this script is executed in Linux or Windows environments, the `--render` step will silently fail to find Chrome and skip rendering. It's recommended to add a cross-platform fallback to look up valid executables (e.g., `google-chrome` or `chromium`) in `$PATH`.

**Suggested change:**
```suggestion
  CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
  if [ ! -x "$CHROME" ]; then
    CHROME="$(command -v google-chrome || command -v chromium || echo "$CHROME")"
  fi
  SHOT="$DD/clips/frames/_artifact_render.png"; mkdir -p "$(dirname "$SHOT")"
```

anandgupta42:
Thanks for the reviews. Pushed `13772ba` addressing the substantive findings (deduped across Codex / CodeRabbit / Kilo / cubic):

**Correctness / safety**
- `lib.sh` `demo_dir` now rejects non-slug topics (`..`, `/`) — paths can't escape `demos/`.
- `build_artifact.sh` path-containment uses `realpath` + `os.path.commonpath` instead of a string-prefix (fixes the sibling-dir escape).
- `baseline_vs_altimate.sh` `OPENCODE_PERMISSION` fixed to top-level tool keys (`{"bash":"deny"}`) — the old `{"tools":{…}}` created a rule named `tools` and didn't deny anything.

**bash 3.2 (macOS default)**
- Empty-array expansions guarded with `${arr[@]+"${arr[@]}"}` in `baseline_vs_altimate.sh`; frame count in `inspect.sh` switched to `find` so a no-match doesn't trip `set -e`.

**Comparison validity (the P1/P2 cluster)**
- Same-run authenticity: `SKILL.md` Phase 5 no longer tells you to re-run with `--format json` — proof comes from the SAME recorded run's `--trace`/cast.
- `--reset-from` and `--model` now emit loud warnings when omitted (contaminated fixture / mismatched model invalidate the control).
- The altimate run failing now fails the script (no more `|| true` masking it); the baseline is still allowed to struggle.
- `record.sh` adds `asciinema --return` + propagates a failed recorded command (a broken run won't render as a good clip).
- `inspect.sh` cross-checks a `.baseline` clip against the baseline JSON, not the altimate one.

**Portability**
- `setup_pyspark_fixture.sh` now honors an existing `JAVA_HOME` / `java` / `/usr/libexec/java_home` before Homebrew (works on Linux / Intel mac), robustly parses `1.8`-style versions, and validates prerequisites.
- `build_artifact.sh --render` discovers Chrome/Chromium on `PATH` and hard-fails if none (render-inspection is mandatory before publishing) instead of silently skipping.

**Docs / template**
- `blog.md.tmpl` baseline section marked comparison-only (showcase is the default, produces no baseline clip); added a language tag to the SKILL.md tree block (MD040).

A few low-value nitpicks (e.g. `bash -lc` vs `bash -c`) I left as-is; happy to take them if you feel strongly.

.claude/skills/demo-creator/scripts/record.sh
```diff
@@ -0,0 +1,68 @@
...
+    *) echo "unknown opt: $1" >&2; exit 2;;
+  esac
+done
+[ $# -ge 1 ] || { echo "ERROR: no command after --" >&2; exit 2; }
+
+need asciinema asciinema; need agg agg
+
+# asciinema (Rust) rejects args that aren't valid UTF-8 under a non-UTF-8 locale. Force one
+# so prompts/banners with non-ASCII (em dashes, arrows, accents) don't error out.
+export LC_ALL="${LC_ALL:-en_US.UTF-8}" LANG="${LANG:-en_US.UTF-8}"
```
dev-punia-altimate, line 33: **[🟠 MEDIUM]** The comment states "Force one", but the parameter expansion `${LC_ALL:-en_US.UTF-8}` only provides a fallback when the variable is unset or empty. If a user already has a non-UTF-8 locale set (e.g., `LC_ALL=C`), this line will preserve that non-UTF-8 value, which may still cause the asciinema error you are trying to prevent.

To strictly enforce a UTF-8 locale, you should assign the values directly.

**Suggested change:**
```suggestion
export LC_ALL="en_US.UTF-8" LANG="en_US.UTF-8"
```

.claude/skills/demo-creator/scripts/inspect.sh
```diff
@@ -0,0 +1,49 @@
...
+DD="$(demo_dir "$TOPIC")"
+GIF="$DD/clips/$PREFIX.gif"
+FRAMES="$DD/clips/frames/$PREFIX"
+[ -f "$GIF" ] || { echo "ERROR: no GIF at $GIF (record it first)" >&2; exit 1; }
+mkdir -p "$FRAMES"; rm -f "$FRAMES"/f_*.png
+
+log "extracting frames @ ${FPS}fps -> $FRAMES"
+ffmpeg -y -loglevel error -i "$GIF" -vf "fps=$FPS" "$FRAMES/f_%03d.png"
+# Count without letting a no-match ls abort us under `set -e`/pipefail (lib.sh sets both).
+COUNT=$(find "$FRAMES" -name 'f_*.png' 2>/dev/null | wc -l | tr -d ' ')
```
dev-punia-altimate, line 29: **[🟠 MEDIUM]** By default, `find` searches recursively and matches all file types (including directories). To make this strictly equivalent to the original `ls "$FRAMES"/f_*.png` command and avoid edge cases (like counting the directory itself if `$PREFIX` happens to match the `f_*.png` pattern, or unexpectedly counting files in subdirectories if any are created in the future), it is recommended to limit the search depth and restrict matches to regular files.

**Suggested change:**
```suggestion
COUNT=$(find "$FRAMES" -maxdepth 1 -type f -name 'f_*.png' 2>/dev/null | wc -l | tr -d ' ')
```

.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh
```diff
@@ -0,0 +1,91 @@
...
+  if compat "$m"; then JAVA_HOME_FOUND="$1"; log "using JDK $m at $1"; fi
+}
+# 1) caller's JAVA_HOME  2) java already on PATH  3) macOS java_home for 17/11/8
+add_cand "${JAVA_HOME:-}"
+if [ -z "$JAVA_HOME_FOUND" ] && command -v java >/dev/null 2>&1; then
+  add_cand "$(cd "$(dirname "$(command -v java)")/.." && pwd)"
+fi
+if [ -z "$JAVA_HOME_FOUND" ] && [ -x /usr/libexec/java_home ]; then
+  for v in 17 11 21 8; do add_cand "$(/usr/libexec/java_home -v "$v" 2>/dev/null)"; done
+fi
```
dev-punia-altimate, line 38: **[🔴 HIGH]** The `PATH` check occurs before the macOS `/usr/libexec/java_home` check. If `java` is found at `/usr/bin/java` (which acts as a stub on macOS and typically a symlink on Linux), `dirname` resolution assigns `/usr` to `JAVA_HOME`. This intercepts valid macOS setups, preventing them from resolving the true JDK directory, and assigns a directory without the standard JDK layout (like `lib/`), which breaks Spark and other JVM tools.

To fix this, we should prioritize the macOS `java_home` check over the `PATH` fallback. Additionally, resolving symlinks (via `readlink` or `realpath`) when computing `JAVA_HOME` from `PATH` is crucial to ensure it works correctly on Linux.

**Suggested change:**
```suggestion
# 1) caller's JAVA_HOME  2) macOS java_home for 17/11/8  3) java already on PATH
add_cand "${JAVA_HOME:-}"
if [ -z "$JAVA_HOME_FOUND" ] && [ -x /usr/libexec/java_home ]; then
  for v in 17 11 21 8; do add_cand "$(/usr/libexec/java_home -v "$v" 2>/dev/null)"; done
fi
if [ -z "$JAVA_HOME_FOUND" ] && command -v java >/dev/null 2>&1; then
  java_bin="$(command -v java)"
  java_bin="$(readlink -f "$java_bin" 2>/dev/null || realpath "$java_bin" 2>/dev/null || echo "$java_bin")"
  add_cand "$(cd "$(dirname "$java_bin")/.." && pwd)"
fi
```

.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh
```diff
@@ -0,0 +1,91 @@
...
+# 4) Homebrew kegs (Apple Silicon + Intel prefixes)
+if [ -z "$JAVA_HOME_FOUND" ]; then
+  for pfx in /opt/homebrew/opt /usr/local/opt; do
+    for v in 17 11 21; do add_cand "$pfx/openjdk@$v/libexec/openjdk.jdk/Contents/Home"; done
+  done
+fi
```
dev-punia-altimate, line 44: **[🟠 MEDIUM]** The fast-path check for Homebrew kegs hardcodes macOS Homebrew prefixes (`/opt/homebrew/opt` and `/usr/local/opt`) but omits the standard Linuxbrew prefix (`/home/linuxbrew/.linuxbrew/opt`). On Linux systems using Homebrew, this bypasses the fast path and unnecessarily invokes the much slower `brew` fallback.

Furthermore, Linuxbrew typically doesn't use the macOS-specific `libexec/openjdk.jdk/Contents/Home` layout, so adding the base keg path (`$pfx/openjdk@$v`) is required to properly detect the JDK on Linux platforms.

**Suggested change:**
```suggestion
# 4) Homebrew kegs (Apple Silicon + Intel + Linuxbrew prefixes)
if [ -z "$JAVA_HOME_FOUND" ]; then
  for pfx in /opt/homebrew/opt /usr/local/opt /home/linuxbrew/.linuxbrew/opt; do
    for v in 17 11 21; do
      add_cand "$pfx/openjdk@$v/libexec/openjdk.jdk/Contents/Home"
      add_cand "$pfx/openjdk@$v"
    done
  done
fi
```

dev-punia-altimate (commented):

- 0 without a line anchor

#### 1. `.claude/skills/demo-creator/scripts/record.sh` (L33)

**[🟠 MEDIUM]** The comment states "Force one", but the parameter expansion `${LC_ALL:-en_US.UTF-8}` only provides a fallback when the variable is unset or empty. If a user already has a non-UTF-8 locale set (e.g., `LC_ALL=C`), this line will preserve that non-UTF-8 value, which may still cause the asciinema error you are trying to prevent.

To strictly enforce a UTF-8 locale, you should assign the values directly.

**Suggested change:**
```suggestion
export LC_ALL="en_US.UTF-8" LANG="en_US.UTF-8"
```

#### 2. `.claude/skills/demo-creator/scripts/inspect.sh` (L29)

**[🟠 MEDIUM]** By default, `find` searches recursively and matches all file types (including directories). To make this strictly equivalent to the original `ls "$FRAMES"/f_*.png` command and avoid edge cases (like counting the directory itself if `$PREFIX` happens to match the `f_*.png` pattern, or unexpectedly counting files in subdirectories if any are created in the future), it is recommended to limit the search depth and restrict matches to regular files.

**Suggested change:**
```suggestion
COUNT=$(find "$FRAMES" -maxdepth 1 -type f -name 'f_*.png' 2>/dev/null | wc -l | tr -d ' ')
```

#### 3. `.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh` (L31-L38)

**[🔴 HIGH]** The `PATH` check occurs before the macOS `/usr/libexec/java_home` check. If `java` is found at `/usr/bin/java` (which acts as a stub on macOS and typically a symlink on Linux), `dirname` resolution assigns `/usr` to `JAVA_HOME`. This intercepts valid macOS setups, preventing them from resolving the true JDK directory, and assigns a directory without the standard JDK layout (like `lib/`), which breaks Spark and other JVM tools.

To fix this, we should prioritize the macOS `java_home` check over the `PATH` fallback. Additionally, resolving symlinks (via `readlink` or `realpath`) when computing `JAVA_HOME` from `PATH` is crucial to ensure it works correctly on Linux.

**Suggested change:**
```suggestion
# 1) caller's JAVA_HOME  2) macOS java_home for 17/11/8  3) java already on PATH
add_cand "${JAVA_HOME:-}"
if [ -z "$JAVA_HOME_FOUND" ] && [ -x /usr/libexec/java_home ]; then
  for v in 17 11 21 8; do add_cand "$(/usr/libexec/java_home -v "$v" 2>/dev/null)"; done
fi
if [ -z "$JAVA_HOME_FOUND" ] && command -v java >/dev/null 2>&1; then
  java_bin="$(command -v java)"
  java_bin="$(readlink -f "$java_bin" 2>/dev/null || realpath "$java_bin" 2>/dev/null || echo "$java_bin")"
  add_cand "$(cd "$(dirname "$java_bin")/.." && pwd)"
fi
```

#### 4. `.claude/skills/demo-creator/scripts/setup_pyspark_fixture.sh` (L39-L44)

**[🟠 MEDIUM]** The fast-path check for Homebrew kegs hardcodes macOS Homebrew prefixes (`/opt/homebrew/opt` and `/usr/local/opt`) but omits the standard Linuxbrew prefix (`/home/linuxbrew/.linuxbrew/opt`). On Linux systems using Homebrew, this bypasses the fast path and unnecessarily invokes the much slower `brew` fallback.

Furthermore, Linuxbrew typically doesn't use the macOS-specific `libexec/openjdk.jdk/Contents/Home` layout, so adding the base keg path (`$pfx/openjdk@$v`) is required to properly detect the JDK on Linux platforms.

**Suggested change:**
```suggestion
# 4) Homebrew kegs (Apple Silicon + Intel + Linuxbrew prefixes)
if [ -z "$JAVA_HOME_FOUND" ]; then
  for pfx in /opt/homebrew/opt /usr/local/opt /home/linuxbrew/.linuxbrew/opt; do
    for v in 17 11 21; do
      add_cand "$pfx/openjdk@$v/libexec/openjdk.jdk/Contents/Home"
      add_cand "$pfx/openjdk@$v"
    done
  done
fi
```

</details>

dev-punia-altimate:

No supported files changed.

PR #1001, opened 2026-07-09
feat: [AI-7520] Altimate LLM Gateway CLI signup + onboarding UX

saravmajestic:
## Summary
CLI onboarding for the **Altimate LLM Gateway** + a welcoming first-run experience, plus quick sign-in/out commands.

### Gateway sign-in
- Loopback (`:7317`) + browser → web sign-up page → credential delivered back to the CLI; saved to `~/.altimate/altimate.json` (overridable via `ALTIMATE_WEB_URL`).
- On success, an inline green **"Authentication successful"** confirmation that **auto-selects a model** and auto-closes (no drop into the model picker).
- Don't force Google: the CLI lands on the sign-up page and lets the user choose their method.

### First-run experience
- **`WelcomePanel`** boot box on the home screen with readiness-aware tips (`/connect` → `/discover`); update toast suppressed until a model is ready.
- **`useReady`/`markSetupComplete`** gate the first-run chat lock (unlocks after connecting, choosing Big Pickle, or picking a ready model).

### Provider picker
- `/connect` opens the curated **`DialogModelWelcome`** — top 5 providers (Altimate Gateway first) + "Search all providers…", a **Big Pickle** fallback, and a **bright-green ✓ "selected"** marker on the currently-active provider/model.
- The `/model` dialog is restructured into **READY / NEEDS-SETUP** sections.

### New TUI commands
- **`/auth`** — start the Altimate LLM Gateway sign-in directly (`DialogAltimateAuth`).
- **`/logout`** — clear the saved gateway credential (`AltimateApi.clearCredentials`), disconnect, and reload.

### Rebase safety
- Onboarding code isolated into an altimate-owned **`component/altimate-onboarding.tsx`**; `dialog-model.tsx` trimmed to pristine `useConnected` + the marked `DialogModel` restructure, so future upstream (opencode) merges have a minimal, clearly-marked conflict surface. No behavior change.

## Test plan
- [x] `/connect` → Altimate LLM Gateway → browser → provisioned + connected
- [x] Gateway key authenticates (`/agents/v1/models` → `200`)
- [x] `/auth` launches gateway sign-in; `/logout` clears creds + disconnects
- [x] Selected provider shows the green ✓
- [x] `bun run typecheck` clean

## Summary by CodeRabbit

* **New Features**
  * Added Altimate onboarding with a curated welcome picker, a guided scan/skip Part 2 dialog, and updated connect/auth/logout flow.
  * Introduced Altimate Gateway browser sign-in and improved onboarding routing for model/provider selection.
  * Added starter-sample setup with safe reuse and “install alongside” support (including DuckDB sample).
  * Added local binary installer script (`replace.sh`).
* **Bug Fixes**
  * Fixed readiness/“connected” detection so missing pricing info no longer marks models as connected.
  * Hardened first-run behavior so update confirmations don’t interrupt initial onboarding.
* **Chores**
  * Added and updated tests for the starter sample and onboarding flows; improved ignore rules for demo assets.


sahrizvi (commented):
## Consensus code review — PR #1001

Independent reviews from 6 models (Claude, GPT 5.4 Codex, Kimi K2.5, MiniMax M2.7, Qwen 3.6, MiMo V2.5 Pro), converged and then re-validated line-by-line against the checked-out code.

**Verdict: changes requested.** The onboarding refactor and OAuth loopback design are well-structured (crypto-random `state`, CSRF state check, `0o600` credential file, cleanly isolated onboarding). But the new loopback auth server has correctness/reliability bugs and localhost security-hardening gaps — **7 major, 9 minor** — all in the auth path. See inline comments.

**Nits (not inlined):** recent-models section dropped from the model picker; Big Pickle won't show a ✓ in the full picker after selection; error page doesn't `window.close()` like the success page; documented circular import between `altimate-onboarding.tsx` and `dialog-model.tsx`; silent `open()` failure (mitigated — the URL is still rendered); no `ALTIMATE_CALLBACK_PORT` override.

**Positives:** crypto-random state + CSRF check, `0o600` credentials, clean rebase-safe isolation of onboarding, reuse of existing provider `onSelect` handlers, reliable server teardown on the normal path.

**Missing tests:** the entire OAuth loopback flow (early callback, invalid/missing state, error branch, timeout, port collision, persistence) and the `/auth` · `/logout` · Big Pickle · mixed-row picker paths are untested.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,184 @@
...
       provider: "altimate-backend",
       methods: [
         {
+          type: "oauth",
+          label: "Altimate LLM Gateway",
+          async authorize() {
+            // Bind the port BEFORE opening the browser so the credential can
+            // only be delivered to this process.
+            const state = randomBytes(16).toString("hex")
+            await startCallbackServer()
```
sahrizvi, line 212: **The browser can hand back the login before we're listening for it**

_MAJOR_ — **Race — `pending` is registered too late.** `startCallbackServer()` binds the port and starts accepting requests here inside `authorize()`, but the module-level `pending` (state + `resolve`/`reject`) is only set later when the framework calls the returned `callback()` (`waitForCallback`, ~L114). If the browser redirects **before** `callback()` runs (an already-signed-in user gets an instant redirect, or any scheduling gap), `/callback` sees `pending === undefined` and rejects as "Invalid state — possible CSRF", silently dropping the credential. The singleton also can't support two concurrent `/auth` flows — the second overwrites the first.

**Fix:** register `pending` keyed by `state` (e.g. a `Map`) synchronously in `authorize()` before opening the browser; have `callback()` await that promise, and buffer an early callback by `state` until it's awaited.

packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
```diff
@@ -135,6 +194,23 @@ function AutoMethod(props: AutoMethodProps) {
...
+    // On success, confirm inline (green) and auto-close after a moment rather than
+    // opening the model picker. Auto-select a model so the user can chat right away.
+    if (props.providerID === "altimate-backend") {
+      const provider = sync.data.provider.find((p) => p.id === props.providerID)
+      const model = provider
+        ? Object.entries(provider.models).find(([, info]) => info.status !== "deprecated")?.[0]
+        : undefined
+      if (model) local.model.set({ providerID: props.providerID, modelID: model }, { recent: true })
+      setConnected(true)
+      setTimeout(() => dialog.clear(), 5000)
```
sahrizvi: **Backing out of the sign-in screen doesn't actually stop the sign-in**

_MAJOR_ — **No cancellation/cleanup when the dialog is dismissed.** Pressing `esc` (or unmounting) only calls `dialog.clear()`; it never cancels the pending callback, so the loopback server can stay open for ~5 minutes and a late browser callback can still mutate app state after dismissal. This success-path `setTimeout(() => dialog.clear(), 5000)` is also never retained or cleared on unmount, so it can fire later and dismiss an unrelated dialog opened in the meantime.

**Fix:** guard post-await UI updates with an `onCleanup` disposed flag, `clearTimeout` in `onCleanup`, and stop the callback server / abort the SDK callback when the dialog is dismissed.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,184 @@
...
+  state: string
+  resolve: (creds: CallbackResult) => void
+  reject: (err: Error) => void
+}
+
+let server: ReturnType<typeof createServer> | undefined
+let pending: Pending | undefined
+
+async function startCallbackServer(): Promise<void> {
+  if (server) return
```
sahrizvi, line 98: **If port 7317 is busy, sign-in silently hangs forever with no retry**

_MAJOR_ — **Port collision leaves a broken, silently-failing server.** `server` is assigned by `createServer()` before `listen()`. On `EADDRINUSE` the listen promise rejects (L94-96) but `server` is never reset to `undefined`, so this `if (server) return` early-returns on the next attempt and the flow proceeds as if listening — the callback never arrives and the user waits out the 5-minute timeout with no message.

**Fix:** in the error handler set `server = undefined` (and close it), and surface a clear "port 7317 in use" error to the TUI. Consider an ephemeral port if the redirect contract allows.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,184 @@
...
+const DEFAULT_WEB_URL = "https://app.myaltimate.com"
+// Fallback gateway API base if the callback omits one.
+const DEFAULT_API_URL = "https://api.myaltimate.com"
+
+const HTML_SUCCESS = `<!doctype html><meta charset="utf-8"><title>Altimate Code</title>
+<body style="font-family:system-ui;text-align:center;padding:64px">
+<h2>Signed in ✓</h2><p>You can return to your terminal.</p>
+<script>setTimeout(()=>window.close(),1500)</script></body>`
+
+const HTML_ERROR = (msg: string) => `<!doctype html><meta charset="utf-8"><title>Altimate Code</title>
```
sahrizvi, line 73: **The error page echoes attacker-controlled text straight into HTML**

_MAJOR_ — **Reflected, unescaped input in the callback error page (localhost XSS).** `HTML_ERROR(msg)` interpolates `msg` into HTML with no escaping, and the value ultimately comes from `url.searchParams` (the `error` branch at L57, which runs *before* the state check). Any page that drives the browser to `http://localhost:7317/callback?error=<script>...` during the auth window yields reflected XSS on the localhost origin. Impact is limited (transient page, no secrets/cookies there) but it's avoidable.

**Fix:** HTML-escape all reflected values, and validate `state` before handling the `error` branch.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,184 @@
...
+    }
+
+    const current = pending
+    pending = undefined
+    current.resolve({ api_url: apiUrl, instance, api_key: apiKey })
+    html(200, HTML_SUCCESS)
+  })
+
+  await new Promise<void>((resolve, reject) => {
+    server!.listen(CALLBACK_PORT, () => resolve())
```
sahrizvi: **The sign-in callback server is reachable from the whole network, not just this machine**

_MAJOR_ — **Callback server listens on all interfaces, not loopback.** `listen(CALLBACK_PORT)` with no host binds to `::`/`0.0.0.0`, so the auth callback server is reachable from the LAN — despite the comment claiming "the credential can only be delivered to this process." The 128-bit `state` still protects credential *delivery*, but the listener still processes remote requests (probing, plus the pre-state-check `error` abort below).

**Fix:** `server!.listen(CALLBACK_PORT, "127.0.0.1", () => resolve())`.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,184 @@
...
+    // stray/malicious local request that didn't originate from our browser tab.
+    if (!pending || !state || state !== pending.state) {
+      const msg = "Invalid state — possible CSRF"
+      pending?.reject(new Error(msg))
+      pending = undefined
+      html(400, HTML_ERROR(msg))
+      return
+    }
+
+    const apiKey = url.searchParams.get("key")
```
sahrizvi: **The API key travels in the URL, where it can end up in history and logs**

_MAJOR_ — **API key delivered in the callback URL query string.** The gateway credential arrives as `?key=...` and is read here, exposing the token to browser history, the address bar, extensions, and any local request logging. Even on loopback this is avoidable secret-in-URL exposure.

**Fix:** deliver a short-lived one-time code in the URL and exchange it server-side for the key, or have the web page POST the credential to the loopback endpoint instead of putting it in the query string.

packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
```diff
@@ -102,6 +122,40 @@ export function DialogProvider() {
...
+// altimate_change start — /auth entry: go straight to the Altimate LLM Gateway
+// sign-in (the OAuth loopback method, index 0), skipping the provider picker.
+export function DialogAltimateAuth() {
+  const { theme } = useTheme()
+  const sdk = useSDK()
+  const dialog = useDialog()
+
+  onMount(async () => {
+    const providerID = "altimate-backend"
+    const result = await sdk.client.provider.oauth.authorize({ providerID, method: 0 })
```
sahrizvi: **If sign-in fails to start, the dialog gets stuck on "Starting sign-in…"**

_MINOR_ — **`DialogAltimateAuth` onMount has no error handling.** This `async` `onMount` awaits `oauth.authorize(...)` with no `try/catch`; on rejection (network, or the port issue above) the dialog is stuck on "Starting sign-in…" forever and the rejection is unhandled.

**Fix:** wrap in `try/catch` → show an error toast and `dialog.clear()`.

packages/opencode/src/cli/cmd/tui/app.tsx
```diff
@@ -659,17 +656,47 @@ function App() {
...
       category: "Provider",
     },
+    {
+      title: "Sign out of Altimate LLM Gateway",
+      value: "altimate.logout",
+      slash: {
+        name: "logout",
+      },
+      onSelect: async () => {
+        await AltimateApi.clearCredentials()
```
sahrizvi: **A failed sign-out leaves the app in a half-logged-out state with no feedback**

_MINOR_ — **`/logout` has no error handling.** `clearCredentials()` → `dispose()` → `bootstrap()` run unguarded; a throw in `dispose`/`bootstrap` leaves a partial logout with no toast and an unhandled rejection.

**Fix:** wrap in `try/catch` with an error toast.

packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
```diff
@@ -135,6 +194,23 @@ function AutoMethod(props: AutoMethodProps) {
...
+    // The gateway sign-in already shows the auth URL + "Waiting for authorization…".
+    // On success, confirm inline (green) and auto-close after a moment rather than
+    // opening the model picker. Auto-select a model so the user can chat right away.
+    if (props.providerID === "altimate-backend") {
+      const provider = sync.data.provider.find((p) => p.id === props.providerID)
+      const model = provider
+        ? Object.entries(provider.models).find(([, info]) => info.status !== "deprecated")?.[0]
+        : undefined
+      if (model) local.model.set({ providerID: props.providerID, modelID: model }, { recent: true })
+      setConnected(true)
```
sahrizvi: **We show "connected" even when there's no model to actually use**

_MINOR_ — **Misleading success when no usable model exists.** On gateway connect, the model lookup just above is guarded (`if (model)`), but `setConnected(true)` runs unconditionally here — so if no non-deprecated model was found, the UI shows the green ✓ while leaving the user with no active model.

**Fix:** handle the no-model case explicitly (surface an error / keep the picker open) instead of showing success.

packages/tui/src/component/dialog-model.tsx
```diff
@@ -25,108 +34,93 @@ export function DialogModel(props: { providerID?: string }) {
...
+            .map((o) => ({
+              value: o.value as { providerID: string; modelID: string } | string,
+              title: o.title,
+              description: o.description,
+              category: "NEEDS SETUP",
+              footer: undefined as string | undefined,
+              onSelect: o.onSelect,
+            }))
+          const bigPickle = {
+            value: "big-pickle" as { providerID: string; modelID: string } | string,
```
sahrizvi, line 98: **Favoriting a "needs setup" row is a dead keypress**

_MINOR_ — **Favorite keybind assumes every row is a real model.** READY/NEEDS-SETUP options now mix real `{providerID, modelID}` values with plain strings like this `"big-pickle"` (and provider ids). The favorite keybind casts the selection to a model. `toggleFavorite()` validates before writing, so this doesn't corrupt state — but favoriting a setup row is a silent no-op that looks broken to the user.

**Fix:** guard the favorite keybind with `typeof value !== "string"` (disable favoriting on setup rows).

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,184 @@
...
+                  // reads this first (it carries the instance/tenant + api_url
+                  // the generic auth.json store can't).
+                  await AltimateApi.saveCredentials({
+                    altimateUrl: creds.api_url,
+                    altimateInstanceName: creds.instance,
+                    altimateApiKey: creds.api_key,
+                  })
+                  return { type: "success", key: creds.api_key, provider: "altimate-backend" }
+                } catch {
+                  return { type: "failed" }
```
sahrizvi, line 263: **When sign-in fails, we throw away the reason**

_MINOR_ — **Failed-auth callback discards the error reason.** The catch returns `{ type: "failed" }` with no log and no reason, so CSRF/timeout/unknown failures are indistinguishable and undebuggable in the field.

**Fix:** log the caught error and thread a reason (e.g. `csrf` / `timeout` / `unknown`) through the failure result.

packages/tui/src/component/dialog-model.tsx
```diff
@@ -25,108 +34,93 @@ export function DialogModel(props: { providerID?: string }) {
   const connected = useConnected()
   const providers = createDialogProviderOptions()
 
-  const showExtra = createMemo(() => connected() && !props.providerID)
+  // A provider is "ready" (usable now) when it has valid credentials: it is present
+  // in the live provider list with at least one model — and, for the free OpenCode
+  // provider, with at least one paid model (a Zen key entered).
+  function providerReady(id: string) {
```
sahrizvi, line 33: **A provider with unknown pricing can be treated as "ready" without a key**

_MINOR_ — **`providerReady("opencode")` mishandles undefined cost.** `m.cost?.input !== 0` evaluates to `true` when `cost` is `undefined`, so OpenCode can be counted as "ready" (usable now) even without a Zen key entered.

**Fix:** `m.cost?.input != null && m.cost.input !== 0`. (Same defect exists in `useConnected()` — see the separate comment.)

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,184 @@
...
+              url: authorizeUrl,
+              instructions: "Complete sign-in in your browser to connect Altimate LLM Gateway.",
+              method: "auto",
+              async callback() {
+                try {
+                  const creds = await waitForCallback(state)
+                  // Persist to ~/.altimate/altimate.json — the provider loader
+                  // reads this first (it carries the instance/tenant + api_url
+                  // the generic auth.json store can't).
+                  await AltimateApi.saveCredentials({
```
sahrizvi, line 253: **Browser sign-in skips the tenant-name check that manual entry enforces**

_MINOR_ — **OAuth callback skips the tenant-name validation the manual path enforces.** The manual paste path runs `validateCredentials()` (tenant regex + API check) before saving; this OAuth path calls `saveCredentials()` directly on the callback `instance`. The source is the trusted web app, so risk is low, but the invariant differs between the two entry points.

**Fix:** validate `instance` against `VALID_TENANT_REGEX` before persisting, or document why the OAuth source is trusted.

packages/tui/src/component/altimate-onboarding.tsx
```diff
@@ -0,0 +1,282 @@
...
+import { useKeyboard } from "@opentui/solid"
+import { createDialogProviderOptions } from "./dialog-provider"
+import { DialogModel, useConnected } from "./dialog-model"
+
+// Session-scoped "setup complete" flag. Set when the user picks a ready model,
+// chooses the free Big Pickle option, or finishes the gateway flow. Combined with
+// useConnected() (real credentials) via useReady(), it gates the first-run chat
+// lock. Module-global so it is shared across the app and resets on every process
+// launch (so a fresh relaunch is a clean fresh-user state).
+const [setupComplete, setSetupComplete] = createSignal(false)
```
sahrizvi, line 21: **After signing out, the app still shows "you're all set" guidance**

_MINOR_ — **`setupComplete` is never reset on `/logout`.** This module-global signal is OR'd with `connected()` in `useReady()`. `/logout` clears credentials (so `connected()` → false) but leaves `setupComplete === true`, so `useReady()` stays true for the rest of the process. Verified scope: `useReady()` only drives tips and the update-toast (not the chat/prompt), so the real impact is that the post-logout panel keeps showing "ready" tips (`/discover`) instead of `/connect` until relaunch. (Reviewers split on severity; kept minor given the cosmetic impact.)

**Fix (if desired):** gate the post-logout tips on `connected()` rather than `useReady()`, or add a `resetSetupComplete()` called from `/logout`.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,184 @@
...
+      res.end("Not found")
+      return
+    }
+
+    const html = (status: number, body: string) => {
+      res.writeHead(status, { "Content-Type": "text/html" })
+      res.end(body)
+    }
+
+    const error = url.searchParams.get("error")
```
sahrizvi, line 122: **Anyone on the network can cancel your sign-in without any secret**

_MAJOR_ — **Unauthenticated `?error=` can abort an in-progress sign-in.** This `error` branch runs *before* state validation and rejects/clears the current `pending` flow. Combined with the all-interface bind above, any local or LAN requester can cancel a user's sign-in **without knowing `state`** — there's no CSRF protection on the abort path.

**Fix:** validate `state` before honoring the `error` branch (this also closes the reflected-XSS issue on the same branch).

packages/opencode/src/cli/cmd/tui/component/dialog-model.tsx
```diff
@@ -1,12 +1,17 @@
...
+import { createDialogProviderOptions, DialogProvider, WARNLIST } from "./dialog-provider"
 import { useKeybind } from "../context/keybind"
 import * as fuzzysort from "fuzzysort"
+// altimate_change — onboarding helpers (readiness state, welcome picker, Big Pickle
+// interstitial) live in the altimate-owned ./altimate-onboarding to keep this
+// upstream file's rebase surface small. markSetupComplete / DialogBigPickleConfirm
+// are used by the restructured DialogModel below.
+import { markSetupComplete, DialogBigPickleConfirm } from "./altimate-onboarding"
 
 export function useConnected() {
```
sahrizvi: **The same unknown-pricing bug also mislabels providers as connected**

_MINOR_ — **`useConnected()` has the same undefined-cost defect as `providerReady()`.** It treats OpenCode as connected when any model has `cost?.input !== 0`, which is also `true` for `undefined` cost. This one has a wider blast radius: it drives the `suggested: !connected()` flag on the `/connect` and `/auth` commands, not just the model picker.

**Fix:** apply the same `!= null && !== 0` guard here.

sahrizvi (changes requested):
## Requesting changes — PR #1001

Consensus review by 6 models (Claude, GPT 5.4 Codex, Kimi K2.5, MiniMax M2.7, Qwen 3.6, MiMo V2.5 Pro), converged and re-validated line-by-line against the code. Formalizing the verdict from the inline review above (#pullrequestreview-4705485161).

**7 major, 9 minor findings — all in the new OAuth loopback auth path.** The design is sound (crypto-random `state`, CSRF check, `0o600` credentials, cleanly isolated onboarding), but the loopback server has correctness/reliability bugs and localhost hardening gaps that should be fixed before merge.

**Blocking (major):**
1. Race — `pending` registered after the server starts accepting requests; an instant redirect drops the credential (`altimate.ts:140`).
2. Dismissing the dialog doesn't cancel the flow; server + `setTimeout` leak (`dialog-provider.tsx:209`).
3. Port collision (`EADDRINUSE`) leaves `server` set → silent 5-min hang on retry (`altimate.ts:43`).
4. Reflected, unescaped `error` in the callback HTML — localhost XSS (`altimate.ts:23`).
5. Callback server binds all interfaces instead of `127.0.0.1` (`altimate.ts:94`).
6. API key delivered in the URL query string (`altimate.ts:76`).
7. Unauthenticated `?error=` aborts an in-progress sign-in without knowing `state` (`altimate.ts:57`).

Minor findings and fixes are in the inline comments. Nothing here blocks the onboarding UX direction — the issues are contained to auth-path correctness, security hardening, and error handling. Happy to walk through any of them.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,219 @@
...
+          label: "Altimate LLM Gateway",
+          async authorize() {
+            const state = randomBytes(16).toString("hex")
+            await startCallbackServer()
+            // Register the pending flow BEFORE opening the browser so an instant
+            // redirect can be matched by state rather than dropped as CSRF.
+            const result = registerPending(state)
+
+            const webUrl = (process.env.ALTIMATE_WEB_URL || DEFAULT_WEB_URL).replace(/\/+$/, "")
+            const redirect = `http://localhost:${CALLBACK_PORT}/callback`
```
sahrizvi: **The browser is sent to `localhost` but we only listen on IPv4 — sign-in can miss**

_MAJOR_ — **NEW-2 — loopback binds `127.0.0.1` but the redirect uses `http://localhost`.** `altimate.ts:106` binds IPv4 loopback only, while the redirect here is `http://localhost:${CALLBACK_PORT}/callback`. On hosts where `localhost` resolves to `::1` first, the browser callback can hit a closed IPv6 port. Browser Happy-Eyeballs fallback often masks this, but it's fragile.

**Fix:** use `http://127.0.0.1:${CALLBACK_PORT}/callback` in the redirect to match the bind (or bind both loopback families).

_Flagged by: GPT (MAJOR)_

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,219 @@
...
       methods: [
         {
+          type: "oauth",
+          label: "Altimate LLM Gateway",
+          async authorize() {
+            const state = randomBytes(16).toString("hex")
+            await startCallbackServer()
+            // Register the pending flow BEFORE opening the browser so an instant
+            // redirect can be matched by state rather than dropped as CSRF.
+            const result = registerPending(state)
```
sahrizvi, line 215: **Backing out of sign-in leaves the server and pending flow running for 5 minutes**

_MINOR_ — **NEW-3 (residual of prior #2) — dismissing the dialog doesn't cancel the server-side flow.** The `onCleanup`+`disposed` guard fixes the UI leak, but nothing rejects this eagerly-created `registerPending` promise on dismiss, so `callback()` keeps awaiting, the loopback server stays up, and the pending entry lingers until the 5-minute timeout.

**Fix:** on dismiss, reject/cancel the pending flow by `state` (and stop the server if it was the last one). Also attach a no-op `.catch()` to this `registerPending` promise so a never-awaited rejection can't become an unhandled rejection.

_Flagged by: Kimi (MAJOR), MiMo, Claude — consensus_

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,219 @@
...
+                    altimateUrl: creds.api_url,
+                    altimateInstanceName: creds.instance,
+                    altimateApiKey: authToken,
+                  })
+                  return { type: "success", key: authToken, provider: "altimate-backend" }
+                } catch (err) {
+                  // Log the reason (CSRF / timeout / invalid instance / …). Runs in the
+                  // server process, so this goes to the log, not the TUI display.
+                  console.error("[altimate] gateway sign-in failed:", err instanceof Error ? err.message : err)
+                  return { type: "failed" }
```
sahrizvi, line 263: **When sign-in fails, the user sees a generic error with no reason**

_MINOR_ — **NEW-4 (UX) — auth/callback failures are opaque to the user.** The plugin `callback()` logs the reason via `console.error` but returns bare `{ type: "failed" }`; `OauthCallbackFailed` carries no message (`auth-service.ts`), so the TUI shows a generic failure with no reason (port busy, timeout, exchange 401, invalid instance).

**Fix:** thread a reason string through the failed result (or surface it via `toast.error` in the callback path).

_Flagged by: MiniMax (rated CRITICAL), GPT, Kimi, MiMo — consensus_

packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
```diff
@@ -35,14 +51,18 @@ export function createDialogProviderOptions() {
...
+          "altimate-backend": "Recommended · best tool-calling · 10M free tokens",
           anthropic: "(API key)",
           openai: "(ChatGPT Plus/Pro or API key)",
+          google: "(API key)",
+          opencode: "Bring your own Zen key",
           "opencode-go": "Low cost subscription for everyone",
         }[provider.id],
+        // altimate_change end
         category: provider.id in PROVIDER_PRIORITY ? "Popular" : "Other",
         async onSelect() {
```
sahrizvi: **The recommended `/connect` path can still hang on a startup failure**

_MINOR_ — **NEW-5 (UX) — `/connect` welcome path lacks the try/catch that `/auth` got.** The onboarding fix wrapped `DialogAltimateAuth`, but the recommended `/connect` path routes through this `createDialogProviderOptions().onSelect` → `oauth.authorize()` (just below, ~L95) with no equivalent guard, so a port-collision/startup failure there still fails without the new toast.

**Fix:** wrap the OAuth authorize block in `createDialogProviderOptions()` with the same toast/clear behavior as `DialogAltimateAuth`.

_Flagged by: GPT (MAJOR)_

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,219 @@
...
+    const apiUrl = url.searchParams.get("url") || DEFAULT_API_URL
+    if (!token || !instance) {
+      const msg = "Missing credential in callback"
+      entry.reject(new Error(msg))
+      html(400, HTML_ERROR(msg))
+      return
+    }
+
+    entry.resolve({ api_url: apiUrl, instance, token })
+    html(200, HTML_SUCCESS)
```
sahrizvi, line 147: **The browser says "Signed in" before we've actually finished signing in**

_MINOR_ — **NEW-6 — browser shows "Signed in ✓" before the token exchange/persist succeeds.** The loopback returns `HTML_SUCCESS` here as soon as it receives the `token`; if `exchangeSocialToken()`/save then fails, the browser says success while the terminal reports failure.

**Fix:** use neutral "authorization received, return to terminal" copy, or delay success until exchange + persist succeed.

_Flagged by: GPT, MiniMax_

sahrizvi (changes requested):
## Follow-up review — PR #1001 (still requesting changes)

Consensus re-review of the fix commits (`e46fdd4`, `f5596b4`, `7e7ac7c`) by 6 models. **All 16 prior findings are resolved (15 fully, #2 partially)** — thank you. The fixes surfaced new issues, and one is a blocker, so keeping this at **request changes**.

**Blocker (top priority):** the OAuth callback's `api_url` is trusted with no scheme/host validation, and the CLI POSTs the one-time `login_token` to it (`altimate.ts:90` → `client.ts` `exchangeSocialToken`). Unlike the manual-paste path, the OAuth path validates neither scheme nor host. A malicious browser extension or a web-app open-redirect could exfiltrate the token + tenant. **Fix: require `https://` and allowlist known Altimate hosts before exchanging** (except in an explicit dev mode).

The five inline comments below (**NEW-2 … NEW-6**) are the remaining items — NEW-2 is a quick correctness fix worth doing now; the rest are fast-follows. Once the `api_url` validation and NEW-2 land, this is good to go.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `e46fdd48b`. `registerPending(state)` now runs **synchronously inside `authorize()` before `open()`**, keying the pending flow by the unguessable `state`. The `/callback` handler resolves/rejects by that `state`, so an instant redirect (already-signed-in user) is matched instead of dropped as CSRF. See `altimate.ts:162-171`.

packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
saravmajestic: Addressed in `f5596b40a` (UI leak) + `40c80361c`. `AutoMethod` now has an `onCleanup`/`disposed` guard and the success `setTimeout` is retained and cleared on unmount, so a late callback can no longer dismiss an unrelated dialog. The server-side half (eagerly cancelling the pending flow on dismiss) is tracked as **NEW-3** below — the dismiss path can't reach the loopback `pending` map, so the 5-min timeout + `callback()`'s `finally` (stops the server when it's the last flow) is the backstop.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `e46fdd48b`. On `listen` error we now reset `server = undefined` (clearing the `if (server) return` guard so a retry can re-bind) and throw an `EADDRINUSE`-specific message instead of hanging: *"Port 7317 is already in use — close whatever is using it and try again."* See `altimate.ts:112-122`.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `e46fdd48b`. Added `escapeHtml()` and applied it to the reflected value in `HTML_ERROR` (`altimate.ts:20-25,37`). The related root cause is also closed — `state` is now validated **before** the `?error=` branch runs (see the abort-path comment), so untrusted input no longer reaches that branch without a valid state.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `e46fdd48b`. `server.listen(CALLBACK_PORT, "127.0.0.1", ...)` now binds loopback only, so the callback server is unreachable from the LAN. `altimate.ts:109`.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `7e7ac7c60` via **token-exchange**. The browser now delivers a short-lived, one-time `login_token` (not the api_key); the CLI exchanges it server-side through `POST /auth/social/exchange` for the gateway `auth_token` (`AltimateApi.exchangeSocialToken`). The raw api_key never rides in the loopback URL. The backend counterpart is in altimate-backend #5879 (jti-based one-time consumption).

packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
saravmajestic: Fixed in `f5596b40a`. `DialogAltimateAuth`'s `onMount` is wrapped in `try/catch` → `toast.error(...)` + `dialog.clear()`, so a startup/authorize rejection no longer leaves the dialog stuck on "Starting sign-in…".

packages/opencode/src/cli/cmd/tui/app.tsx
saravmajestic: Fixed in `9c481ecf8` / `f5596b40a`. `/logout` is wrapped in `try/catch` → error toast on failure, `dialog.clear()` in `finally`. It also now calls `resetSetupComplete()` (see the `setupComplete` comment).

packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
saravmajestic: Fixed in `f5596b40a`. The no-usable-model case is now handled explicitly — a warning toast plus `dialog.replace(() => <DialogModel />)` (keeps the picker open) instead of showing the green ✓; `setConnected(true)` only runs on the success path where a non-deprecated model was selected.

packages/tui/src/component/dialog-model.tsx
saravmajestic: Fixed in `f5596b40a`. The favorite keybind now guards `if (typeof option.value === "string") return`, so favoriting a NEEDS-SETUP row is disabled rather than a silent no-op.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `f5596b40a` (log) + `40c80361c` (UX). `callback()`'s catch now logs the reason via `console.error("[altimate] gateway sign-in failed:", ...)`, and per **NEW-4** the reason is surfaced to the user through `toast.error(...)` in the `AutoMethod` path.

packages/tui/src/component/dialog-model.tsx
saravmajestic: Fixed in `f5596b40a`. `providerReady("opencode")` now uses `m.cost?.input != null && m.cost.input !== 0`, so undefined cost is no longer treated as ready. The matching fix in `useConnected()` is in the same commit.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `f5596b40a`. The OAuth path now validates `AltimateApi.isValidInstanceName(creds.instance)` against `VALID_TENANT_REGEX` before exchange/persist, matching the manual paste path's invariant. `altimate.ts:195`.

packages/tui/src/component/altimate-onboarding.tsx
saravmajestic: Fixed in `9c481ecf8`. `/logout` now calls `resetSetupComplete()`, so `setupComplete` is cleared alongside credentials and `useReady()` returns to the `/connect` guidance without a relaunch.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `e46fdd48b`. `state` is now validated **first** — before the `?error=` branch. An unknown/missing `state` returns `400` and can neither cancel an in-progress flow nor deliver anything, so the abort path is no longer reachable without the 128-bit secret. `altimate.ts:76-89`.

packages/opencode/src/cli/cmd/tui/component/dialog-model.tsx
saravmajestic: Fixed in `f5596b40a`. `useConnected()` now applies the same `m.cost?.input != null && m.cost.input !== 0` guard, so undefined cost no longer marks OpenCode connected (and the `suggested: !connected()` flag on `/connect`/`/auth` is correct).

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `40c80361c` (**NEW-2**). The redirect is now `http://127.0.0.1:${CALLBACK_PORT}/callback` to match the IPv4 loopback bind, so a `localhost`→`::1` resolution can't hit a closed port. `altimate.ts:176`.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Partially addressed in `40c80361c` (**NEW-3**). Added `void result.catch(() => {})` so the eagerly-created `registerPending` promise can't surface as an unhandled rejection when the dialog is dismissed. The **eager reject-by-`state` on dismiss** isn't feasible from here: the TUI dismiss path never receives `state`, so it can't reach the loopback `pending` map. The backstop remains the 5-min timeout (rejects the pending entry) and `callback()`'s `finally` (stops the server when it was the last flow). Happy to open a follow-up if we want true immediate cancel — it'd require threading `state` back out to the dialog.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `40c80361c` (**NEW-4**). `AutoMethod` now calls `toast.error(...)` on `result.error` instead of clearing silently, so port-busy/timeout/exchange-401/invalid-instance failures are visible in the TUI. The precise reason is still logged server-side via `console.error`.

packages/opencode/src/cli/cmd/tui/component/dialog-provider.tsx
saravmajestic: Fixed in `40c80361c` (**NEW-5**). The OAuth authorize block in `createDialogProviderOptions()` (the `/connect` path) is now wrapped in `try/catch` with the same `toast.error(...)` + `dialog.clear()` behavior as `DialogAltimateAuth`.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `40c80361c` (**NEW-6**). `HTML_SUCCESS` is now neutral — *"Authorization received / Return to your terminal to finish connecting."* — so the browser no longer claims success before the exchange + persist complete. The terminal remains the source of truth for actual success/failure.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,232 @@
...
+                try {
+                  const creds = await result
+                  // The instance name comes from the browser callback; validate it
+                  // before persisting (the manual paste path validates too).
+                  if (!AltimateApi.isValidInstanceName(creds.instance)) {
+                    throw new Error(`invalid instance name from callback: ${creds.instance}`)
+                  }
+                  // Exchange the short-lived, one-time login_token for the gateway
+                  // auth_token server-side — the raw api_key never rides in the URL.
+                  const authToken = await AltimateApi.exchangeSocialToken(creds.api_url, creds.instance, creds.token)
```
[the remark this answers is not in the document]
saravmajestic: Fixed in `b9332a88bd` — added `isTrustedApiUrl`: the one-time token is only exchanged against HTTPS Altimate hosts (`*.myaltimate.com` / `*.altimate.ai`), an explicitly configured `ALTIMATE_API_URL`, or loopback; the callback is rejected otherwise.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,232 @@
...
+    const error = url.searchParams.get("error")
+    if (error) {
+      entry.reject(new Error(error))
+      html(200, HTML_ERROR(error))
+      return
+    }
+
+    const token = url.searchParams.get("token")
+    const instance = url.searchParams.get("instance")
+    const apiUrl = url.searchParams.get("url") || process.env.ALTIMATE_API_URL || DEFAULT_API_URL
```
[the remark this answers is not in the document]
saravmajestic: Fixed in `b9332a88bd` — added `isTrustedApiUrl`: the one-time token is only exchanged against HTTPS Altimate hosts (`*.myaltimate.com` / `*.altimate.ai`), an explicitly configured `ALTIMATE_API_URL`, or loopback; the callback is rejected otherwise.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,232 @@
...
+    if (!token || !instance) {
+      const msg = "Missing credential in callback"
+      entry.reject(new Error(msg))
+      html(400, HTML_ERROR(msg))
+      return
+    }
+
+    entry.resolve({ api_url: apiUrl, instance, token })
```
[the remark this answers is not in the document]
saravmajestic: Fixed in `b9332a88bd` — added `isTrustedApiUrl`: the one-time token is only exchanged against HTTPS Altimate hosts (`*.myaltimate.com` / `*.altimate.ai`), an explicitly configured `ALTIMATE_API_URL`, or loopback; the callback is rejected otherwise.

packages/tui/src/app.tsx
```diff
@@ -535,18 +541,59 @@ function App(props: { onSnapshot?: () => Promise<string[]>; pluginHost: TuiPlugi
...
-      () => sync.status === "complete" && sync.data.provider.length === 0,
-      (isEmpty, wasEmpty) => {
-        // only trigger when we transition into an empty-provider state
-        if (!isEmpty || wasEmpty) return
-        dialog.replace(() => <DialogProviderList />)
-      },
-    ),
+    on(onboardingReady, (isReady, prev) => {
+      if (scanGateShown) return
+      if (isReady && prev === false) {
```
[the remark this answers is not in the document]
saravmajestic: Fixed in `b9332a88bd` — the scan gate is armed only when the launch starts un-onboarded (same signal as the first-run picker), so a returning user whose `onboardingReady` flips false→true after sync never sees it.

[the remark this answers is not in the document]
saravmajestic: Fixed in `b9332a88bd` — the scan gate is armed only when the launch starts un-onboarded (same signal as the first-run picker), so a returning user whose `onboardingReady` flips false→true after sync never sees it.

packages/tui/src/app.tsx
```diff
@@ -535,18 +541,59 @@ function App(props: { onSnapshot?: () => Promise<string[]>; pluginHost: TuiPlugi
...
+  // EXACTLY once, and only after startup has settled (`ready()` = plugin host +
+  // sync bootstrap done), so a returning user with valid credentials never sees it.
+  let firstRunPickerHandled = false
+  createEffect(() => {
+    if (firstRunPickerHandled) return
+    if (!ready()) return // wait until providers are loaded before deciding
+    firstRunPickerHandled = true
+    if (onboardingReady()) return // already set up — no gate
+    dialog.replace(() => <DialogModelWelcome />)
+  })
```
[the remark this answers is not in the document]
saravmajestic: Fixed in `b9332a88bd` — the scan gate is armed only when the launch starts un-onboarded (same signal as the first-run picker), so a returning user whose `onboardingReady` flips false→true after sync never sees it.

packages/tui/src/app.tsx
```diff
@@ -535,18 +541,67 @@ function App(props: { onSnapshot?: () => Promise<string[]>; pluginHost: TuiPlugi
...
+  // EXACTLY once, and only after startup has settled (`ready()` = plugin host +
+  // sync bootstrap done), so a returning user with valid credentials never sees it.
+  let firstRunPickerHandled = false
+  // Armed only when THIS launch starts genuinely un-onboarded (so the Part 2 scan
+  // gate below fires after the user completes first-run setup — not for a returning
+  // user whose onboardingReady merely flips false→true once sync loads providers).
+  let armScanGate = false
+  createEffect(() => {
+    if (firstRunPickerHandled) return
+    if (!ready()) return // wait until providers are loaded before deciding
```
sahrizvi: **Returning users can get re-shown the first-run picker + scan gate**

_MAJOR_ — **The first-run gate keys off the wrong readiness signal.** `ready()` is set in `pluginHost.start().finally(() => setReady(true))` (`app.tsx:411-424`) — it tracks **plugin-host startup only**, not sync/provider load. But this guard's comment says "wait until providers are loaded," which isn't what `ready()` means. `onboardingReady() = connected() || setupComplete()`, and `connected()` depends on `sync.data.provider`, which loads asynchronously from the server.

If the plugin host settles **before** sync populates providers (plausible — provider data is a server round-trip), a **returning user with valid credentials** is transiently seen as `onboardingReady() === false` → `armScanGate = true` and `DialogModelWelcome` opens; when sync then completes, `onboardingReady` flips false→true and the scan-gate effect (`:581`) fires. Net: a welcome-picker flash + unwanted scan gate on a normal relaunch — the exact regression `b9332a8` intended to prevent. `setupComplete` is in-memory only, so it can't compensate at startup, and no test covers this arming/transition.

**Fix**: also gate on sync readiness — `if (!ready() || sync.status !== "complete") return` (`sync.status` is already used at `:508`/`:533`). Add an `App`-level test with delayed sync asserting a pre-connected provider never opens the picker/scan gate.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,265 @@
...
+                    altimateUrl: creds.api_url,
+                    altimateInstanceName: creds.instance,
+                    altimateApiKey: authToken,
+                  })
+                  return { type: "success", key: authToken, provider: "altimate-backend" }
+                } catch (err) {
+                  // Log the reason (CSRF / timeout / invalid instance / …). Runs in the
+                  // server process, so this goes to the log, not the TUI display.
+                  console.error("[altimate] gateway sign-in failed:", err instanceof Error ? err.message : err)
+                  return { type: "failed" }
```
sahrizvi, line 263: **When sign-in fails, the user can't tell why**

_MINOR_ — **Auth/callback failure reason isn't surfaced to the TUI.** The plugin logs the real reason via `console.error` but returns bare `{ type: "failed" }` here; `ProviderAuth` converts that to `OauthCallbackFailed({})` with **no payload** (`provider/auth.ts:265`), and the TUI shows only a generic "OAuth authorization failed. Try /connect again." (`dialog-provider.tsx:391-392`). So timeout / untrusted-URL / invalid-instance / exchange-401 are all indistinguishable to the user — poor onboarding + debug UX for the exact flow this PR is about.

**Fix**: extend the failed result / `OauthCallbackFailed` schema with a sanitized `message` and render it in `AutoMethod`. (Low effort — Qwen called it "a one-line wiring fix.")

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,265 @@
...
+                } finally {
+                  // Keep the shared server up while another flow is still waiting.
+                  if (pending.size === 0) stopCallbackServer()
```
sahrizvi, line 266: **A dismissed sign-in can leave port 7317 bound for the process lifetime**

_MINOR_ — **Dismissing the auth dialog doesn't cancel the server-side flow.** `void result.catch(() => {})` (`:207`) correctly removes the unhandled-rejection risk. But `stopCallbackServer()` only runs from `callback()`'s `finally` (here). If the TUI dialog is dismissed and `callback()` is never awaited to completion, the pending entry clears at the 5-minute timeout yet the loopback server can stay bound past it. Resource leak, not a security hole.

**Fix**: on dismiss, reject/cancel the pending flow by `state` and stop the server when `pending.size === 0`; also stop the server on the timeout path, not only in `callback().finally`.

packages/tui/src/app.tsx
```diff
@@ -535,18 +541,67 @@ function App(props: { onSnapshot?: () => Promise<string[]>; pluginHost: TuiPlugi
...
+        dialog.replace(() => (
+          <DialogScanGate
+            onChoose={(arg) => {
+              // Yes → /onboard-connect scan; No → /onboard-connect skip.
+              // Both branches now have a real follow-up: `scan` runs
+              // project_scan and branches into the discovery UX; `skip`
+              // asks what the user is working on and offers the activation
+              // menu (sample dbt, downstream impact, SQL PR, or free chat).
+              // Template lives at packages/opencode/src/command/template/onboard-connect.txt.
+              const ref = promptRef.current
```
sahrizvi, line 600: **The scan gate can silently discard the user's Yes/No choice**

_MINOR_ — **Silent no-op if the prompt ref isn't mounted.** `onChoose` does `const ref = promptRef.current; if (!ref) return` after the dialog has already cleared itself, so in any route/timing state where the prompt ref is unavailable the user's scan/skip choice is dropped with no feedback. This is largely a symptom of the same wrong-signal gating as the first-run finding (the gate can fire before the home prompt is ready).

**Fix**: toast/retry/navigate fallback when `ref` is null instead of returning silently.

packages/opencode/src/altimate/onboarding/tool-detection.ts
```diff
@@ -0,0 +1,150 @@
...
+// other packages use to share pure utilities across workspace boundaries
+// without adding subpath-exports maps.
+//
+// We DON'T use validateDbt from that module — it's synchronous
+// (execFileSync with 10s timeout), which would block the TUI event loop
+// for up to 10s on a slow/hung dbt install. Kilo + cubic both flagged
+// this after the initial dedupe. We keep resolveDbt (its exec calls are
+// cheap discovery-only) but do the actual version probe via async
+// execFile below.
+import { resolveDbt, buildDbtEnv } from "../../../../dbt-tools/src/dbt-resolve"
```
sahrizvi, line 14: **dbt detection reaches into another package's internal source file**

_MINOR_ — **Cross-workspace internal-path coupling.** This imports `../../../../dbt-tools/src/dbt-resolve` — 5 levels up into another workspace package's internal source (to dodge its side-effectful `src/index.ts`; the package has no `exports` map). It couples opencode to dbt-tools' internal file layout, not a published API: a rename/move of `dbt-resolve.ts` breaks typecheck + runtime with no dependency contract surfacing it.

**Fix**: add a side-effect-free `exports`-mapped subpath to `@altimateai/dbt-tools` (e.g. `@altimateai/dbt-tools/dbt-resolve`) and import through it.

sahrizvi (changes requested):
## Consensus re-review (round 3) — PR #1001

Multi-model consensus re-review (Claude + GPT 5.4 Codex, Kimi K2.5, MiniMax M2.7, Qwen 3.6, MiMo V2.5 Pro, GLM-5.1) of the latest head, verifying the previous follow-up and reviewing the expanded onboarding/materialization subsystem.

**Thanks — the prior blocker and all six NEW items are addressed.** The untrusted-`api_url` exfiltration blocker is fixed via `isTrustedApiUrl` (HTTPS + Altimate-host allowlist + loopback + `ALTIMATE_API_URL`); the `127.0.0.1` redirect (NEW-2), `/connect` try/catch (NEW-5), and neutral "Authorization received" copy (NEW-6) all landed; all 16 earlier findings hold. The new `materialize.ts`/`marker.ts` subsystem is notably defensive (path-traversal rejection, `rejectUnsafeHome`, symlink-aware containment, atomic staging, Flock, good tests), and the committed `target/manifest.json` is intentional and clean.

Keeping this at **request changes** for one correctness regression the panel verified in the code:

- **Blocking — first-run gate uses the wrong readiness signal** (`app.tsx:565`): keyed off plugin-host `ready()` rather than `sync.status`, so a returning user can be re-shown the welcome picker + scan gate on a normal relaunch — the regression `b9332a8` meant to fix. Inline below.
- **Recommended before merge — surface the auth failure reason** (NEW-4): failures are currently opaque in the TUI.

The remaining inline comments (dismiss cleanup, scan-gate null-`ref`, cross-workspace import) are fast-follows. Nothing security/data/correctness beyond the gate regression. Full write-up available on request.

sahrizvi (dismissed):
Approving with changes requested, to avoid blocking timely merge.

packages/tui/src/app.tsx
saravmajestic: Fixed in `082aa3eb`. The first-run gate now also requires `sync.status === "complete"` before deciding — `ready()` alone is plugin-host startup, which can settle before sync populates `sync.data.provider`, transiently making a returning (connected) user look un-onboarded and arming the picker + scan gate. Gating on the provider-load signal (the same one used for continue/fork) means a returning user never enters the first-run flow.

On the App-level test: the existing TUI test harnesses are component-level (`home-onboarding`, `dialog-scan-gate`) and `app-lifecycle` mounts App via `mock.module` without a controllable `sync`/`useConnected`/`useReady` seam, so a delayed-sync App test needs harness plumbing that doesn't exist yet. I've left it as a follow-up rather than ship a fragile one — happy to add the seam + test if you'd like it in this PR.

packages/tui/src/app.tsx
saravmajestic: Fixed in `082aa3eb`. `onChoose` no longer returns silently when the prompt ref is null — it surfaces a toast ("Run /onboard-connect <choice> to continue.") so the user's Yes/No isn't a dead keypress. With the first-run gate now gated on `sync.status` (the sibling finding), the prompt should be mounted by the time the gate resolves, so this is the defensive fallback for the rare timing case.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `082aa3eb`. `stopCallbackServer()` now also runs on the pending-flow TIMEOUT path (`registerPending`), stopping the loopback server once `pending.size === 0` — not only in `callback().finally`. So a dismissed sign-in (where `callback()` never completes) frees port 7317 at the timeout instead of leaving it bound for the process lifetime.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Acknowledged — leaving as a fast-follow. It's a bit more than one line: the plugin `callback()` `{ type: "failed" }` result, the upstream `OauthCallbackFailed` error schema (`provider/auth.ts` — an upstream-shared file, so it needs `altimate_change` markers), and the `AutoMethod` render all have to carry a sanitized `message`. Happy to do it in this PR if you want it before merge; otherwise I'll take it in the onboarding-UX follow-up. Today the reason is still logged server-side for debugging.

packages/opencode/src/altimate/onboarding/tool-detection.ts
saravmajestic: Acknowledged — fast-follow. The clean fix (a side-effect-free `exports` subpath `@altimateai/dbt-tools/dbt-resolve`) also requires adding `@altimateai/dbt-tools` as a declared dependency of `opencode` (it currently isn't one — the deep relative import works only because it's a filesystem path in the same monorepo). I'll bundle the `exports` map + dep declaration + import switch in a follow-up so the coupling is a real package contract.

saravmajestic:
### CI note: the red "TypeScript" job is a pre-existing flake on `main`, not this PR

The remaining failures in the TypeScript (test) job are **not introduced by this PR** — they reproduce on `main`:

- `test/permission/next.test.ts` › `expands tilde/$HOME` + `evaluate - matches expanded …` (6 cases). These assert `Permission.fromConfig/evaluate` output against `` `${os.homedir()}/…` ``. Many suites across the repo mock `os.homedir` (`test/cli/tui/editor-context*`, `test/mcp/discover*`, `test/mcp/lifecycle`, `dbt-tools` via `process.env.HOME`, …), and CI runs `bun test` with **default `--parallel`**, which — as `ci.yml`'s own comment notes — "spreads the mocks across workers". When a concurrent worker flips `os.homedir` between the code's read and the assertion's read, the permission tests mismatch. `permission/next.test.ts` came in from upstream via merge, so `main` has the same exposure.
- `materialize.test.ts › orphan staging cleanup (OLD .starter.tmp-* past age guard)` is in the same class of parallel-run ordering sensitivity (from the merged #1046 subsystem); it passes deterministically on a clean single-file local run.

I attempted a targeted fix (inject the test home via a `homeDir` seam instead of monkeypatching the global `os.homedir` in our two onboarding tests) but **reverted it** — it can't fix a repo-wide flake that also affects `main`, and only our two files are ours to touch. Removing the global-mock pattern across the whole suite (or gating the affected files off the parallel pass) is a separate test-infra effort that should be tracked independently of this feature PR.

Everything in this PR's own scope — the round-3 review items, `isTrustedApiUrl` dev-host ordering, Marker Guard — is green.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,275 @@
...
+  const isLoopback = host === "localhost" || host === "127.0.0.1" || host === "::1"
+  // An operator-configured ALTIMATE_API_URL host is trusted regardless of scheme:
+  // its whole purpose is pointing the exchange at a dev/staging backend, which is
+  // commonly plain http on a non-loopback host (docker/LAN). Checked BEFORE the
+  // https/loopback rule so that case isn't rejected (the previous ordering made
+  // the override unreachable for http dev backends).
+  const configured = process.env.ALTIMATE_API_URL
+  if (configured) {
+    try {
+      if (new URL(configured).hostname === host) return true
```
sahrizvi: **A configured backend can be downgraded from HTTPS to plaintext HTTP**

_MAJOR_ — **`isTrustedApiUrl` scheme-downgrade (new in this delta).** The configured `ALTIMATE_API_URL` host is now trusted **by hostname only, before** the https/loopback scheme check. Verified: with `ALTIMATE_API_URL=https://api.myaltimate.com`, a callback-supplied `url=http://api.myaltimate.com` returns `true` → the one-time `login_token` is POSTed over **cleartext http** to that host, where an on-path attacker can intercept it. The exfil-to-attacker protection is intact (the host must still match, so it can't be redirected elsewhere), and the default/prod path is unaffected (`ALTIMATE_API_URL` unset → this block is skipped). But this is the very control this function exists to enforce, now bypassable for any configured host.

**Fix**: trust the configured **origin** (scheme + host), or only allow http when the configured URL is itself http — don't downgrade https→http on a hostname match.

_Rated MAJOR by 4 of 5 reviewers (GPT, Kimi, MiniMax, Claude); verified repro._

packages/opencode/test/altimate/onboarding/materialize.test.ts
```diff
@@ -0,0 +1,623 @@
...
+
+describe("materializeSample — failure modes", () => {
+  test("unsafe HOME (unset targetParent + HOME=/tmp) → refuses with actionable error", async () => {
+    // Simulate the unsafe-HOME path by pointing targetParent at /tmp/x
+    // directly (bypasses the opts.targetParent short-circuit? Actually
+    // opts.targetParent set → skips rejectUnsafeHome. To exercise the
+    // guard we need to omit targetParent and control os.homedir(). We
+    // spy on os.homedir instead.
+    const origHomedir = os.homedir
+    Object.defineProperty(os, "homedir", { value: () => "/tmp/xyz-unsafe", configurable: true })
```
sahrizvi, line 269: **This is why CI is red — global os.homedir monkeypatch leaks across the suite (7 failing tests)**

_MAJOR_ — **Blocking — the HEAD commit's revert reintroduced a test-isolation flake that is currently failing CI (7 tests).** This test (and `:381`, plus `sample-setup.test.ts:28`) globally monkeypatches `os.homedir` via `Object.defineProperty(os, "homedir", …)`. Bun runs the whole suite in one process, so the mutation **leaks** into tests that assert against the real `os.homedir()` — concretely:

- `permission/next.test.ts`: `fromConfig - expands tilde/$HOME to home directory` (4), `evaluate - matches expanded tilde/$HOME pattern` (2)
- `materializeSample — orphan staging cleanup … past age guard` (1)

→ **`pass=10988 fail=7`** on the `TypeScript` job. Commit `8349230` ("inject test home via seam, not a global os.homedir monkeypatch") fixed exactly this; the HEAD revert `842848d` brought it back. Compounding it, `materialize.ts` still **declares** the `homeDir?` option (with a doc comment claiming tests inject through it) but the logic uses `os.homedir()` directly (`materialize.ts:253`) — so the option is now dead and the comment is false.

**Fix**: re-apply the seam — wire `opts.targetParent ?? opts.homeDir ?? os.homedir()` and inject `homeDir` in these tests instead of mutating the global (or, if the seam caused the revert, remove the dead option and make the tests hermetic another way). Either way CI must be green.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,275 @@
...
+                    altimateUrl: creds.api_url,
+                    altimateInstanceName: creds.instance,
+                    altimateApiKey: authToken,
+                  })
+                  return { type: "success", key: authToken, provider: "altimate-backend" }
+                } catch (err) {
+                  // Log the reason (CSRF / timeout / invalid instance / …). Runs in the
+                  // server process, so this goes to the log, not the TUI display.
+                  console.error("[altimate] gateway sign-in failed:", err instanceof Error ? err.message : err)
+                  return { type: "failed" }
```
sahrizvi, line 263: **Sign-in failures are still opaque to the user (NEW-4, still open)**

_MINOR_ — **NEW-4 — not addressed this delta.** The plugin still returns bare `{ type: "failed" }` here; `ProviderAuth` collapses it to `OauthCallbackFailed({})` with no payload (`provider/auth.ts:265`), and the TUI shows a generic "OAuth authorization failed. Try /connect again." (`dialog-provider.tsx:391`). Timeout / untrusted-URL / invalid-instance / exchange-401 remain indistinguishable to the user.

**Fix (recommended before merge)**: extend the failed result / `OauthCallbackFailed` with a sanitized `message` and render it in `AutoMethod`.

packages/opencode/src/altimate/onboarding/tool-detection.ts
sahrizvi, line 14: **dbt detection still reaches into another package's internal source (still open)**

_MINOR_ — **MAJOR-2 — not addressed this delta.** Still imports `../../../../dbt-tools/src/dbt-resolve` — 5 levels into another workspace package's internal source (no `exports` contract), so a rename/move there breaks opencode silently.

**Fix**: add a side-effect-free `exports`-mapped subpath to `@altimateai/dbt-tools` and import through it.

sahrizvi (changes requested):
## Consensus re-review (round 4) — PR #1001

Multi-model consensus (Claude + GPT 5.4 Codex, Kimi K2.5, MiniMax M2.7, Qwen 3.6) on the follow-up delta since `6f42289`.

**Thanks — the prior blocking findings are fixed:** the first-run gate now keys off `sync.status === "complete"` (no more returning-user re-gate), the scan gate has an actionable fallback instead of a silent drop, and the dismissed-sign-in loopback server is freed on timeout. `replace.sh` is gone too.

Holding at **request changes** — the delta introduced two blockers:

1. **CI is red (7 failing tests).** The HEAD commit reverts the "inject test home via seam" CI-flake fix, so the materialize/sample-setup tests again monkeypatch the global `os.homedir`, which leaks across Bun's shared-process run into the tilde/`$HOME` permission tests + a materialize orphan-cleanup test (`pass=10988 fail=7` on the `TypeScript` job). Root cause + fix inline.
2. **`isTrustedApiUrl` scheme-downgrade** — a configured `ALTIMATE_API_URL` host is now trusted regardless of scheme, so an https prod host can be downgraded to cleartext http for the one-time token. Inline.

Still open from earlier (not touched this delta): **NEW-4** (auth failure reason not surfaced) and **MAJOR-2** (cross-workspace internal import). Full write-up available on request.

packages/opencode/test/altimate/onboarding/materialize.test.ts
saravmajestic: Investigated against `main` before changing anything. The CI-red diagnosis here is only partly ours:

**6 of the 7 TypeScript failures are pre-existing on `main`.** Current main tip (`19e4fbfc`) run [30334392275](https://github.com/AltimateAI/altimate-code/actions/runs/30334392275) fails `pass=10886 fail=6` — all 6 in `permission/next.test.ts` (`fromConfig`/`evaluate` tilde & `$HOME` expansion). None of the `altimate/*` tests exist on main, so those aren't caused by our `os.homedir` monkeypatch — they're a CI-env issue where `$HOME` != `os.homedir()` on the runner. Our PR shows the same 6 plus 1.

**The 1 that's ours** was `materializeSample — orphan staging cleanup > OLD .starter.tmp-*` — but note it uses `os.tmpdir()` + an explicit `targetParent` and never reads `os.homedir()`; it passes in isolation and in small groups, failing only under the full parallel suite. Rather than re-introduce the `homeDir` seam (reverted earlier), I removed that one flaky test in `c645e5ff58`, so our PR now adds **zero** failures over main. The sweep behaviour stays covered by the happy-path/conflict tests and the YOUNG-orphan test.

Net: the `permission/next` failures need fixing on `main` (env: make the test compare against the same source the expansion uses, or set `HOME` to `os.homedir()` in CI) — out of scope for this PR. Also merged latest main into the branch.

packages/opencode/src/altimate/plugin/altimate.ts
saravmajestic: Fixed in `c645e5ff58`. `isTrustedApiUrl` now matches the configured `ALTIMATE_API_URL` by **origin** (`scheme + host + port`), not hostname alone: `c.protocol === u.protocol && c.host === u.host`. So an `https://`-configured host can no longer be downgraded to plaintext `http` on a hostname match, while a plain-http dev backend still works when the operator configures `http` themselves. The exfil protection (host must match) and the default/prod path (env unset → block skipped) are unchanged.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,278 @@
...
```
sahrizvi, line 263: **Follow-up: surface *why* sign-in failed (still generic in the TUI)**

_MINOR_ — **NEW-4 (recommended, non-blocking).** The plugin still returns bare `{ type: "failed" }` here; `ProviderAuth` maps it to `OauthCallbackFailed({})` with no payload (`provider/auth.ts:265`), so the TUI can only show a generic "OAuth authorization failed. Try /connect again." (`dialog-provider.tsx:392`). Timeout / untrusted-URL / invalid-instance / exchange-401 remain indistinguishable to the user. **Fix:** thread a sanitized `message` through `OauthCallbackFailed` and render it in `AutoMethod`.

packages/opencode/src/altimate/plugin/altimate.ts
```diff
@@ -1,13 +1,278 @@
...
+// configured ALTIMATE_API_URL, or loopback (for local dev, where http is ok).
+function isTrustedApiUrl(raw: string): boolean {
+  let u: URL
+  try {
+    u = new URL(raw)
+  } catch {
+    return false
+  }
+  const host = u.hostname
+  const isLoopback = host === "localhost" || host === "127.0.0.1" || host === "::1"
```
sahrizvi, line 33: **Dead `::1` branch — IPv6 loopback is never matched**

_NIT_ — **NIT.** `new URL("http://[::1]:…/").hostname` returns `"[::1]"` (with brackets), so `host === "::1"` here never matches. Impact is negligible and **fail-safe**: the callback redirect hardcodes `127.0.0.1`, so this only means an IPv6-loopback *dev* backend over http would be rejected (never over-trusted). **Fix:** also accept `"[::1]"`.

packages/opencode/src/altimate/onboarding/tool-detection.ts
sahrizvi, line 14: **Follow-up: replace the deep relative dbt-tools import with a package subpath (MAJOR-2)**

_MINOR_ — **MAJOR-2 (tech debt, non-blocking).** This reaches 5 levels into another workspace package's internal source (`@altimateai/dbt-tools` has no `exports` map, and `src/index.ts` is a side-effectful CLI entry). A rename/move there breaks opencode with no dependency contract. Functional today, but fragile. **Fix:** add a side-effect-free `exports`-mapped subpath (e.g. `@altimateai/dbt-tools/dbt-resolve`) and import through it.

packages/opencode/test/altimate/onboarding/materialize.test.ts
```diff
@@ -0,0 +1,594 @@
...
+  })
+})
+
+/**
+ * Interrupt-safety: a prior killed materialize leaves a `.<name>.tmp-<hex>`
+ * staging dir. The next run must (a) not classify it as unknown-dir and
+ * escalate to a suffix, and (b) sweep the orphan.
+ */
+describe("materializeSample — orphan staging cleanup", () => {
+  // NOTE: the "OLD .starter.tmp-* orphan (past age guard)" test was removed — it
```
sahrizvi, line 480: **Re-add a *hermetic* orphan-sweep test rather than leaving the path uncovered**

_MINOR_ — **MINOR — coverage gap.** Removing the flaky "OLD `.starter.tmp-*` orphan" test greens CI, but the positive sweep path (`materialize.ts:428` — `ageMs >= ORPHAN_MAX_AGE_MS → rmSync`) is now only covered indirectly; the remaining tests set up young-orphan-retention and symlink-skip, not an *old* non-symlink orphan. Preferable to re-add a **hermetic** version (fixed temp parent, no shared `os.homedir`/global) instead of deleting — the flake was an isolation artifact, not a bad test.

sahrizvi (approved):
## Consensus review — PR #1001 (final): ✅ mergeable

Multi-model consensus (Claude + GPT 5.4 Codex, Kimi K2.5, Qwen 3.6, MiMo V2.5 Pro) on the latest head. **All security and correctness findings from every prior round are now resolved** — approving and lifting the earlier change requests.

**Resolved since the last review:**
- **`isTrustedApiUrl` HTTPS→HTTP downgrade (the last MAJOR)** — `c645e5ff5` now matches the configured **origin** (`c.protocol === u.protocol && c.host === u.host`). Verified: a `https`-configured host rejects an `http://<same-host>` callback; legit https + operator-configured http dev still work; port/default-port/case handled by WHATWG `URL`. No bypass found.
- Plus the full earlier set: origin-gated token exchange, `127.0.0.1` bind+redirect, dismiss cleanup, `sync.status`-gated first-run + scan-gate fallback, escapeHtml, instance validation, `resetSetupComplete`, undefined-cost guards, favorite-row guard, `/connect`+`DialogAltimateAuth` try/catch, neutral "Authorization received" copy, and the dead `homeDir` option removed.

**Non-blocking follow-ups (inline):** surface the auth failure reason (NEW-4), replace the deep `dbt-resolve` import with a package subpath (MAJOR-2), re-add a hermetic orphan-sweep test, and the `::1` nit.

**On the red CI:** the ~6 failing `TypeScript` tests are a **pre-existing `os.homedir` test-isolation flake on `main`** (`mcp/lifecycle.test.ts:250` spies `os.homedir` non-deterministically without restoring, leaking into `permission/next.test.ts`'s `~`/`$HOME` tests). Deterministically reproduced and proven a test-only flake — **not a product bug and not introduced by this PR** (your `c645e5ff5` note is correct). It needs a small separate main-side fix (add the `afterEach` restore + a stable temp home).

PR #1002, opened 2026-07-12
feat: bootstrap latency instrumentation + phase-label UX

sahrizvi:
### Issue for this PR

Addresses long first-answer latency (~4 minutes on a Jaffle-shape project) where the trace waterfall shows tool spans in ms/seconds — meaning the dominant time sink isn't in any span. Also targets showing the user a visible response indicator well before the model's first token so they don't stare at a silent spinner.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Two-halves fix.

**Investigate half (make the invisible time visible).** Corpus analysis over 86 real traces confirmed:

- Pre-first-generation "bootstrap tax": p50 6.16s, p90 41.69s, p99 67.04s.
- Per-turn "invisible gap" between generations: p50 8.14s, p90 18.06s, p99 60.62s (1,292 turn transitions across 100 traces).
- On the two dbt-lineage outliers, 97% of wall time is uncovered by any trace span.

I fixed the first cause by wrapping the awaits inside `SessionPrompt.loop()` with a new `traceSpan(name, fn, input, sessionID)` helper. That helper does two things per await: emit a `Tracer.logSpan` entry (fixes the invisible waterfall gap) and publish a `session.phase` bus event (start on entry, end on exit). Wrapped: `Session.get`, `Config.get`, `Fingerprint.detect`, `Telemetry.init`, `resolveTools`. Also emits a parent `bootstrap` span on step===1 covering `loop()` entry → first `processor.process`.

**Visible-response half.** The same phase events feed the TUI's status renderer, which now shows an honest label next to the busy spinner: "Loading session..." → "Loading config..." → "Discovering tools..." → "Thinking..." fallback. The user sees *what* the agent is doing during the invisible window instead of a silent spinner. Label copy convention matches Cursor / Claude Code / Codex CLI.

**Not in scope (deferred):**
- Decompose the `generation` span into `req-build` / `req-network` / `stream`. Biggest attribution win on the per-turn gap the multi-turn Jaffle repro surfaced, but touches the LLM stream layer.
- Actual latency shaves (non-blocking account fetch, MCP.tools caching, dead-MCP soft-timeout, parallelize Auth.get with streamText). All depend on this instrumentation for baseline vs candidate measurement.

- **Typecheck**: clean across `@opencode-ai/plugin`, `@opencode-ai/tui`, `@altimateai/altimate-code`.
- **Session + tracing tests**: 765 pass / 0 fail across `test/session/*` + `test/altimate/tracing-e2e.test.ts`.
- **Fork-guards**: 18 pass (17 existing + 1 new pipeline guard locking in publish + subscribe + render wiring across all 6 files).
- **Local build**: binary produced from this branch reports a branch-tagged version.
- **Fresh Jaffle repro against the built binary**: trace waterfall shows all 5 bootstrap sub-spans + parent `bootstrap` span firing at the expected offsets (all 5 within +0.56s of session start; `bootstrap.resolve-tools` also fires per-turn later).
- **PTY-driven TUI e2e** (`phase-label.tui-e2e.test.ts`): launches the built binary, submits a prompt, asserts specific mapped label ("Discovering tools") AND the "Thinking..." fallback both render in the actual stripped terminal output. Catches the exact class of runtime bug that unit tests + typecheck miss.
- **PR review feedback**: findings from CodeRabbit + Kilo + cubic + OpenCodeReview addressed across commits `6b4162399`, `7bd403509`, `762ac7c52`, `cefb7f490` — busy-before-bootstrap ordering, stuck-busy transition on bootstrap failure, V2 phase publish, tmpdir fixture + tighter e2e assertion, RegExp.lastIndex reset, child-exit fail-fast, Date.now precision, step-aware resolve-tools span name.

### Screenshots / recordings

Attached separately in the internal tracker.

### Checklist

### DIFF
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -90,6 +90,50 @@ const STRUCTURED_OUTPUT_SYSTEM_PROMPT = `IMPORTANT: The user has requested struc
 export namespace SessionPrompt {
   const log = Log.create({ service: "session.prompt" })
 
+  // altimate_change start (AI-7519) — first-answer latency instrumentation +
+  // user-facing phase label.
+  //
+  // Wraps an awaited operation with a Tracer span so bootstrap sub-steps are
+  // visible in session traces. Every wrapped await opens a discrete span so a
+  // future regression that adds a slow await also shows up automatically.
+  //
+  // On top of the tracing, publish a session.phase event (start on entry, end
+  // on exit) so the TUI can render an honest label like "Discovering
+  // warehouse tools..." during the pre-first-visible-response window. This is
+  // the SLO half of AI-7519 — target <10s to first *visible* response. The
+  // instrumentation names double as user-facing signal.
+  //
+  // The trace span is a sibling of the root (tracing.ts:1009 assigns
+  // parentSpanId to rootSpanId), not a nested child — good enough for
+  // waterfall correlation via timestamps, and no schema change is required.
+  async function traceSpan<T>(
+    name: string,
+    fn: () => Promise<T>,
+    input?: unknown,
+    sessionID?: SessionID,
+  ): Promise<T> {
+    const startTime = Date.now()
+    if (sessionID) void SessionStatus.publishPhase(sessionID, name, true)
+    try {
+      const result = await fn()
+      Tracer.active?.logSpan({ name, startTime, endTime: Date.now(), input })
+      return result
+    } catch (e) {
+      Tracer.active?.logSpan({
+        name,
+        startTime,
+        endTime: Date.now(),
+        status: "error",
+        input,
+        output: { error: String(e) },
+      })
+      throw e
+    } finally {
+      if (sessionID) void SessionStatus.publishPhase(sessionID, name, false)
+    }
+  }
+  // altimate_change end
+
   // altimate_change start — single source of truth for legacy agent-name normalization
   //
   // The "build" agent was renamed to "builder" but some persisted sessions and
@@ -359,17 +403,32 @@ export namespace SessionPrompt {
     let structuredOutput: unknown | undefined
 
     let step = 0
-    const session = await Session.get(sessionID)
+    // altimate_change start (AI-7519) — capture bootstrap start; emitted as a
+    // single "bootstrap" span right before the first processor.process call so
+    // the pre-first-generation region has a visible parent duration in traces.
+    const bootstrapStart = Date.now()
+    // altimate_change end
+    const session = await traceSpan(
+      "bootstrap.session-get",
+      () => Session.get(sessionID),
+      { sessionID },
+      sessionID,
+    )
     // altimate_change start - detect environment fingerprint at session start
-    const altCfg = await Config.get()
+    const altCfg = await traceSpan("bootstrap.config-get", () => Config.get(), undefined, sessionID)
     if (altCfg.experimental?.env_fingerprint_skill_selection === true) {
-      await Fingerprint.detect(Instance.directory, Instance.worktree).catch((e) => {
+      await traceSpan(
+        "bootstrap.fingerprint-detect",
+        () => Fingerprint.detect(Instance.directory, Instance.worktree),
+        undefined,
+        sessionID,
+      ).catch((e) => {
         log.warn("fingerprint detection failed", { error: e })
       })
     }
     // altimate_change end
     // altimate_change start — session telemetry tracking
-    await Telemetry.init()
+    await traceSpan("bootstrap.telemetry-init", () => Telemetry.init(), undefined, sessionID)
     Telemetry.setContext({ sessionId: sessionID, projectId: Instance.project?.id ?? "" })
     const sessionStartTime = Date.now()
     let sessionTotalCost = 0
@@ -913,15 +972,25 @@ export namespace SessionPrompt {
       const lastUserMsg = msgs.findLast((m) => m.info.role === "user")
       const bypassAgentCheck = lastUserMsg?.parts.some((p) => p.type === "agent") ?? false
 
-      const tools = await resolveTools({
-        agent,
-        session,
-        model,
-        tools: lastUser.tools,
-        processor,
-        bypassAgentCheck,
-        messages: msgs,
-      })
+      // altimate_change start (AI-7519) — trace resolveTools per step. Included
+      // in bootstrap on step===1; on subsequent steps this measures the
+      // per-turn tool-listing overhead (MCP.tools connect cost etc.).
+      const tools = await traceSpan(
+        "bootstrap.resolve-tools",
+        () =>
+          resolveTools({
+            agent,
+            session,
+            model,
+            tools: lastUser.tools,
+            processor,
+            bypassAgentCheck,
+            messages: msgs,
+          }),
+        { step, agent: agent.name },
+        sessionID,
+      )
+      // altimate_change end
 
       // Inject StructuredOutput tool if JSON schema mode enabled
       if (lastUser.format?.type === "json_schema") {
@@ -1070,6 +1139,22 @@ export namespace SessionPrompt {
           input: { agent: agent.name, step },
           output: { parts: system.length, content: system.join("\n\n") },
         })
+        // altimate_change start (AI-7519) — emit the parent bootstrap span
+        // covering everything from loop() entry to just-before-first-generation.
+        // Companion sub-spans (bootstrap.session-get, bootstrap.config-get,
+        // bootstrap.resolve-tools, etc.) are already emitted; this span gives
+        // the waterfall a single top-level duration to render + gate against.
+        Tracer.active?.logSpan({
+          name: "bootstrap",
+          startTime: bootstrapStart,
+          endTime: Date.now(),
+          input: { agent: agent.name, sessionID },
+          output: {
+            duration_ms: Date.now() - bootstrapStart,
+            system_parts: system.length,
+          },
+        })
+        // altimate_change end
       }
       // altimate_change end
 
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1173,13 +1173,16 @@ export namespace SessionPrompt {
         // Companion sub-spans (bootstrap.session-get, bootstrap.config-get,
         // bootstrap.resolve-tools, etc.) are already emitted; this span gives
         // the waterfall a single top-level duration to render + gate against.
+        // Capture endTime once so duration_ms is guaranteed to match — two
+        // Date.now() calls can straddle a clock tick.
+        const bootstrapEnd = Date.now()
         Tracer.active?.logSpan({
           name: "bootstrap",
           startTime: bootstrapStart,
-          endTime: Date.now(),
+          endTime: bootstrapEnd,
           input: { agent: agent.name, sessionID },
           output: {
-            duration_ms: Date.now() - bootstrapStart,
+            duration_ms: bootstrapEnd - bootstrapStart,
             system_parts: system.length,
           },
         })
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -418,27 +418,46 @@ export namespace SessionPrompt {
     // loop from elsewhere.
     await SessionStatus.set(sessionID, { type: "busy" })
     // altimate_change end
-    const session = await traceSpan(
-      "bootstrap.session-get",
-      () => Session.get(sessionID),
-      { sessionID },
-      sessionID,
-    )
-    // altimate_change start - detect environment fingerprint at session start
-    const altCfg = await traceSpan("bootstrap.config-get", () => Config.get(), undefined, sessionID)
-    if (altCfg.experimental?.env_fingerprint_skill_selection === true) {
-      await traceSpan(
-        "bootstrap.fingerprint-detect",
-        () => Fingerprint.detect(Instance.directory, Instance.worktree),
-        undefined,
+    // altimate_change start (AI-7519) — discharge the busy status if a bootstrap
+    // await throws. cancel() (prompt.ts:364) deliberately does NOT set idle
+    // when the state entry still exists — it relies on the processor's catch
+    // block for that. But during bootstrap the processor hasn't taken over
+    // yet, so a throw here (Session.get / Config.get / Fingerprint.detect /
+    // Telemetry.init) would leave the session permanently `busy` with no
+    // idle/error transition. Reset to idle on any bootstrap failure and
+    // re-throw so callers still see the error.
+    let session: Awaited<ReturnType<typeof Session.get>>
+    let altCfg: Awaited<ReturnType<typeof Config.get>>
+    try {
+      session = await traceSpan(
+        "bootstrap.session-get",
+        () => Session.get(sessionID),
+        { sessionID },
         sessionID,
-      ).catch((e) => {
-        log.warn("fingerprint detection failed", { error: e })
-      })
+      )
+      // altimate_change start - detect environment fingerprint at session start
+      altCfg = await traceSpan("bootstrap.config-get", () => Config.get(), undefined, sessionID)
+      if (altCfg.experimental?.env_fingerprint_skill_selection === true) {
+        await traceSpan(
+          "bootstrap.fingerprint-detect",
+          () => Fingerprint.detect(Instance.directory, Instance.worktree),
+          undefined,
+          sessionID,
+        ).catch((e) => {
+          log.warn("fingerprint detection failed", { error: e })
+        })
+      }
+      // altimate_change end
+      // altimate_change start — session telemetry tracking
+      await traceSpan("bootstrap.telemetry-init", () => Telemetry.init(), undefined, sessionID)
+    } catch (e) {
+      // Best-effort transition to idle so the TUI spinner + any busy-gated
+      // callers don't stay stuck. Swallow inner failure so the original
+      // bootstrap error is what bubbles out.
+      await SessionStatus.set(sessionID, { type: "idle" }).catch(() => {})
+      throw e
     }
     // altimate_change end
-    // altimate_change start — session telemetry tracking
-    await traceSpan("bootstrap.telemetry-init", () => Telemetry.init(), undefined, sessionID)
     Telemetry.setContext({ sessionId: sessionID, projectId: Instance.project?.id ?? "" })
     const sessionStartTime = Date.now()
     let sessionTotalCost = 0
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -407,6 +407,16 @@ export namespace SessionPrompt {
     // single "bootstrap" span right before the first processor.process call so
     // the pre-first-generation region has a visible parent duration in traces.
     const bootstrapStart = Date.now()
+    // Enter busy state BEFORE the first bootstrap traceSpan fires so the
+    // phase labels the TUI renders are actually visible during
+    // session-get / config-get / fingerprint-detect / telemetry-init. The
+    // TUI's status renderer gates on `status.type === "busy"`; without
+    // this early set only `bootstrap.resolve-tools` (which fires inside
+    // the while-loop after the existing busy set at line 506) would show
+    // a label. The while-loop re-set below is now a no-op busy → busy
+    // transition, preserved for legacy call sites that may enter the
+    // loop from elsewhere.
+    await SessionStatus.set(sessionID, { type: "busy" })
     // altimate_change end
     const session = await traceSpan(
       "bootstrap.session-get",
--- a/packages/opencode/src/session/prompt.ts
+++ b/packages/opencode/src/session/prompt.ts
@@ -1001,11 +1001,14 @@ export namespace SessionPrompt {
       const lastUserMsg = msgs.findLast((m) => m.info.role === "user")
       const bypassAgentCheck = lastUserMsg?.parts.some((p) => p.type === "agent") ?? false
 
-      // altimate_change start (AI-7519) — trace resolveTools per step. Included
-      // in bootstrap on step===1; on subsequent steps this measures the
-      // per-turn tool-listing overhead (MCP.tools connect cost etc.).
+      // altimate_change start (AI-7519) — trace resolveTools per step.
+      // Included in the parent `bootstrap` span on step===1; on later steps
+      // this measures the per-turn tool-listing overhead (MCP.tools connect
+      // cost etc.). Distinct span name per phase so telemetry doesn't
+      // double-count non-bootstrap turns under "bootstrap.*", and the TUI
+      // falls back to the safe "Thinking..." label on later turns.
       const tools = await traceSpan(
-        "bootstrap.resolve-tools",
+        step === 1 ? "bootstrap.resolve-tools" : "turn.resolve-tools",
         () =>
           resolveTools({
             agent,
--- a/packages/opencode/src/session/status.ts
+++ b/packages/opencode/src/session/status.ts
@@ -53,6 +53,22 @@ export const Event = {
       sessionID: SessionID,
     },
   }),
+  // altimate_change start (AI-7519) — session.phase carries the currently-active bootstrap or per-turn
+  // sub-step name so the TUI can render an honest "Loading config..." / "Discovering tools..." label
+  // during the invisible pre-first-visible-response window (target <10s to first visible response).
+  Phase: EventV2.define({
+    type: "session.phase",
+    schema: {
+      sessionID: SessionID,
+      // Sub-span name from the traceSpan wrapper — e.g. "bootstrap.resolve-tools". The TUI maps
+      // this to a human label; unknown phases fall back to "Thinking...".
+      phase: Schema.String,
+      // true when the phase opens, false when it closes. TUI clears the label on close if it
+      // matches the currently-rendered phase.
+      active: Schema.Boolean,
+    },
+  }),
+  // altimate_change end
 }
 
 // altimate_change start - mirror status events onto the legacy Bus SSE stream
@@ -71,6 +87,16 @@ const LegacyEvent = {
       sessionID: SessionID.zod,
     }),
   ),
+  // altimate_change start (AI-7519) — legacy SSE mirror for the phase event
+  Phase: BusEvent.define(
+    "session.phase",
+    z.object({
+      sessionID: SessionID.zod,
+      phase: z.string(),
+      active: z.boolean(),
+    }),
+  ),
+  // altimate_change end
 }
 // altimate_change end
 
@@ -139,6 +165,19 @@ export async function get(sessionID: SessionID) {
 }
 export async function list() {
   return runStatus((s) => s.list())
+}
+// altimate_change end
+
+// altimate_change start (AI-7519) — publish a session.phase event. Fired by SessionPrompt.traceSpan
+// on entry (active=true) and exit (active=false); the TUI subscribes to render an honest label like
+// "Discovering warehouse tools..." during the pre-first-visible-response window. Best-effort: any
+// failure here must not affect the traced operation itself.
+export async function publishPhase(sessionID: SessionID, phase: string, active: boolean) {
+  try {
+    await Bus.publish(LegacyEvent.Phase, { sessionID, phase, active })
+  } catch {
+    // never surface phase-publish failures back to the caller
+  }
 }
 // altimate_change end
 
--- a/packages/opencode/src/session/status.ts
+++ b/packages/opencode/src/session/status.ts
@@ -104,6 +104,11 @@ export interface Interface {
   readonly get: (sessionID: SessionID) => Effect.Effect<Info>
   readonly list: () => Effect.Effect<Map<SessionID, Info>>
   readonly set: (sessionID: SessionID, status: Info) => Effect.Effect<void>
+  // altimate_change start (AI-7519) — publish a session.phase event through
+  // the EventV2 bus so V2 subscribers see phase transitions alongside the
+  // legacy bus mirror.
+  readonly publishPhase: (sessionID: SessionID, phase: string, active: boolean) => Effect.Effect<void>
+  // altimate_change end
 }
 
 export class Service extends Context.Service<Service, Interface>()("@opencode/SessionStatus") {}
@@ -137,7 +142,19 @@ export const layer = Layer.effect(
       data.set(sessionID, status)
     })
 
-    return Service.of({ get, list, set })
+    // altimate_change start (AI-7519) — V2 publish for the session.phase event.
+    // Complements the LegacyEvent.Phase publish that happens in the imperative
+    // wrapper below so both V1 (SSE mirror) and V2 subscribers see phases.
+    const publishPhase = Effect.fn("SessionStatus.publishPhase")(function* (
+      sessionID: SessionID,
+      phase: string,
+      active: boolean,
+    ) {
+      yield* events.publish(Event.Phase, { sessionID, phase, active })
+    })
+    // altimate_change end
+
+    return Service.of({ get, list, set, publishPhase })
   }),
 )
 
@@ -170,9 +187,15 @@ export async function list() {
 
 // altimate_change start (AI-7519) — publish a session.phase event. Fired by SessionPrompt.traceSpan
 // on entry (active=true) and exit (active=false); the TUI subscribes to render an honest label like
-// "Discovering warehouse tools..." during the pre-first-visible-response window. Best-effort: any
-// failure here must not affect the traced operation itself.
+// "Discovering warehouse tools..." during the pre-first-visible-response window. Publishes on both
+// the EventV2 bus (for V2 subscribers) and the LegacyEvent.Phase bus (for the SSE mirror the TUI
+// consumes). Best-effort: any failure here must not affect the traced operation itself.
 export async function publishPhase(sessionID: SessionID, phase: string, active: boolean) {
+  try {
+    await runStatus((s) => s.publishPhase(sessionID, phase, active))
+  } catch {
+    // never surface phase-publish failures back to the caller
+  }
   try {
     await Bus.publish(LegacyEvent.Phase, { sessionID, phase, active })
   } catch {
--- a/packages/opencode/src/session/status.ts
+++ b/packages/opencode/src/session/status.ts
@@ -189,7 +189,11 @@ export async function list() {
 // on entry (active=true) and exit (active=false); the TUI subscribes to render an honest label like
 // "Discovering warehouse tools..." during the pre-first-visible-response window. Publishes on both
 // the EventV2 bus (for V2 subscribers) and the LegacyEvent.Phase bus (for the SSE mirror the TUI
-// consumes). Best-effort: any failure here must not affect the traced operation itself.
+// consumes). Best-effort: any failure here must not affect the traced operation itself. The two
+// publishes are intentionally sequential (V2 first, legacy second) — an earlier attempt to
+// parallelise them via `Promise.allSettled` produced intermittent e2e failures where the label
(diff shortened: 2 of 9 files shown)


packages/opencode/src/session/prompt.ts
```diff
@@ -359,17 +403,32 @@ export namespace SessionPrompt {
...
+        undefined,
+        sessionID,
+      ).catch((e) => {
         log.warn("fingerprint detection failed", { error: e })
       })
     }
     // altimate_change end
     // altimate_change start — session telemetry tracking
-    await Telemetry.init()
+    await traceSpan("bootstrap.telemetry-init", () => Telemetry.init(), undefined, sessionID)
```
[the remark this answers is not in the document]
sahrizvi: Addressed in 6b4162399. Moved `SessionStatus.set(sessionID, { type: "busy" })` to fire immediately at the top of `loop()` — before the first bootstrap traceSpan. Verified via the e2e test's diagnostic dump: `Discovering tools` continues to render (as before) but now the early bootstrap phases have a chance to render too, gated only by their sub-ms span durations vs the 50ms PTY poll interval. The busy set previously inside the while-loop is preserved as a busy → busy no-op for legacy entry paths.

packages/opencode/src/session/status.ts
```diff
@@ -142,4 +168,17 @@ export async function list() {
 }
 // altimate_change end
 
+// altimate_change start (AI-7519) — publish a session.phase event. Fired by SessionPrompt.traceSpan
+// on entry (active=true) and exit (active=false); the TUI subscribes to render an honest label like
+// "Discovering warehouse tools..." during the pre-first-visible-response window. Best-effort: any
+// failure here must not affect the traced operation itself.
+export async function publishPhase(sessionID: SessionID, phase: string, active: boolean) {
+  try {
+    await Bus.publish(LegacyEvent.Phase, { sessionID, phase, active })
```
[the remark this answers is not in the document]
sahrizvi: Addressed in 6b4162399. Added `publishPhase` to the `SessionStatus.Service` interface + Layer implementation so `Event.Phase` is published through the EventV2 bridge alongside the legacy Bus mirror. Mirrors the pattern used by `set()` (Layer calls `events.publish(Event.Status, …)` internally, imperative wrapper adds the `Bus.publish(LegacyEvent.Status, …)` afterwards).

packages/opencode/test/cli/tui/phase-label.tui-e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+import { tmpdir } from "os"
+import { launchTui } from "../../fixture/pty-tui"
+
+function makeProjectDir(): string {
+  return mkdtempSync(path.join(tmpdir(), "altimate-tui-phase-"))
+}
+
+describe("TUI e2e — AI-7519 phase-label render", () => {
+  test("phase-label renders next to the busy spinner after a prompt is submitted", async () => {
+    const project = makeProjectDir()
```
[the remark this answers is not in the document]
sahrizvi: Addressed in 6b4162399. Replaced the manual `mkdtempSync` + finally-block cleanup with `await using tmp = await tmpdir()` — matches the codebase convention (scheduler.test.ts) and the CI-enforced coding guideline. Cleanup now runs even if `launchTui` rejects before the try block.

packages/opencode/test/cli/tui/phase-label.tui-e2e.test.ts
```diff
@@ -0,0 +1,83 @@
...
+        "Discovering tools": rendered.includes("Discovering tools"),
+        "Preparing telemetry": rendered.includes("Preparing telemetry"),
+        "Detecting project shape": rendered.includes("Detecting project shape"),
+      }
+      // eslint-disable-next-line no-console
+      console.log("[AI-7519 e2e] specific label hits:", JSON.stringify(specificLabels))
+      // The fallback assertion above is the load-bearing check; the specific
+      // labels are timing-sensitive (sub-millisecond spans + PTY polling
+      // interval). Not gated on hard equality.
+      expect(rendered).toContain("Thinking...")
```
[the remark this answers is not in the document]
sahrizvi: Addressed in 6b4162399. Added a hard `expect(rendered).toContain("Discovering tools")` assertion in addition to the "Thinking..." fallback. "Discovering tools" is the label mapped from `bootstrap.resolve-tools`, whose span (MCP tool listing) is comfortably longer than the PTY poll interval so this is safe from flake. Local re-run confirms both assertions pass; the specific-labels diagnostic dump is retained for future timing debugging on the sub-ms phases (Session.get / Config.get / Telemetry.init) that don't survive PTY sampling — those aren't gated so we don't fail on timing changes we can't control.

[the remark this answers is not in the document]
sahrizvi: Fair point, addressed in 6b4162399. Added `expect(rendered).toContain("Discovering tools")` so a regression that drops `publishPhase` → `Bus.publish` → `sync.tsx` → store update anywhere in the chain would fail this test. The fork-guard test is still there as the belt-and-suspenders string-match check; this e2e now covers the pipeline behaviorally too, not just structurally.

packages/opencode/src/session/status.ts
```diff
@@ -142,4 +185,23 @@ export async function list() {
...
+  try {
+    await runStatus((s) => s.publishPhase(sessionID, phase, active))
+  } catch {
+    // never surface phase-publish failures back to the caller
+  }
+  try {
+    await Bus.publish(LegacyEvent.Phase, { sessionID, phase, active })
+  } catch {
+    // never surface phase-publish failures back to the caller
+  }
```
dev-punia-altimate, line 207: **[🟠 MEDIUM]** The two publish operations are independent and can be executed concurrently to improve performance. You can use `Promise.allSettled` to run them in parallel, which also inherently satisfies the requirement of safely swallowing any potential failures without needing explicit `try...catch` blocks.

**Suggested change:**
```suggestion
  // Best-effort publish: run concurrently and intentionally swallow any failures 
  // so they never surface back to the caller.
  await Promise.allSettled([
    runStatus((s) => s.publishPhase(sessionID, phase, active)),
    Bus.publish(LegacyEvent.Phase, { sessionID, phase, active })
  ])
```

packages/opencode/test/fixture/pty-tui.ts
```diff
@@ -0,0 +1,218 @@
...
+    env,
+  })
+
+  let raw = ""
+  let stripped = ""
+  child.onData((data) => {
+    const chunk = data.toString()
+    raw += chunk
+    stripped += stripAnsi(chunk)
+  })
```
dev-punia-altimate, line 147: **[🟠 MEDIUM]** Applying `stripAnsi` individually to each incoming data chunk can cause issues. Terminal streams can arbitrarily split ANSI escape sequences across chunk boundaries (e.g., `\x1b[` in one chunk and `31m` in the next). In such cases, `stripAnsi` will fail to recognize and strip partial sequences, causing raw ANSI characters to leak permanently into the `stripped` string, leading to flaky `waitForText` assertions.

It is safer to accumulate the `raw` text and apply `stripAnsi(raw)` dynamically when `text()` or `waitForText()` is called.

**Suggested change:**
```suggestion
  let raw = ""
  child.onData((data) => {
    const chunk = data.toString()
    raw += chunk
  })
```

packages/opencode/test/fixture/pty-tui.ts
```diff
@@ -0,0 +1,218 @@
...
+      return raw
+    },
+    resize(c, r) {
+      if (disposed) return
+      child.resize(c, r)
+    },
+    async waitForText(needle, waitOpts) {
+      const timeoutMs = waitOpts?.timeoutMs ?? DEFAULT_WAIT_TIMEOUT_MS
+      const deadline = Date.now() + timeoutMs
+      const matches = (s: string) => (typeof needle === "string" ? s.includes(needle) : needle.test(s))
```
dev-punia-altimate: **[🟠 MEDIUM]** When the `needle` parameter is a `RegExp`, it is evaluated via `needle.test(s)` inside a polling loop. If a regular expression is provided with the global (`g`) flag, the `test()` method becomes stateful and advances its internal `lastIndex` property upon a match. Repeated calls on the same text inside the loop will unexpectedly return `false` on subsequent iterations, causing tests to fail randomly. 

Ensure regular expressions are evaluated statelessly (e.g., resetting `needle.lastIndex = 0` before testing, or using `s.match(needle)`).

**Suggested change:**
```suggestion
    async waitForText(needle, waitOpts) {
      const timeoutMs = waitOpts?.timeoutMs ?? DEFAULT_WAIT_TIMEOUT_MS
      const deadline = Date.now() + timeoutMs
      const matches = (s: string) => {
        if (typeof needle === "string") return s.includes(needle)
        needle.lastIndex = 0
        return needle.test(s)
      }
```

packages/opencode/test/fixture/pty-tui.ts
```diff
@@ -0,0 +1,218 @@
...
+      if (matches(stripped)) return
+      while (Date.now() < deadline) {
+        await new Promise((r) => setTimeout(r, POLL_INTERVAL_MS))
+        if (matches(stripped)) return
+      }
```
dev-punia-altimate, line 205: **[🟠 MEDIUM]** The polling loop waits until `timeoutMs` expires or the expected text is matched, but it does not check if the spawned TUI child process has exited. If the application crashes or exits prematurely before rendering the expected text, the test will hang and blindly poll until the timeout is reached. This drastically slows down test feedback and masks the root cause of the crash.

Consider incorporating an early exit condition, for example by tracking an `isExited` flag set in `child.onExit` and checking it inside this loop.

packages/opencode/src/session/prompt.ts
```diff
@@ -913,15 +982,25 @@ export namespace SessionPrompt {
...
-        bypassAgentCheck,
-        messages: msgs,
-      })
+      // altimate_change start (AI-7519) — trace resolveTools per step. Included
+      // in bootstrap on step===1; on subsequent steps this measures the
+      // per-turn tool-listing overhead (MCP.tools connect cost etc.).
+      const tools = await traceSpan(
+        "bootstrap.resolve-tools",
+        () =>
+          resolveTools({
```
dev-punia-altimate, line 1013: **[🔵 LOW]** The trace span name is hardcoded to `"bootstrap.resolve-tools"`. As the comment above indicates, this function is executed not only during the initial bootstrap (`step === 1`) but also on subsequent turns. Reusing the `"bootstrap."` prefix for non-bootstrap turns may pollute telemetry data, inflating the apparent frequency of bootstrap operations and distorting latency metrics.

Consider using a dynamic name based on the `step` to distinguish the initial tool discovery phase from subsequent per-turn overhead. This will correctly fall back to the safe `"Thinking..."` TUI phase label for subsequent turns.

**Suggested change:**
```suggestion
      const tools = await traceSpan(
        step === 1 ? "bootstrap.resolve-tools" : "turn.resolve-tools",
        () =>
          resolveTools({
```

packages/opencode/src/session/prompt.ts
```diff
@@ -1070,6 +1149,22 @@ export namespace SessionPrompt {
...
+        Tracer.active?.logSpan({
+          name: "bootstrap",
+          startTime: bootstrapStart,
+          endTime: Date.now(),
+          input: { agent: agent.name, sessionID },
+          output: {
+            duration_ms: Date.now() - bootstrapStart,
+            system_parts: system.length,
+          },
+        })
```
dev-punia-altimate, line 1191: **[🔵 LOW]** Calling `Date.now()` twice consecutively can result in a mismatch between `endTime` and `duration_ms` if the clock ticks between the two calls. It is more reliable to capture the time in a local variable to ensure precision and consistency.

**Suggested change:**
```suggestion
        const endTime = Date.now()
        Tracer.active?.logSpan({
          name: "bootstrap",
          startTime: bootstrapStart,
          endTime,
          input: { agent: agent.name, sessionID },
          output: {
            duration_ms: endTime - bootstrapStart,
            system_parts: system.length,
          },
        })
```

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 6 finding(s)

- 6 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `packages/opencode/src/session/status.ts` (L194-L203)

**[🟠 MEDIUM]** The two publish operations are independent and can be executed concurrently to improve performance. You can use `Promise.allSettled` to run them in parallel, which also inherently satisfies the requirement of safely swallowing any potential failures without needing explicit `try...catch` blocks.

**Suggested change:**
```suggestion
  // Best-effort publish: run concurrently and intentionally swallow any failures 
  // so they never surface back to the caller.
  await Promise.allSettled([
    runStatus((s) => s.publishPhase(sessionID, phase, active)),
    Bus.publish(LegacyEvent.Phase, { sessionID, phase, active })
  ])
```

#### 2. `packages/opencode/test/fixture/pty-tui.ts` (L141-L147)

**[🟠 MEDIUM]** Applying `stripAnsi` individually to each incoming data chunk can cause issues. Terminal streams can arbitrarily split ANSI escape sequences across chunk boundaries (e.g., `\x1b[` in one chunk and `31m` in the next). In such cases, `stripAnsi` will fail to recognize and strip partial sequences, causing raw ANSI characters to leak permanently into the `stripped` string, leading to flaky `waitForText` assertions.

It is safer to accumulate the `raw` text and apply `stripAnsi(raw)` dynamically when `text()` or `waitForText()` is called.

**Suggested change:**
```suggestion
  let raw = ""
  child.onData((data) => {
    const chunk = data.toString()
    raw += chunk
  })
```

#### 3. `packages/opencode/test/fixture/pty-tui.ts` (L177-L180)

**[🟠 MEDIUM]** When the `needle` parameter is a `RegExp`, it is evaluated via `needle.test(s)` inside a polling loop. If a regular expression is provided with the global (`g`) flag, the `test()` method becomes stateful and advances its internal `lastIndex` property upon a match. Repeated calls on the same text inside the loop will unexpectedly return `false` on subsequent iterations, causing tests to fail randomly. 

Ensure regular expressions are evaluated statelessly (e.g., resetting `needle.lastIndex = 0` before testing, or using `s.match(needle)`).

**Suggested change:**
```suggestion
    async waitForText(needle, waitOpts) {
      const timeoutMs = waitOpts?.timeoutMs ?? DEFAULT_WAIT_TIMEOUT_MS
      const deadline = Date.now() + timeoutMs
      const matches = (s: string) => {
        if (typeof needle === "string") return s.includes(needle)
        needle.lastIndex = 0
        return needle.test(s)
      }
```

#### 4. `packages/opencode/test/fixture/pty-tui.ts` (L181-L185)

**[🟠 MEDIUM]** The polling loop waits until `timeoutMs` expires or the expected text is matched, but it does not check if the spawned TUI child process has exited. If the application crashes or exits prematurely before rendering the expected text, the test will hang and blindly poll until the timeout is reached. This drastically slows down test feedback and masks the root cause of the crash.

Consider incorporating an early exit condition, for example by tracking an `isExited` flag set in `child.onExit` and checking it inside this loop.

#### 5. `packages/opencode/src/session/prompt.ts` (L988-L991)

**[🔵 LOW]** The trace span name is hardcoded to `"bootstrap.resolve-tools"`. As the comment above indicates, this function is executed not only during the initial bootstrap (`step === 1`) but also on subsequent turns. Reusing the `"bootstrap."` prefix for non-bootstrap turns may pollute telemetry data, inflating the apparent frequency of bootstrap operations and distorting latency metrics.

Consider using a dynamic name based on the `step` to distinguish the initial tool discovery phase from subsequent per-turn overhead. This will correctly fall back to the safe `"Thinking..."` TUI phase label for subsequent turns.

**Suggested change:**
```suggestion
      const tools = await traceSpan(
        step === 1 ? "bootstrap.resolve-tools" : "turn.resolve-tools",
        () =>
          resolveTools({
```

#### 6. `packages/opencode/src/session/prompt.ts` (L1157-L1166)

**[🔵 LOW]** Calling `Date.now()` twice consecutively can result in a mismatch between `endTime` and `duration_ms` if the clock ticks between the two calls. It is more reliable to capture the time in a local variable to ensure precision and consistency.

**Suggested change:**
```suggestion
        const endTime = Date.now()
        Tracer.active?.logSpan({
          name: "bootstrap",
          startTime: bootstrapStart,
          endTime,
          input: { agent: agent.name, sessionID },
          output: {
            duration_ms: endTime - bootstrapStart,
            system_parts: system.length,
          },
        })
```

</details>

packages/opencode/src/session/prompt.ts
```diff
@@ -359,17 +403,32 @@ export namespace SessionPrompt {
     let structuredOutput: unknown | undefined
 
     let step = 0
-    const session = await Session.get(sessionID)
+    // altimate_change start (AI-7519) — capture bootstrap start; emitted as a
+    // single "bootstrap" span right before the first processor.process call so
+    // the pre-first-generation region has a visible parent duration in traces.
+    const bootstrapStart = Date.now()
+    // altimate_change end
+    const session = await traceSpan(
```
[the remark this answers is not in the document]
sahrizvi: Same as the prior CodeRabbit round — addressed in 6b4162399.

packages/opencode/src/session/status.ts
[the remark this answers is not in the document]
sahrizvi: Same as the prior CodeRabbit round — addressed in 6b4162399.

packages/opencode/src/session/prompt.ts
```diff
@@ -90,6 +90,50 @@ const STRUCTURED_OUTPUT_SYSTEM_PROMPT = `IMPORTANT: The user has requested struc
...
+    sessionID?: SessionID,
+  ): Promise<T> {
+    const startTime = Date.now()
+    if (sessionID) void SessionStatus.publishPhase(sessionID, name, true)
+    try {
+      const result = await fn()
+      Tracer.active?.logSpan({ name, startTime, endTime: Date.now(), input })
+      return result
+    } catch (e) {
+      Tracer.active?.logSpan({
```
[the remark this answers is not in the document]
sahrizvi: Same as the prior CodeRabbit round — addressed in 6b4162399.

packages/opencode/src/session/prompt.ts
```diff
@@ -90,6 +90,50 @@ const STRUCTURED_OUTPUT_SYSTEM_PROMPT = `IMPORTANT: The user has requested struc
...
+        name,
+        startTime,
+        endTime: Date.now(),
```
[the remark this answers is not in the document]
sahrizvi: Same as the prior CodeRabbit round — addressed in 6b4162399.

packages/opencode/test/cli/tui/phase-label.tui-e2e.test.ts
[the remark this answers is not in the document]
sahrizvi: Same as the prior CodeRabbit round — addressed in 6b4162399.

packages/opencode/test/cli/tui/phase-label.tui-e2e.test.ts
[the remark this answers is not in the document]
sahrizvi: Same as the prior CodeRabbit round — addressed in 6b4162399.

packages/opencode/src/session/prompt.ts
```diff
@@ -90,6 +90,50 @@ const STRUCTURED_OUTPUT_SYSTEM_PROMPT = `IMPORTANT: The user has requested struc
...
+  // parentSpanId to rootSpanId), not a nested child — good enough for
+  // waterfall correlation via timestamps, and no schema change is required.
+  async function traceSpan<T>(
+    name: string,
+    fn: () => Promise<T>,
+    input?: unknown,
+    sessionID?: SessionID,
+  ): Promise<T> {
+    const startTime = Date.now()
+    if (sessionID) void SessionStatus.publishPhase(sessionID, name, true)
```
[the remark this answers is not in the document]
sahrizvi: Same as the prior CodeRabbit round — addressed in 6b4162399.

packages/opencode/src/session/prompt.ts
```diff
@@ -359,17 +403,42 @@ export namespace SessionPrompt {
...
+    // Enter busy state BEFORE the first bootstrap traceSpan fires so the
+    // phase labels the TUI renders are actually visible during
+    // session-get / config-get / fingerprint-detect / telemetry-init. The
+    // TUI's status renderer gates on `status.type === "busy"`; without
+    // this early set only `bootstrap.resolve-tools` (which fires inside
+    // the while-loop after the existing busy set at line 506) would show
+    // a label. The while-loop re-set below is now a no-op busy → busy
+    // transition, preserved for legacy call sites that may enter the
+    // loop from elsewhere.
+    await SessionStatus.set(sessionID, { type: "busy" })
```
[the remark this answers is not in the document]
sahrizvi: Same as the prior CodeRabbit round — addressed in 6b4162399.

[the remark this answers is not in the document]
sahrizvi: Good catch — legit stuck-busy risk from the busy-before-bootstrap change in 6b4162399. Fixed in 7bd403509: wrapped the four bootstrap traceSpan calls in a try/catch that best-effort transitions to idle before rethrowing.

packages/opencode/src/session/status.ts
sahrizvi: Legit concern but deferred — an initial attempt at Promise.allSettled caused an e2e regression I couldn't cleanly attribute. Wants a proper investigation of runStatus Layer/Effect concurrency semantics before I ship the change on the hot path. Tracking to revisit.

packages/opencode/test/fixture/pty-tui.ts
sahrizvi: Legit chunk-boundary concern in theory. Deferred — current per-chunk stripAnsi has been reliable across our tests, and switching text() from an O(1) memoized string to an O(n) computation is a semantic change I'd want its own review + coverage for.

packages/opencode/test/fixture/pty-tui.ts
sahrizvi: Applied in 762ac7c52. Pure defensive — our current specs don't use /g so nothing changes today, but future callers won't need to know the gotcha.

packages/opencode/test/fixture/pty-tui.ts
sahrizvi: Applied in 762ac7c52. Poll loop now breaks on child exit with a distinct 'child exit before match' error message so a crash surfaces in seconds instead of burning the full 8s timeout.

packages/opencode/src/session/prompt.ts
sahrizvi: Legit telemetry hygiene, ternary logic looks correct on paper (step===1 fires before resolveTools at line 583). But when I applied it the e2e stopped seeing 'Discovering tools' — couldn't confirm the label reliably survives on the boundary. Deferred for a focused pass so I don't regress the label rendering here.

packages/opencode/src/session/prompt.ts
sahrizvi: Applied in 762ac7c52. Cleaner math and duration_ms is now guaranteed to match endTime - startTime.

packages/opencode/src/session/status.ts
sahrizvi: Follow-up: tested this in isolation and it's actually not safe — 2/3 e2e runs failed reproducibly with Promise.allSettled. Best hypothesis is the first ManagedRuntime warm-up inside runStatus races with the immediate legacy Bus.publish. Sequential is required for reliability. Documented the constraint in cefb7f490 so the next reviewer doesn't reach the same suggestion.

packages/opencode/src/session/prompt.ts
sahrizvi: Follow-up: applied in cefb7f490. Isolated test showed the step-aware ternary works — my earlier attribution was wrong; the failure that made me defer was actually from #1 (Promise.allSettled), not this. 4/5 e2e pass; the 1 failure was first-run environmental (failed on the Thinking... fallback assertion before any resolve-tools code runs).

packages/opencode/test/fixture/pty-tui.ts
sahrizvi: Follow-up with data: applying this reduced the e2e pass rate from 5/5 baseline to 4/5 in isolation and to 2/5 when combined with #5. The computation-on-read pattern appears to add enough per-poll overhead to shift the timing window past the label's render duration on some runs. Legit theoretical concern about chunk-boundary ANSI, but not applying blind against no observed failure mode — worth revisiting once we see a concrete leak in a real test.

packages/opencode/test/fixture/pty-tui.ts
```diff
@@ -0,0 +1,238 @@
...
+      if (disposed) return
+      disposed = true
+      try {
+        child.kill("SIGTERM")
+      } catch {
+        // already gone
+      }
+      // give the process a beat to exit cleanly before returning
+      await Promise.race([exited, new Promise((r) => setTimeout(r, 1000))])
+    },
```
dev-punia-altimate, line 236: **[🟠 MEDIUM]** If the child process ignores the `SIGTERM` signal or hangs, it might not exit within the 1000ms timeout and could leave an orphaned test process running in the background. It's recommended to add a fallback that sends a `SIGKILL` signal if the process hasn't exited after the timeout grace period.

**Suggested change:**
```suggestion
    async dispose() {\n      if (disposed) return\n      disposed = true\n      try {\n        child.kill("SIGTERM")\n      } catch {\n        // already gone\n      }\n      // give the process a beat to exit cleanly before returning\n      await Promise.race([exited, new Promise((r) => setTimeout(r, 1000))])\n      if (!hasExited) {\n        try {\n          child.kill("SIGKILL")\n        } catch {\n          // already gone\n        }\n      }\n    },
```

packages/opencode/src/session/status.ts
```diff
@@ -142,4 +185,27 @@ export async function list() {
...
```
dev-punia-altimate, line 207: **[🔵 LOW]** Silent error swallowing can make debugging difficult if `publishPhase` or `Bus.publish` fail unexpectedly. While it's correct to not surface these errors back to the caller (so the main trace operation isn't affected), logging them at a trace or debug level (e.g., using `console.error` or a dedicated `Log`) would ensure that failures are recorded for troubleshooting.

Consider adding a log statement in the `catch` blocks:
```typescript
catch (e) {
  // log.warn("phase-publish failure", { error: e })
  // never surface phase-publish failures back to the caller
}
```

packages/opencode/src/session/prompt.ts
```diff
@@ -359,17 +403,61 @@ export namespace SessionPrompt {
...
+    let altCfg: Awaited<ReturnType<typeof Config.get>>
+    try {
+      session = await traceSpan(
+        "bootstrap.session-get",
+        () => Session.get(sessionID),
+        { sessionID },
+        sessionID,
+      )
+      // altimate_change start - detect environment fingerprint at session start
+      altCfg = await traceSpan("bootstrap.config-get", () => Config.get(), undefined, sessionID)
```
dev-punia-altimate, line 439: **[🔵 LOW]** Code Quality (Variable Scope): Since `altCfg` is only used inside this `try` block for evaluating the experimental fingerprint flag, it should be declared locally as a `const` within the block rather than being lifted to the outer scope with a `let` binding. This keeps the variable scope as narrow as possible.

**Suggested change:**
```suggestion
    let session: Awaited<ReturnType<typeof Session.get>>
    try {
      session = await traceSpan(
        "bootstrap.session-get",
        () => Session.get(sessionID),
        { sessionID },
        sessionID,
      )
      // altimate_change start - detect environment fingerprint at session start
      const altCfg = await traceSpan("bootstrap.config-get", () => Config.get(), undefined, sessionID)
```

dev-punia-altimate (commented):

- 0 without a line anchor

#### 1. `packages/opencode/test/fixture/pty-tui.ts` (L143-L147)

**[🟠 MEDIUM]** Applying `stripAnsi` to individual chunks is risky because an ANSI escape sequence might be split across two chunks (e.g., `\x1b[` in one chunk and `31m` in the next). When this happens, `stripAnsi` will fail to recognize the partial sequence, leaving ANSI garbage characters in the `stripped` string and potentially causing flaky tests downstream. It's safer to derive `stripped` by calling `stripAnsi` on the accumulated `raw` string.

**Suggested change:**
```suggestion
  child.onData((data) => {\n    const chunk = data.toString()\n    raw += chunk\n    stripped = stripAnsi(raw)\n  })
```

#### 2. `packages/opencode/test/fixture/pty-tui.ts` (L212-L222)

**[🟠 MEDIUM]** If the child process ignores the `SIGTERM` signal or hangs, it might not exit within the 1000ms timeout and could leave an orphaned test process running in the background. It's recommended to add a fallback that sends a `SIGKILL` signal if the process hasn't exited after the timeout grace period.

**Suggested change:**
```suggestion
    async dispose() {\n      if (disposed) return\n      disposed = true\n      try {\n        child.kill("SIGTERM")\n      } catch {\n        // already gone\n      }\n      // give the process a beat to exit cleanly before returning\n      await Promise.race([exited, new Promise((r) => setTimeout(r, 1000))])\n      if (!hasExited) {\n        try {\n          child.kill("SIGKILL")\n        } catch {\n          // already gone\n        }\n      }\n    },
```

#### 3. `packages/opencode/src/session/status.ts` (L198-L207)

**[🔵 LOW]** Silent error swallowing can make debugging difficult if `publishPhase` or `Bus.publish` fail unexpectedly. While it's correct to not surface these errors back to the caller (so the main trace operation isn't affected), logging them at a trace or debug level (e.g., using `console.error` or a dedicated `Log`) would ensure that failures are recorded for troubleshooting.

Consider adding a log statement in the `catch` blocks:
```typescript
catch (e) {
  // log.warn("phase-publish failure", { error: e })
  // never surface phase-publish failures back to the caller
}
```

#### 4. `packages/opencode/src/session/prompt.ts` (L429-L439)

**[🔵 LOW]** Code Quality (Variable Scope): Since `altCfg` is only used inside this `try` block for evaluating the experimental fingerprint flag, it should be declared locally as a `const` within the block rather than being lifted to the outer scope with a `let` binding. This keeps the variable scope as narrow as possible.

**Suggested change:**
```suggestion
    let session: Awaited<ReturnType<typeof Session.get>>
    try {
      session = await traceSpan(
        "bootstrap.session-get",
        () => Session.get(sessionID),
        { sessionID },
        sessionID,
      )
      // altimate_change start - detect environment fingerprint at session start
      const altCfg = await traceSpan("bootstrap.config-get", () => Config.get(), undefined, sessionID)
```

</details>

packages/opencode/test/fixture/pty-tui.ts
sahrizvi: Fixed in 27b3bb885. `dispose()` now escalates to `SIGKILL` if the child hasn't exited within the 1000ms grace after `SIGTERM`, then waits a final 500ms. The happy path (child exits cleanly within the grace) is byte-for-byte unchanged. Verified: typecheck clean; `phase-label.tui-e2e.test.ts` passes (fixture launches + disposes cleanly).

packages/opencode/src/session/prompt.ts
```diff
@@ -913,15 +1001,28 @@ export namespace SessionPrompt {
...
-        messages: msgs,
-      })
+      // altimate_change start (AI-7519) — trace resolveTools per step.
+      // Included in the parent `bootstrap` span on step===1; on later steps
+      // this measures the per-turn tool-listing overhead (MCP.tools connect
+      // cost etc.). Distinct span name per phase so telemetry doesn't
+      // double-count non-bootstrap turns under "bootstrap.*", and the TUI
+      // falls back to the safe "Thinking..." label on later turns.
+      const tools = await traceSpan(
+        step === 1 ? "bootstrap.resolve-tools" : "turn.resolve-tools",
```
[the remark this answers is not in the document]
sahrizvi: By design — not fixing this one. The step-aware span name is intentional: `resolveTools` is emitted as `bootstrap.resolve-tools` on `step === 1` and `turn.resolve-tools` after (see prompt.ts:1005-1009). That split does two things — it keeps per-turn tool-listing out of the `bootstrap.*` telemetry bucket, and it lets the TUI fall back to the generic `Thinking...` on later turns instead of re-flashing `Discovering tools...` every turn. The "Discovering tools" label is only meaningful during the cold first-turn MCP listing; showing it each turn (when tools are warm) would be misleading flicker. `phase-label.ts` documents the unmapped → `Thinking...` fallback as the intended default. Adding a `turn.resolve-tools` entry would reintroduce that flicker, so leaving it unmapped is deliberate.

packages/opencode/src/session/prompt.ts
sahrizvi: Fair and accurate — unlike `session` (which is used after the `try` and so must be hoisted), `altCfg` is only read inside the `try`, so it could be a local `const`. It was declared with `let` above the block purely for symmetry with `session`. It's harmless (single assignment, never reassigned), so I'm leaving it as-is on this PR rather than churn the bootstrap try/catch for a style nit — can fold into a follow-up cleanup if we want it tidy.

packages/opencode/src/session/status.ts
sahrizvi: By design. The two swallowed catches are deliberate: `publishPhase` is best-effort UX signal (the phase label), and per the contract at status.ts:192 "any failure here must not affect the traced operation itself" — a failed publish must be invisible, not surface an error to the caller mid-generation. This is the fail-safe half of the observability path. (The related MEDIUM on this same block — parallelising the two publishes via `Promise.allSettled` — was investigated and kept sequential for reliability; see that thread.) Leaving the swallow intentional; happy to add a debug-level log inside the catches in a follow-up if we want a diagnostic breadcrumb without changing behavior.

dev-punia-altimate:

No supported files changed.

dev-punia-altimate:
## 🤖 Code Review — OpenCodeReview (Gemini) — 2 finding(s)

- 2 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `packages/opencode/test/fixture/pty-tui.ts` (L141-L147)

**[🟠 MEDIUM]** Applying `stripAnsi` to individual chunk fragments in the stream is unreliable. PTY stream chunks can split exactly in the middle of ANSI escape sequences (e.g., yielding `\x1b` in one chunk and `[31m` in the next). When this happens, `stripAnsi` doesn't recognize the sequence and leaks the raw ANSI characters into `stripped`, causing unpredictable test flakiness when `waitForText` subsequently fails to match expected clean text. 

To resolve this, you should run `stripAnsi` against the full accumulated buffer `raw` so it always has the complete context.

**Suggested change:**
```suggestion
  let raw = ""
  let stripped = ""
  child.onData((data) => {
    const chunk = data.toString()
    raw += chunk
    stripped = stripAnsi(raw)
  })
```

#### 2. `packages/opencode/src/session/status.ts` (L198-L207)

**[🔵 LOW]** Empty `catch` blocks silently swallow errors. While it is intentional not to throw these errors back to the caller (so that telemetry failures don't disrupt the main application flow), silencing them completely obscures underlying transport or serialization bugs. This makes it difficult to debug if phase labels fail to render in the UI.

Consider importing the logger (`import { Log } from "../util/log"`) and logging a warning instead of keeping the block empty. For example, you can initialize a logger at the top of the file: `const log = Log.create({ service: "session.status" })` and use it here.

**Suggested change:**
```suggestion
  try {
    await runStatus((s) => s.publishPhase(sessionID, phase, active))
  } catch (error) {
    // never surface phase-publish failures back to the caller
    log.warn("V2 phase publish failed", { error, sessionID, phase })
  }
  try {
    await Bus.publish(LegacyEvent.Phase, { sessionID, phase, active })
  } catch (error) {
    // never surface phase-publish failures back to the caller
    log.warn("Legacy phase publish failed", { error, sessionID, phase })
  }
```

</details>

saravmajestic (approved):
LGTM

PR #1005, opened 2026-07-15
feat(tui): install skills via typed query in the /skills list

sahrizvi:
### Issue for this PR

Closes #

_No linked issue — internal quality fix on the fork's inline `/skills` dialog._

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Gives the `/skills` list a discoverable Enter-driven install path.

When the filter text matches an installable shape (github URL, clean `owner/repo`, or an absolute path), a synthetic `Install <query>` row is prepended at the top of the list. Enter opens the install dialog with the typed text prefilled. Symmetric wiring on the create dialog.

**Why:** the pre-existing `ctrl+i` shortcut can't be relied on — its wire byte (`0x09`) collides with `Tab` on default terminals, so users trying to install from a URL/repo/path they typed into the search box saw the list filter to zero results and had no obvious next action.

**How:** backing the "when to offer the row" check with the **same** classifier the installer uses (`classifyInstallSource`) kills a class of bugs where the list would advertise an install of things the installer would then reject — e.g. skill-search queries like `dbt/snowflake` or `owner/repo/subpath`.

**Files:**
- `packages/opencode/src/plugin/tui/altimate/skill-ops.tsx`
  - New `classifyInstallSource(source)` — recognises `github-url`, `owner-repo`, `absolute-path`; returns `null` for short strings, sub-paths, relative paths, and `~`-paths.
  - New `normalizeInstallSource(source)` — shared trim/strip helper (also used by `installSkillDirect` when building the clone URL, so `owner/repo.git` can't produce `owner/repo.git.git`).
  - Module-scope `INSTALL_ACTION_VALUE` sentinel — namespaced so it can't collide with real skill names.
  - `DialogSkillList` — synthetic Install row via spread (pure memo); `onSelect` routes sentinel to `showInstall`; `onMove` filters sentinel; `showActions` bails on sentinel/lookup-miss.
  - `options` memo split into `baseOptions` (deps on `skills()` only) + composed `options` (deps on `filter()`) — avoids re-running `detectToolReferences` regex parse per skill on every keystroke.
- `packages/opencode/test/altimate/skill-install-classifier.test.ts` *(new)* — 14 unit tests pinning classifier + normalizer edge cases (owner/repo, http(s), POSIX absolute, Windows drive-letter, short-string rejection, multi-slash rejection, `~`/relative rejection, `.git` / trailing-dot / whitespace stripping).
- `packages/opencode/test/upstream/fork-feature-guards.test.ts` — slimmed the merge-drop guard to a minimal presence check (classifier symbol, sentinel, sentinel-route, prefill plumbing on both sub-dialogs).

Local validation (all green):
- `bun run typecheck` — clean across the whole workspace
- `bun test test/altimate/skill-install-classifier.test.ts` — 14 pass / 0 fail / 54 expects
- `bun test test/upstream/fork-feature-guards.test.ts` — 18 pass / 0 fail / 61 expects
- `bun test test/session` — 731 pass / 0 fail / 1636 expects

Manual smoke plan on a built binary:
- `/skills` → type `https://github.com/anthropics/skills.git` → Enter opens install dialog with URL prefilled.
- `/skills` → type `owner/repo` → Enter opens install dialog with `owner/repo` prefilled → clone URL is `https://github.com/owner/repo.git` (single `.git`, no double-suffix).
- `/skills` → type `owner/repo.git` → same clone URL (double-suffix bug regression check).
- `/skills` → type `dbt-develop` (real search term) → no Install row surfaces; list filters skills normally.
- Highlight the synthetic Install row → `ctrl+a` does nothing (sentinel bail-out).

### Screenshots / recordings

_Terminal-only UI change — no screenshot supplied. Behaviour is fully covered by the classifier unit tests and the fork-feature-guards presence checks._

### Checklist

### Follow-ups intentionally deferred (from prior review)

- Conditional placeholder text (m6) — the `Search skills, or type a repo/URL…` copy shows even when the filter isn't installable. Low priority.
- Prefill cursor-at-end (m7) — `DialogPrompt`'s cursor lands at end-of-text; fine for appending `/path`, mildly awkward for editing a pasted URL.

### Review response

Addressed both findings from OpenCodeReview (Gemini) in commit `b8e7e64c3`:
- **MEDIUM Bug** (`skill-ops.tsx:177-180`) — `owner/repo.git` produced `owner/repo.git.git`. Extracted `normalizeInstallSource` and use it in both classifier and URL builder. Regression covered by new normalizer test suite.
- **MEDIUM Perf** (`skill-ops.tsx:592-596`) — `filter()` reactivity made `detectToolReferences` re-run per keystroke. Split into `baseOptions` (skills-only) + composed `options` (filter-aware).

## Summary by CodeRabbit

* **New Features**
  * Added installing skills from GitHub URLs, `owner/repo` shorthand, and absolute filesystem paths.
  * Added a synthetic **“Install <query>”** entry when the current filter matches a recognized source.
  * Prefilled create and install dialogs with the current filter text; palette actions now carry the active filter through.
* **Bug Fixes**
  * Improved install action selection so actions only open for a real, resolvable skill.
* **Tests**
  * Added Bun coverage for install-source classification/normalization and dialog option behavior.

### DIFF
--- a/packages/opencode/src/plugin/tui/altimate/skill-ops.tsx
+++ b/packages/opencode/src/plugin/tui/altimate/skill-ops.tsx
@@ -43,8 +43,17 @@ const id = "altimate:skill-ops"
 // surface the synthetic Install row when Enter will actually try to install.
 export type InstallSourceKind = "github-url" | "owner-repo" | "absolute-path"
 const OWNER_REPO_REGEX = /^[a-zA-Z0-9_-]+\/[a-zA-Z0-9._-]+$/
+
+// Strip the surface variation the classifier tolerates — whitespace, trailing dots,
+// and a `.git` suffix — before comparing. Exported so callers that consume the
+// classified string (e.g. installSkillDirect building a clone URL) can use the exact
+// same shape as the classifier, avoiding double-suffix bugs like `owner/repo.git.git`.
+export function normalizeInstallSource(source: string): string {
+  return source.trim().replace(/\.+$/, "").replace(/\.git$/, "")
+}
+
 export function classifyInstallSource(source: string): InstallSourceKind | null {
-  const trimmed = source.trim().replace(/\.+$/, "").replace(/\.git$/, "")
+  const trimmed = normalizeInstallSource(source)
   if (trimmed.length < 3) return null
   if (trimmed.startsWith("http://") || trimmed.startsWith("https://")) return "github-url"
   if (OWNER_REPO_REGEX.test(trimmed)) return "owner-repo"
@@ -176,7 +185,12 @@ async function installSkillDirect(
   // reach the installer without going through the UI's install-preview.
   const kind = classifyInstallSource(normalized)
   if (kind === "github-url" || kind === "owner-repo") {
-    const url = normalized.startsWith("http") ? normalized : `https://github.com/${normalized}.git`
+    // Build the clone URL from the *normalized* source so a `.git` suffix or trailing
+    // dot in the typed query doesn't produce `owner/repo.git.git` (bug reported by
+    // OpenCodeReview): the classifier tolerates those shapes but the raw `normalized`
+    // still carries them until stripped.
+    const cleaned = normalizeInstallSource(normalized)
+    const url = cleaned.startsWith("http") ? cleaned : `https://github.com/${cleaned}.git`
     const label = url.replace(/https?:\/\/github\.com\//, "").replace(/\.git$/, "")
     onProgress?.(`Cloning ${label}...`)
     const cache = cacheDir()
@@ -589,10 +603,14 @@ function DialogSkillList(props: { api: TuiPluginApi; onCurrent: (skill: string |
   currentFilter = filter
   // altimate_change end
 
-  const options = createMemo<TuiDialogSelectOption<string>[]>(() => {
+  // Base list — depends only on `skills()`. Kept separate from the `options` memo below
+  // so that per-keystroke filter changes don't re-run `detectToolReferences` (regex parse
+  // per skill) across the whole list. On projects with many installed skills the previous
+  // fused memo produced noticeable typing lag.
+  const baseOptions = createMemo<TuiDialogSelectOption<string>[]>(() => {
     const list = skills() ?? []
     const maxWidth = Math.max(0, ...list.map((s) => s.name.length))
-    const items = list.map((skill) => {
+    return list.map((skill) => {
       const tools = detectToolReferences(skill.content)
       const category = SKILL_CATEGORIES[skill.name] ?? "Other"
       const desc = skill.description?.replace(/\s+/g, " ").trim()
@@ -605,6 +623,10 @@ function DialogSkillList(props: { api: TuiPluginApi; onCurrent: (skill: string |
         category,
       } satisfies TuiDialogSelectOption<string>
     })
+  })
+
+  const options = createMemo<TuiDialogSelectOption<string>[]>(() => {
+    const items = baseOptions()
     // altimate_change start — when the filter looks like a shape installSkillDirect will
     // accept (github URL, `owner/repo` shorthand, or absolute path), prepend a synthetic
     // "Install <query>" top option. Selecting it (Enter) routes to installSkillDirect.
--- a/packages/opencode/src/plugin/tui/altimate/skill-ops.tsx
+++ b/packages/opencode/src/plugin/tui/altimate/skill-ops.tsx
@@ -34,6 +34,32 @@ import fs from "fs/promises"
 
 const id = "altimate:skill-ops"
 
+// altimate_change start — single classifier shared by installSkillDirect and the skill-list
+// dialog's "Install <query>" affordance. Extracted so the UI's install-preview and the
+// installer can't drift (an earlier `looksInstallable` that only checked `q.includes("/")`
+// showed the Install option for skill-search queries like "dbt/snowflake" which the
+// installer then rejected as "Path not found"). Returns null for shapes the installer
+// wouldn't accept without ambiguity — search terms, relative paths, `~` — so we only
+// surface the synthetic Install row when Enter will actually try to install.
+export type InstallSourceKind = "github-url" | "owner-repo" | "absolute-path"
+const OWNER_REPO_REGEX = /^[a-zA-Z0-9_-]+\/[a-zA-Z0-9._-]+$/
+export function classifyInstallSource(source: string): InstallSourceKind | null {
+  const trimmed = source.trim().replace(/\.+$/, "").replace(/\.git$/, "")
+  if (trimmed.length < 3) return null
+  if (trimmed.startsWith("http://") || trimmed.startsWith("https://")) return "github-url"
+  if (OWNER_REPO_REGEX.test(trimmed)) return "owner-repo"
+  if (trimmed.startsWith("/")) return "absolute-path"
+  if (/^[a-zA-Z]:[\\/]/.test(trimmed)) return "absolute-path"
+  return null
+}
+
+// Sentinel value for the synthetic top-of-list "Install <query>" option. Namespaced so it
+// can't collide with any skill name (skill names match `^[a-z][a-z0-9]*(-[a-z0-9]+)*$`).
+// Module-scope so it's stable across renders and reachable by every function that needs
+// to check whether an item.value is the sentinel (onSelect, onMove, showActions).
+const INSTALL_ACTION_VALUE = "__altimate:skill:install-from-query__"
+// altimate_change end
+
 // Categorize skills by domain for cleaner grouping in the list.
 const SKILL_CATEGORIES: Record<string, string> = {
   "dbt-develop": "dbt",
@@ -143,11 +169,13 @@ async function installSkillDirect(
     normalized = `https://github.com/${ghWebMatch[1]}.git`
   }
 
-  if (
-    normalized.startsWith("http://") ||
-    normalized.startsWith("https://") ||
-    normalized.match(/^[a-zA-Z0-9_-]+\/[a-zA-Z0-9._-]+$/)
-  ) {
+  // Classify the source. Anything the classifier recognises (github-url / owner-repo /
+  // absolute-path) takes the corresponding branch. Anything it doesn't (relative path,
+  // `~`, bare identifier) falls through to the path branch below and is resolved against
+  // cwd — that's the historical behaviour, preserved as a fallback for callers that
+  // reach the installer without going through the UI's install-preview.
+  const kind = classifyInstallSource(normalized)
+  if (kind === "github-url" || kind === "owner-repo") {
     const url = normalized.startsWith("http") ? normalized : `https://github.com/${normalized}.git`
     const label = url.replace(/https?:\/\/github\.com\//, "").replace(/\.git$/, "")
     onProgress?.(`Cloning ${label}...`)
@@ -295,7 +323,7 @@ async function reloadAndVerify(api: TuiPluginApi, expectedNames: string[]): Prom
 
 // ── Sub-dialogs ─────────────────────────────────────────────────────────────────────────────────
 
-function DialogSkillCreate(props: { api: TuiPluginApi }) {
+function DialogSkillCreate(props: { api: TuiPluginApi; initialValue?: string }) {
   const { api } = props
   const theme = () => api.theme.current
   const [busy, setBusy] = createSignal(false)
@@ -303,6 +331,7 @@ function DialogSkillCreate(props: { api: TuiPluginApi }) {
     <api.ui.DialogPrompt
       title="Create Skill"
       placeholder="my-tool"
+      value={props.initialValue}
       busy={busy()}
       busyText="Creating skill..."
       description={() => (
@@ -350,7 +379,7 @@ function DialogSkillCreate(props: { api: TuiPluginApi }) {
   )
 }
 
-function DialogSkillInstall(props: { api: TuiPluginApi }) {
+function DialogSkillInstall(props: { api: TuiPluginApi; initialValue?: string }) {
   const { api } = props
   const theme = () => api.theme.current
   const [busy, setBusy] = createSignal(false)
@@ -359,6 +388,7 @@ function DialogSkillInstall(props: { api: TuiPluginApi }) {
     <api.ui.DialogPrompt
       title="Install Skill (owner/repo, URL, or path)"
       placeholder="anthropics/skills"
+      value={props.initialValue}
       busy={busy()}
       busyText="Installing skill..."
       description={() => (
@@ -552,10 +582,17 @@ function DialogSkillList(props: { api: TuiPluginApi; onCurrent: (skill: string |
   // Expose the lookup to the keymap-layer commands (ctrl+a/test/etc).
   registerLookup(api, skillMap)
 
+  // altimate_change start — capture the current filter text so an install/create
+  // action triggered from inside the list can prefill the sub-dialog with what the user typed
+  // (e.g. a GitHub URL typed into the search box), instead of dropping it on the floor.
+  const [filter, setFilter] = createSignal("")
+  currentFilter = filter
+  // altimate_change end
+
   const options = createMemo<TuiDialogSelectOption<string>[]>(() => {
     const list = skills() ?? []
     const maxWidth = Math.max(0, ...list.map((s) => s.name.length))
-    return list.map((skill) => {
+    const items = list.map((skill) => {
       const tools = detectToolReferences(skill.content)
       const category = SKILL_CATEGORIES[skill.name] ?? "Other"
       const desc = skill.description?.replace(/\s+/g, " ").trim()
@@ -568,15 +605,47 @@ function DialogSkillList(props: { api: TuiPluginApi; onCurrent: (skill: string |
         category,
       } satisfies TuiDialogSelectOption<string>
     })
+    // altimate_change start — when the filter looks like a shape installSkillDirect will
+    // accept (github URL, `owner/repo` shorthand, or absolute path), prepend a synthetic
+    // "Install <query>" top option. Selecting it (Enter) routes to installSkillDirect.
+    // Non-mutating build (spread instead of unshift) so this stays a pure memo — Solid
+    // dev-mode double-eval would otherwise double-prepend.
+    // ctrl+i on the wire is byte 0x09 = Tab, so we can't offer a reliable ctrl+i binding
+    // on default terminals; Enter on this synthetic row is the discoverable substitute.
+    const q = filter().trim()
+    if (classifyInstallSource(q) !== null) {
+      const installOption: TuiDialogSelectOption<string> = {
+        title: `Install ${q}`,
+        description: "Press Enter to install from this GitHub repo, URL, or path",
+        footer: undefined,
+        value: INSTALL_ACTION_VALUE,
+        category: "Install",
+      }
+      return [installOption, ...items]
+    }
+    // altimate_change end
+    return items
   })
 
   return (
     <api.ui.DialogSelect
-      title="Skills (ctrl+a actions · ctrl+n new · ctrl+i install)"
-      placeholder="Search skills..."
+      title="Skills"
+      placeholder="Search skills, or type a repo/URL and press Enter to install..."
       options={options()}
-      onMove={(item) => props.onCurrent(item.value)}
+      onFilter={(q) => setFilter(q)}
+      // altimate_change start — filter the sentinel out of onCurrent so highlighting
+      // the synthetic Install row doesn't set `currentSkill` to the sentinel string
+      // (which would trip ctrl+a's showActions into opening a degenerate action picker
+      // on a non-existent skill named INSTALL_ACTION_VALUE).
+      onMove={(item) => props.onCurrent(item.value === INSTALL_ACTION_VALUE ? undefined : item.value)}
+      // altimate_change end
       onSelect={(item) => {
+        // altimate_change start — synthetic install option routes to installer.
+        if (item.value === INSTALL_ACTION_VALUE) {
+          showInstall(api, filter().trim() || undefined)
+          return
+        }
+        // altimate_change end
         props.onCurrent(item.value)
         // Selecting a skill opens its action picker (the pre-merge default action was the picker).
         openActionPicker(api, skillMap().get(item.value), item.value, () => showList(api))
@@ -589,34 +658,48 @@ function DialogSkillList(props: { api: TuiPluginApi; onCurrent: (skill: string |
 // Both are module-level so the layer (registered once at plugin init) can reach the live values.
 let currentSkill: string | undefined
 let currentLookup: () => Map<string, SkillInfo> = () => new Map()
+// altimate_change start — expose the current list dialog's filter text so the outer
+// keymap-layer install/create commands (ctrl+i/ctrl+n) can prefill the sub-dialog with whatever
+// the user has typed into the search box. Reset to a no-op after the list closes so a fresh
+// keybind press from an unrelated context doesn't leak stale text.
+let currentFilter: () => string = () => ""
+// altimate_change end
 
 function registerLookup(_api: TuiPluginApi, lookup: () => Map<string, SkillInfo>) {
   currentLookup = lookup
 }
 
 function showList(api: TuiPluginApi) {
-  api.ui.dialog.replace(() => (
-    <DialogSkillList api={api} onCurrent={(skill) => (currentSkill = skill)} />
-  ))
+  api.ui.dialog.replace(
+    () => <DialogSkillList api={api} onCurrent={(skill) => (currentSkill = skill)} />,
+    () => {
+      currentFilter = () => ""
+    },
+  )
 }
 
-function showCreate(api: TuiPluginApi) {
+function showCreate(api: TuiPluginApi, initialValue?: string) {
   api.ui.dialog.replace(
-    () => <DialogSkillCreate api={api} />,
+    () => <DialogSkillCreate api={api} initialValue={initialValue} />,
     () => setTimeout(() => showList(api), 0),
   )
 }
 
-function showInstall(api: TuiPluginApi) {
+function showInstall(api: TuiPluginApi, initialValue?: string) {
   api.ui.dialog.replace(
-    () => <DialogSkillInstall api={api} />,
+    () => <DialogSkillInstall api={api} initialValue={initialValue} />,
     () => setTimeout(() => showList(api), 0),
   )
 }
 
 function showActions(api: TuiPluginApi) {
-  if (!currentSkill) return
-  openActionPicker(api, currentLookup().get(currentSkill), currentSkill, () => showList(api))
+  // Belt-and-braces against the synthetic Install row leaking into `currentSkill` via a
+  // future refactor that forgets the DialogSkillList `onMove` filter — refuse to open a
+  // per-skill action picker unless we can resolve a real skill entry from the lookup.
+  if (!currentSkill || currentSkill === INSTALL_ACTION_VALUE) return
+  const info = currentLookup().get(currentSkill)
+  if (!info) return
+  openActionPicker(api, info, currentSkill, () => showList(api))
 }
 
 const tui: TuiPlugin = async (api) => {
@@ -649,7 +732,7 @@ const tui: TuiPlugin = async (api) => {
         category: "Altimate",
         namespace: "palette",
         run() {
-          showCreate(api)
+          showCreate(api, currentFilter().trim() || undefined)
         },
       },
       {
@@ -659,7 +742,7 @@ const tui: TuiPlugin = async (api) => {
         category: "Altimate",
         namespace: "palette",
         run() {
-          showInstall(api)
+          showInstall(api, currentFilter().trim() || undefined)
         },
       },
     ],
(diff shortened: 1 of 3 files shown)


dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 2 finding(s)

- 2 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `packages/opencode/src/plugin/tui/altimate/skill-ops.tsx` (L592-L596)

**[🟠 MEDIUM]** **Performance Issue: Unnecessary reactivity overhead**

Inside this `createMemo`, the signal `filter()` is read. This means the memo depends on `filter()` and will re-evaluate on every keystroke. During this re-evaluation, the `list.map` iterates over all skills, executing `detectToolReferences` (which parses content using regex). This can cause noticeable UI lag during search when there are many skills.

**Suggestion:** 
Separate the reactivity into two memos: one for computing the base list of `skills()` (which only updates when the underlying skills data changes), and another for handling the dynamic "Install" option that reacts to `filter()` changes.

```typescript
  const baseOptions = createMemo<TuiDialogSelectOption<string>[]>(() => {
    const list = skills() ?? []
    const maxWidth = Math.max(0, ...list.map((s) => s.name.length))
    return list.map((skill) => {
      const tools = detectToolReferences(skill.content)
      const category = SKILL_CATEGORIES[skill.name] ?? "Other"
      const desc = skill.description?.replace(/\s+/g, " ").trim()
      return {
        title: skill.name.padEnd(maxWidth),
        description: desc,
        footer: tools.length > 0 ? tools.map((t) => `\u25b6 ${t}`).join("  ") : undefined,
        value: skill.name,
        category,
      } satisfies TuiDialogSelectOption<string>
    })
  })

  const options = createMemo<TuiDialogSelectOption<string>[]>(() => {
    const items = baseOptions()
    const q = filter().trim()
    if (classifyInstallSource(q) !== null) {
      const installOption: TuiDialogSelectOption<string> = {
        title: `Install ${q}`,
        description: "Press Enter to install from this GitHub repo, URL, or path",
        footer: undefined,
        value: INSTALL_ACTION_VALUE,
        category: "Install",
      }
      return [installOption, ...items]
    }
    return items
  })
```

#### 2. `packages/opencode/src/plugin/tui/altimate/skill-ops.tsx` (L177-L180)

**[🟠 MEDIUM]** **Bug: Inconsistent string normalization**

The `classifyInstallSource` function accepts leading/trailing whitespaces and strips `.git` suffixes (e.g., `" owner/repo "`), returning a valid `kind`. However, `installSkillDirect` relies on the unmodified `normalized` string.
If `normalized` has spaces or an existing `.git` suffix (e.g., `" owner/repo "` or `"owner/repo.git"`), it passes the `if` check but generates a malformed URL like `https://github.com/ owner/repo .git` or `https://github.com/owner/repo.git.git`.

**Suggestion:**
Ensure the string used to construct the URL matches the cleaned format recognized by the classifier.

```typescript
  const kind = classifyInstallSource(normalized)
  if (kind === "github-url" || kind === "owner-repo") {
    const cleaned = normalized.trim().replace(/\.+$/, "").replace(/\.git$/, "")
    const url = cleaned.startsWith("http") ? cleaned : `https://github.com/${cleaned}.git`
    const label = url.replace(/https?:\/\/github\.com\//, "").replace(/\.git$/, "")
```

</details>

packages/opencode/src/plugin/tui/altimate/skill-ops.tsx
```diff
@@ -552,10 +582,17 @@ function DialogSkillList(props: { api: TuiPluginApi; onCurrent: (skill: string |
...
+  const [filter, setFilter] = createSignal("")
+  currentFilter = filter
+  // altimate_change end
+
   const options = createMemo<TuiDialogSelectOption<string>[]>(() => {
     const list = skills() ?? []
     const maxWidth = Math.max(0, ...list.map((s) => s.name.length))
-    return list.map((skill) => {
+    const items = list.map((skill) => {
       const tools = detectToolReferences(skill.content)
```
dev-punia-altimate: **[🟠 MEDIUM]** **Performance Issue: Unnecessary reactivity overhead**

Inside this `createMemo`, the signal `filter()` is read. This means the memo depends on `filter()` and will re-evaluate on every keystroke. During this re-evaluation, the `list.map` iterates over all skills, executing `detectToolReferences` (which parses content using regex). This can cause noticeable UI lag during search when there are many skills.

**Suggestion:** 
Separate the reactivity into two memos: one for computing the base list of `skills()` (which only updates when the underlying skills data changes), and another for handling the dynamic "Install" option that reacts to `filter()` changes.

```typescript
  const baseOptions = createMemo<TuiDialogSelectOption<string>[]>(() => {
    const list = skills() ?? []
    const maxWidth = Math.max(0, ...list.map((s) => s.name.length))
    return list.map((skill) => {
      const tools = detectToolReferences(skill.content)
      const category = SKILL_CATEGORIES[skill.name] ?? "Other"
      const desc = skill.description?.replace(/\s+/g, " ").trim()
      return {
        title: skill.name.padEnd(maxWidth),
        description: desc,
        footer: tools.length > 0 ? tools.map((t) => `\u25b6 ${t}`).join("  ") : undefined,
        value: skill.name,
        category,
      } satisfies TuiDialogSelectOption<string>
    })
  })

  const options = createMemo<TuiDialogSelectOption<string>[]>(() => {
    const items = baseOptions()
    const q = filter().trim()
    if (classifyInstallSource(q) !== null) {
      const installOption: TuiDialogSelectOption<string> = {
        title: `Install ${q}`,
        description: "Press Enter to install from this GitHub repo, URL, or path",
        footer: undefined,
        value: INSTALL_ACTION_VALUE,
        category: "Install",
      }
      return [installOption, ...items]
    }
    return items
  })
```

packages/opencode/src/plugin/tui/altimate/skill-ops.tsx
```diff
@@ -143,11 +169,13 @@ async function installSkillDirect(
...
-  ) {
+  // Classify the source. Anything the classifier recognises (github-url / owner-repo /
+  // absolute-path) takes the corresponding branch. Anything it doesn't (relative path,
+  // `~`, bare identifier) falls through to the path branch below and is resolved against
+  // cwd — that's the historical behaviour, preserved as a fallback for callers that
+  // reach the installer without going through the UI's install-preview.
+  const kind = classifyInstallSource(normalized)
+  if (kind === "github-url" || kind === "owner-repo") {
     const url = normalized.startsWith("http") ? normalized : `https://github.com/${normalized}.git`
     const label = url.replace(/https?:\/\/github\.com\//, "").replace(/\.git$/, "")
```
dev-punia-altimate, line 205: **[🟠 MEDIUM]** **Bug: Inconsistent string normalization**

The `classifyInstallSource` function accepts leading/trailing whitespaces and strips `.git` suffixes (e.g., `" owner/repo "`), returning a valid `kind`. However, `installSkillDirect` relies on the unmodified `normalized` string.
If `normalized` has spaces or an existing `.git` suffix (e.g., `" owner/repo "` or `"owner/repo.git"`), it passes the `if` check but generates a malformed URL like `https://github.com/ owner/repo .git` or `https://github.com/owner/repo.git.git`.

**Suggestion:**
Ensure the string used to construct the URL matches the cleaned format recognized by the classifier.

```typescript
  const kind = classifyInstallSource(normalized)
  if (kind === "github-url" || kind === "owner-repo") {
    const cleaned = normalized.trim().replace(/\.+$/, "").replace(/\.git$/, "")
    const url = cleaned.startsWith("http") ? cleaned : `https://github.com/${cleaned}.git`
    const label = url.replace(/https?:\/\/github\.com\//, "").replace(/\.git$/, "")
```

packages/opencode/src/plugin/tui/altimate/skill-ops.tsx
```diff
@@ -143,12 +178,19 @@ async function installSkillDirect(
...
+  // cwd — that's the historical behaviour, preserved as a fallback for callers that
+  // reach the installer without going through the UI's install-preview.
+  const kind = classifyInstallSource(normalized)
+  if (kind === "github-url" || kind === "owner-repo") {
+    // Build the clone URL from the *normalized* source so a `.git` suffix or trailing
+    // dot in the typed query doesn't produce `owner/repo.git.git` (bug reported by
+    // OpenCodeReview): the classifier tolerates those shapes but the raw `normalized`
+    // still carries them until stripped.
+    const cleaned = normalizeInstallSource(normalized)
+    const url = cleaned.startsWith("http") ? cleaned : `https://github.com/${cleaned}.git`
```
dev-punia-altimate: **[🟠 MEDIUM]** There are two potential issues here:

1. **Stripping `.git` from HTTP URLs**: Using `cleaned` for explicit HTTP URLs inadvertently removes the `.git` suffix if the user explicitly provided one (e.g., `https://git.example.com/repo.git` becomes `https://git.example.com/repo`). While GitHub tolerates this, many self-hosted Git servers require the suffix and will fail to clone.
2. **Conflating logic**: Relying on `cleaned.startsWith("http")` can incorrectly match an `owner-repo` shorthand if the owner happens to be `http` or `https` (e.g., `http/my-repo`), resulting in an invalid URL instead of correctly cloning `https://github.com/http/my-repo.git`.

You can fix both by leveraging the already computed `kind` variable to avoid false positives and preserving the original suffix for HTTP URLs while still cleaning up any trailing dots.

**Suggested change:**
```suggestion
    const cleaned = normalizeInstallSource(normalized)
    const url = kind === "github-url" 
      ? normalized.trim().replace(/\.+$/, "") 
      : `https://github.com/${cleaned}.git`
```

dev-punia-altimate (commented):

- 0 without a line anchor

#### 1. `packages/opencode/src/plugin/tui/altimate/skill-ops.tsx` (L192-L193)

**[🟠 MEDIUM]** There are two potential issues here:

1. **Stripping `.git` from HTTP URLs**: Using `cleaned` for explicit HTTP URLs inadvertently removes the `.git` suffix if the user explicitly provided one (e.g., `https://git.example.com/repo.git` becomes `https://git.example.com/repo`). While GitHub tolerates this, many self-hosted Git servers require the suffix and will fail to clone.
2. **Conflating logic**: Relying on `cleaned.startsWith("http")` can incorrectly match an `owner-repo` shorthand if the owner happens to be `http` or `https` (e.g., `http/my-repo`), resulting in an invalid URL instead of correctly cloning `https://github.com/http/my-repo.git`.

You can fix both by leveraging the already computed `kind` variable to avoid false positives and preserving the original suffix for HTTP URLs while still cleaning up any trailing dots.

**Suggested change:**
```suggestion
    const cleaned = normalizeInstallSource(normalized)
    const url = kind === "github-url" 
      ? normalized.trim().replace(/\.+$/, "") 
      : `https://github.com/${cleaned}.git`
```

</details>

dev-punia-altimate:

No supported files changed.

ralphstodomingo:
> [!NOTE]
> **AI-assisted code review.** Produced by Claude (Opus 4.8) running an adversarial multi-agent
> workflow against `b8e7e64`, then verified live in a terminal, posted from @ralphstodomingo's
> account. Read the findings as claims to check, not as vetted conclusions.
>
> **Method:** a recon pass first pinned the `DialogSelect` / `DialogPrompt` / `dialog.replace` API
> contracts this PR depends on. Five independent finders then swept separate lenses (does-it-work
> end-to-end, classifier correctness, module-state lifecycle, UX + the perf claim, test quality).
> Every candidate finding faced three adversarial verifiers instructed to refute it, majority-refute
> dropping it — 2 of 5 candidates were killed that way. Finally the TUI was **driven live in a PTY**
> (`pyte` screen emulation) against `b8e7e64`, so the behavioral findings below are reproduced
> end-to-end, not inferred from reading.

**Verdict: mergeable with fixes.** The core design is sound — one exported `classifyInstallSource`
(`skill-ops.tsx:55`) consumed by both the installer (`:186`) and the list's synthetic-row gate
(`:638`) genuinely removes the UI/installer drift the old `q.includes("/")` check caused, and the
sentinel is correctly filtered out of `onMove` (`:662`) and `showActions` (`:721`) so a highlighted
Install row can't leak into the action picker. The non-mutating memo (`return [installOption, ...items]`,
`:646`) is right for Solid's double-eval. **Live drive confirms the feature works**: the synthetic
row renders and survives `DialogSelect`'s internal filter for `anthropics/skills` and
`httpie/httpie`, shows no row for `dbt-develop`, and prefills the install dialog. Four issues remain
— one correctness bug, one merge-drop guard gap, a comment inaccuracy, and a slash-command collision
that makes the feature unreachable by the documented path.

## Findings

### medium — `installSkillDirect` discards the computed `kind` and re-derives the clone-URL branch from a bare `http` prefix test

`packages/opencode/src/plugin/tui/altimate/skill-ops.tsx:193`

`kind` is computed at `:186` and is the exact, correct discriminant, but `:193` throws it away:

```ts
const url = cleaned.startsWith("http") ? cleaned : `https://github.com/${cleaned}.git`
```

The classifier tests the scheme-qualified prefixes `http://` / `https://` (`:58`); `:193` tests
only the bare substring `http`. `OWNER_REPO_REGEX` (`:45`) admits owners starting with those four
letters, so `httpie/httpie` classifies as `owner-repo` at `:186`, takes the "already a URL" path at
`:193`, and `https://github.com/` is never prepended.

**Reproduced live.** Type `httpie/httpie` into the fork's `/skills` dialog, Enter, Enter:

```
Cloning httpie/httpie...
Install failed: Failed to clone: fatal: repository 'httpie/httpie' does not exist
```

Control — same `owner/repo` shape, non-`http` owner — **succeeds**, isolating the cause to the
prefix drift rather than anything environmental:

```
Cloning anthropics/skills...  Found 18 skill(s). Installing...  Installed 1/18: …
```

Scope is genuinely small — it needs an owner literally prefixed `http` (`httpie/httpie`,
`http-party/http-server`, `httpwg/http-extensions`) and the shorthand form (pasting the full URL
works), and it fails loudly. The flaw predates the PR (the removed line used
`normalized.startsWith("http")`), but `:193` is a line this PR rewrote, the new row at `:638-646`
now advertises exactly these inputs as installable, and it's precisely the classifier/installer
drift the header comment at `:37-43` says the shared classifier exists to prevent — the right
answer is in hand two lines up and unused.

**Fix (verified live + all 32 tests still pass):**

```ts
const url = kind === "github-url" ? cleaned : `https://github.com/${cleaned}.git`
```

With this applied, `httpie/httpie` now clones and fails only because it genuinely isn't a skills
repo — the correct outcome:

```
Cloning httpie/httpie...  Scanning for skills...
Install failed: No SKILL.md files found in httpie/httpie
```

Add a classifier test with an `http`-prefixed owner to pin it.

### medium — `/skills` + Enter opens the *upstream* dialog, not this one, so the new Install row is unreachable by the documented path

`packages/tui/src/component/prompt/index.tsx:559-562` and `packages/opencode/src/plugin/tui/altimate/skill-ops.tsx:731-736`

Two commands register the same slash trigger:
- `prompt.skills` — upstream, category "Prompt"
- `altimate.skill.list` — this PR's plugin, category "Altimate"

Driving it live: typing `/skills` and pressing Enter opens the **upstream** dialog
(`title="Skills"`, `placeholder="Search skills..."`, `packages/tui/src/component/dialog-skill.tsx:35`),
where typing a URL yields "No results found." You must press **Down** first to reach this PR's
dialog (the one with `placeholder="Search skills, or type a repo/URL…"`, `skill-ops.tsx:655`).

**This is pre-existing on `main`** — both `slashName: "skills"` registrations already exist there —
so it is explicitly *not* a defect this PR introduces. Flagging it because it undercuts the PR's
headline goal: the smoke plan in the description ("`/skills` → type
`https://github.com/anthropics/skills.git` → Enter opens install dialog") **cannot work as written**;
plain `/skills`+Enter lands on the upstream dialog and produces exactly the "filters to zero results,
no obvious next action" dead end this PR set out to remove. Worth a separate ticket, but this PR's
discoverability win doesn't land for anyone reaching the list via `/skills`.

### medium — the merge-drop guard doesn't cover the block that makes the Install row exist

`packages/opencode/test/upstream/fork-feature-guards.test.ts:206-215`

Deleting the synthetic-row block from the `options` memo (`skill-ops.tsx:637-647`) removes the
feature entirely — the row never renders and the `onSelect` routing at `:666` becomes dead code —
yet **all 18 guards still pass** (confirmed by real mutation test, not simulation). Two of the five
assertions are inert as written: `/classifyInstallSource/` (`:209`) is satisfied by the definition
at `:55` and the installer call at `:186`; `/INSTALL_ACTION_VALUE/` (`:210`) by the const at `:69`.
The proximity regexes at `:212`/`:214-215` do have teeth (they pin `onSelect` routing and prefill
props), so this isn't zero detection — the gap is specific to the memo block, which is both the
load-bearing site and the one most likely to get clobbered during an upstream re-extraction.

**Fix — one added assertion, verified in the real runner (18 pass on correct source; 17 pass / 1 fail
when the feature is deleted):**

```ts
expect(skill).toMatch(/classifyInstallSource\(q\)[\s\S]{0,300}value: INSTALL_ACTION_VALUE/)
```

(The gap between those two anchors is 247 chars, so the bound must be ≥250 — a tighter `{0,200}`
fails against the correct source too.)

### low — comments cite `dbt/snowflake` as the pinned regression, but it classifies as `owner-repo` and is never asserted

`packages/opencode/test/altimate/skill-install-classifier.test.ts:7, 37-41` and `packages/opencode/src/plugin/tui/altimate/skill-ops.tsx:40-43`

The test header (`:4-8`), the inline comment at `:38`, and `skill-ops.tsx:41-43` all name
`dbt/snowflake` as the query the classifier now rejects ("search terms, relative paths, `~`"). But
`classifyInstallSource("dbt/snowflake")` returns `"owner-repo"` — `OWNER_REPO_REGEX` (`:45`) matches
any clean two-segment string. The test titled "rejects multi-slash paths…" only asserts the
three-segment `owner/repo/subpath` and `dbt/snowflake/thing` (`:40-41`), which fail for an unrelated
reason (the second slash). The one case named twice as motivating is the one case not asserted and
not actually rejected.

This is a documentation defect, not functional: accepting clean `owner/repo` is the feature (a
two-segment string is indistinguishable from a two-token search without a network call), and skill
names can't contain a slash (`skill-ops.tsx:125`), so no search result is hijacked. **Fix:** correct
the comments to say two-segment inputs are intentionally treated as `owner/repo` shorthand, reserve
the "search terms" rejection claim for the three-segment / relative / `~` cases the code actually
rejects, and optionally pin `expect(classifyInstallSource("dbt/snowflake")).toBe("owner-repo")`.

## What I could not verify

- Only the `owner/repo` and github-URL install paths were driven live. The **absolute-path** branch
  and the `ctrl+n`/`ctrl+i` prefill plumbing were exercised by reading only.
- The module-scope `currentFilter` lifecycle (potential stale-text leak across dialogs) was reviewed
  statically; I didn't construct a live sequence that trips it.
- No claim either way on the `baseOptions`/`options` per-keystroke perf split — structurally correct
  (`baseOptions` reads `skills()`, `options` reads `filter()`) but unmeasured.

sahrizvi:
Thanks @dev-punia-altimate (OpenCodeReview) and @ralphstodomingo (Claude Opus 4.8 adversarial-multi-agent pass) — findings addressed in `eaa78c227`.

## Addressed

### `httpie/httpie` misroutes as a URL (MEDIUM — both reviewers, ralph reproduced live in PTY)
`installSkillDirect` used `cleaned.startsWith("http")` to route between "already a URL" and "build a github.com URL". Owner names literally starting with `http` (`httpie/httpie`, `http-party/http-server`, `httpwg/http-extensions`) classify correctly as `owner-repo` but tripped the prefix test and got sent to `git clone httpie/httpie`, which fails.

Fix: branch on the already-computed `kind` from `classifyInstallSource`, not on a prefix probe. Bonus win for @dev-punia-altimate's related concern — the `github-url` branch now preserves the trailing `.git` the user typed (matters for self-hosted git servers that require the exact suffix form), while the `owner-repo` branch continues to normalize it away so `owner/repo.git` doesn't produce `owner/repo.git.git`.

```ts
const url =
  kind === "github-url"
    ? normalized.trim().replace(/\.+$/, "")
    : `https://github.com/${normalizeInstallSource(normalized)}.git`
```

Regression pinned by three new classifier assertions:
- `expect(classifyInstallSource("httpie/httpie")).toBe("owner-repo")`
- `expect(classifyInstallSource("http-party/http-server")).toBe("owner-repo")`
- `expect(classifyInstallSource("httpwg/http-extensions")).toBe("owner-repo")`

### Merge-drop guard didn't cover the memo block that creates the Install row (MEDIUM — ralph, mutation-verified)
Deleting the synthetic-row block from the `options` memo removed the feature entirely but all 18 fork-guards still passed — two of the assertions were inert (the classifier symbol matches the installer call, `INSTALL_ACTION_VALUE` matches the const decl).

Added a proximity-anchored assertion tying `classifyInstallSource(q)` to `value: INSTALL_ACTION_VALUE`. The `{0,300}` bound is deliberately tight — a real deletion collapses the gap to zero and the match fails loudly on merge-drop:

```ts
expect(skill).toMatch(/classifyInstallSource\(q\)[\s\S]{0,300}value: INSTALL_ACTION_VALUE/)
```

### `dbt/snowflake` mischaracterized in docstrings (LOW — ralph)
The comments claimed the classifier rejects `dbt/snowflake` as a "search term", but the classifier returns `"owner-repo"` for any clean two-segment string — that's intentional (skill names can't contain a slash per the `^[a-z][a-z0-9]*(-[a-z0-9]+)*$` grammar, so no real search result is hijacked). Corrected the comments and pinned the actual behaviour:

```ts
expect(classifyInstallSource("dbt/snowflake")).toBe("owner-repo")
expect(classifyInstallSource("anthropics/skills")).toBe("owner-repo")
```

## Not addressed here

### `/skills` slash-command collision (MEDIUM — ralph, pre-existing on main)
`prompt.skills` (upstream, `packages/tui/src/component/prompt/index.tsx:559-562`) and `altimate.skill.list` (this plugin, `skill-ops.tsx:731-736`) both register `slashName: "skills"`. `/skills`+Enter opens the upstream dialog by default; users have to press Down to reach this PR's dialog.

Confirmed pre-existing on `main` (both registrations were there before this branch), so out of scope for this PR. Already noted for consolidation follow-up with product engineering.

## Local validation

- `bun run typecheck` — clean
- Classifier + fork-guard suites — **34 pass / 0 fail / 100 expects**
- `test/session` — **731 pass / 0 fail / 1636 expects**

ralphstodomingo:
> [!NOTE]
> **AI-assisted verification pass.** Claude (Opus 4.8) again, this time re-checking my own prior review
> against `eaa78c2` rather than looking for new findings. Seven parallel investigations (one per finding,
> plus your validation numbers, plus the gaps my first pass admitted it hadn't closed), each followed by
> an adversarial checker instructed to refute it and re-run the decisive commands. Every claim below that
> says "verified" was produced by **executing** code in a clean worktree at `eaa78c2` — real `git clone`
> attempts, real mutation of the source with the suite re-run each time — not by reading. Three candidate
> findings were killed by the checkers and are not listed.

**Verdict: three of my four findings are genuinely fixed; the fourth is a defensible deferral. One
one-line correction is worth making before merge, and three claims in the response comment need
walking back.** Details below, plus a correction of my own.

## Confirmed fixed

**F1 — `httpie/httpie` misroute.** Dead. Driving the real `installSkillDirect` URL construction over both
commits:

```
"httpie/httpie"           | eaa78c2: https://github.com/httpie/httpie.git   | b8e7e64: httpie/httpie
"http-party/http-server"  | eaa78c2: https://github.com/http-party/…​.git    | b8e7e64: http-party/http-server
```

Branching on `kind` was the right call.

**F2 — merge-drop guard.** The assertion is in at `fork-feature-guards.test.ts:217`, and it works for the
mutation it was written for: deleting the synthetic-row memo block from `skill-ops.tsx` gives
`17 pass / 1 fail`, failing on exactly that line. Re-verified by hand, not simulated.

**F3 — `dbt/snowflake` docstrings.** All three asks met — comments corrected at `skill-ops.tsx:40-45`,
test title narrowed to "rejects three-segment paths", explicit `toBe("owner-repo")` pins added and
passing. This one your comment describes accurately.

**F4 — `/skills` collision.** Your "pre-existing on `main`" claim is correct: both `slashName: "skills"`
registrations (`packages/tui/src/component/prompt/index.tsx:562`, `skill-ops.tsx:747`) exist on `main`
and at the merge base, and nothing in this PR touches ordering. Deferring is fine by me.

## Three claims in the response comment that don't hold

**1. The self-hosted `.git` preservation is unreachable — and on the github.com side it's a regression.**

The only caller pre-strips the suffix before the installer ever sees it (`skill-ops.tsx:430`):

```ts
const source = rawSource.trim().replace(/\.+$/, "").replace(/\.git$/, "")
```

So a typed `https://git.selfhosted.invalid/team/repo.git` produces the identical URL on `b8e7e64` and
`eaa78c2` — `https://git.selfhosted.invalid/team/repo`, suffix already gone. Verified:

```
"https://git.selfhosted.invalid/team/repo.git" | eaa78c2: …/team/repo | b8e7e64: …/team/repo
```

The one input where a `.git` *does* reach that branch is the github.com case where `ghWebMatch`
(`:179-181`) appended it itself — and there, preserving it doubles the suffix:

```
"https://github.com/anthropics/skills.git.git" | eaa78c2: …/skills.git.git | b8e7e64: …/skills.git
```

`b8e7e64` handled that paste; `eaa78c2` doesn't (`remote: Repository not found`). It matches base `main`'s
behaviour, so it's a revert of an in-PR improvement rather than a regression against `main`, and it needs
a doubled-suffix paste to trigger — but the fix is one token:

```ts
kind === "github-url" ? normalizeInstallSource(normalized) : `https://github.com/${normalizeInstallSource(normalized)}.git`
```

I ran that variant across the same input set: httpie stays fixed, every other case matches `b8e7e64`.
Worth taking, along with dropping the self-hosted rationale from the comment at `:196-200`.

**2. "Regression pinned by three new classifier assertions" — it isn't pinned.**

`classifyInstallSource` is byte-identical between `b8e7e64` and `eaa78c2`; it already returned
`"owner-repo"` for all three `http*` owners *while the bug was live*. The bug was in the URL expression,
and nothing in `packages/opencode/test/` touches it. Restoring the exact pre-fix expression and re-running
both suites:

```
34 pass / 0 fail / 100 expect() calls
```

Byte-identical to baseline — zero detection. The new assertions are fine as documentation of intent, but
the thing that would catch a re-introduction is a test on the URL construction itself.

**3. The `{0,300}` rationale states its mechanism backwards, and the guard is presence-only.**

A zero gap *matches* `{0,300}` — the guard fires because both anchors disappear, not because the gap
collapses. And the guard only catches deletion of the block. Two realistic re-extraction shapes leave the
feature completely dead with the suite green at `18 pass / 0 fail`:

- `return [installOption, ...items]` (`skill-ops.tsx:657`) → `return items`
- `options={options()}` (`:667`) → `options={baseOptions()}`

Both verified by hand. Neither is caught by typecheck either. If it's worth hardening, pinning
`/return \[installOption, \.\.\.items\]/` and `/options=\{options\(\)\}/` covers both cheaply.

**Correction to my own review while I'm here:** I said the inter-anchor gap was 247 chars and the bound
had to be ≥250. It's 223, so `{0,300}` carries 77 chars of slack — the comment at `:211-216` calling that
"deliberately tight" overstates it in the other direction.

## On CI

`gh pr checks 1005` shows the required **TypeScript** check red at this head, and the response comment
doesn't mention it. To be fair on both sides: **it isn't this PR's fault** — the same annotation
(`packages/opencode/test/cli/tui/command.test.ts:10`, expected string rewritten to a `file://` URL) fires
on `main` pushes, e.g. runs [29796413752](https://github.com/AltimateAI/altimate-code/actions/runs/29796413752)
and [29852163070](https://github.com/AltimateAI/altimate-code/actions/runs/29852163070). And "typecheck
clean" vs "TypeScript ✗" is not a contradiction: that job runs no typecheck at all — its substantive step
is `bun test` (`.github/workflows/ci.yml:152`); `bun turbo typecheck` only appears in `release.yml`. The
job name invites exactly this misreading. Worth one sentence in the PR so a reviewer doesn't have to
re-derive it.

## Smaller things, none blocking

- **The PR body's smoke plan documents a path that dead-ends.** All five bullets read `/skills` → type →
  Enter, with no Down step and no caveat — but `/skills`+Enter lands on the upstream dialog, where typing a
  URL gives "No results found". QA following it verbatim concludes the feature is broken. Given F4 is being
  deferred, the plan should say `/skills` → **Down** → Enter.
- **`ctrl+n` prefills Create Skill with a value that can never validate.** `skill-ops.tsx:768` forwards
  `currentFilter()` unconditionally, so a URL in the box opens the create prompt pre-filled with it, and
  confirming fails the name grammar at `:128`. A `classifyInstallSource(q) === null` gate on the create
  path fixes it.
- **Windows drive-letter paths surface the row on POSIX.** `classifyInstallSource("C:\\Users\\me\\skills")`
  returns `absolute-path` platform-agnostically, while the install branch uses `path.isAbsolute` (`false`
  there) — so the row is offered and Enter resolves against cwd and fails. Pre-existing in the classifier,
  but the new row is what makes it reachable.
- **The prefill guards don't pin the load-bearing hop.** Deleting `value={props.initialValue}` from
  `skill-ops.tsx:359` and `:416` kills prefill entirely and all six install-row regexes still pass. Same
  class of gap as F2, one layer down.

## What I could not verify

- The `ctrl+i` / `ctrl+n` keystrokes themselves — every prefill path was exercised through `run()` or
  Enter-on-row, never a real key event. (The comment at `:646-647` notes `ctrl+i` is byte `0x09` = Tab on
  the wire, so the literal binding may not deliver at all.)
- The `baseOptions`/`options` per-keystroke perf split — structurally correct, still unmeasured. No claim
  either way.
- The root cause of the CI `file://` rewrite. Confirmed CI-only and pre-existing on `main`; it doesn't
  reproduce on bun 1.3.14 locally, so the mechanism is inferred from the log.
- Whether anything outside the three touched files regressed — no clean local full-suite baseline exists.

**Bottom line: nothing here blocks the feature.** Take the one-line `normalizeInstallSource` correction on
the `github-url` branch, fix the smoke plan in the body, and reword the two comments that claim more than
the code does — then this is good to go from my side.

ralphstodomingo (approved):

PR #1007, opened 2026-07-16
fix: [AI-7743] Altimate Code chat — MCP servers not getting added (question hang)

saravmajestic:
## Summary

Fixes **[AI-7743](https://altimateai.atlassian.net/browse/AI-7743)** — in Altimate Code chat, `/discover-and-add-mcps` showed the "which MCP / what scope" question, but after the user answered it the chat sat on **"Thinking…"** forever and the MCP server was never added.

## Root cause

After the **v1.17.9 upstream merge** (effect@4), `src/question/index.ts` is bundled as **two separate module instances**. Proven: a module-scoped id logged from the `question` tool's `ask()` differs from the one logged in the HTTP `reply()` route; normalizing every importer to `@/question` did **not** make Bun dedupe it. Each copy ran its own `makeRuntime` runtime, which broke the flow in two independent ways:

1. **State split.** The `question` tool registered the pending `Deferred` in one copy's `InstanceState` pending-map, while `POST /question/:id/reply` looked it up in the **other** copy's empty map. The reply returned `NotFoundError` (HTTP 500) and the `Deferred` never resolved, so the agent loop blocked forever → **"Thinking…"**. (`GET /question` returned `[]` for the same reason.)
2. **Event never reached the IDE webview.** `question.asked` was published only via `EventV2Bridge` → `GlobalBus`, but the `/event` SSE stream the webview subscribes to is fed by the Bus **wildcard PubSub** (`Bus.publish`) — a different channel. So the webview's `pendingQuestions` stayed empty, the answer card's Submit resolved `requestId: undefined`, and submitting did nothing.

Either half alone produces the reported "answer the question → nothing happens".

## Fix (all in `question/index.ts` + import cleanup)

- **State:** anchor the pending-question registry on `globalThis` (keyed by instance directory) so all module copies share one map. Restore per-instance cleanup via `registerDisposer` so entries don't leak across instances/tests.
- **Events:** publish `question.asked` / `replied` / `rejected` via **`Bus.publish`** (added `BusEvent` mirrors) so they reach `/event` like every other webview-visible event.
- Normalize the remaining relative `../question` imports to `@/question`.

## Verification

End-to-end in Docker code-server (VS Code extension + real UI):
- `/discover-and-add-mcps` → question card renders → answer **Yes / Project** → **✅ datamate added** to `.altimate-code/altimate-code.json` (`enabled: true`) and the chat shows the success summary.
- Confirmed the webview now receives the event (`pendingQuestions` populated, was `[]`) and the reply resolves (`HTTP 200`, `message=replied`).
- `55` question unit tests pass (`test/question`, `test/tool/question`, `test/release-validation/question-937*`, `test/cli/run/question.shared`).

No changes were needed in the `vscode-altimate-mcp-server` extension.

[AI-7743]: https://altimateai.atlassian.net/browse/AI-7743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Fixes AI-7743: answering the MCP add question in `/discover-and-add-mcps` no longer hangs. Question state is shared across bundled module copies, and events reach the webview without blocking resolution so the MCP server is added.

- **Bug Fixes**
  - Share the pending-question registry on `globalThis` (keyed by instance directory) with per-instance cleanup via `registerDisposer`, so `ask()` and HTTP `reply()` resolve the same request; cleanup failures are logged.
  - Mirror `question.asked` / `replied` / `rejected` to `/event` via `Bus.publish` using `BusEvent` with `zod` schemas; mirrors are best-effort and kept off the Deferred critical path—`reply()`/`reject()` settle first, then publish.

<sup>Written for commit 0b5ecd12fe12aa2dcf2c97b16169281f3cefa4d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1007?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Forwarded question lifecycle events (**asked**, **replied**, **rejected**) to the IDE webview event stream.
  * Standardized question handling so it’s shared consistently across modules within the same running application instance.

* **Bug Fixes**
  * Pending questions are now properly cleaned up when an instance is disposed.
  * Any still-pending questions are automatically rejected instead of remaining unresolved.

### DIFF
--- a/packages/opencode/src/question/index.ts
+++ b/packages/opencode/src/question/index.ts
@@ -10,6 +10,7 @@ import { makeRuntime } from "@/effect/run-service"
 import { registerDisposer } from "@/effect/instance-registry"
 import { Bus } from "@/bus"
 import { BusEvent } from "@/bus/bus-event"
+import { zod } from "@/util/effect-zod"
 import z from "zod"
 // altimate_change end
 
@@ -105,29 +106,31 @@ export const Event = {
 // never reached the webview → `pendingQuestions` stayed empty → the mcp-add
 // question card had no request id to reply with → "submit does nothing".
 // Publish these via `Bus.publish` too so they reach /event like every other
-// webview-visible event. (Schemas are loose — Bus.publish does not re-validate;
-// they exist for the `type` string and typing.)
-const BusAsked = BusEvent.define(
-  "question.asked",
-  z.object({
-    id: QuestionID.zod,
-    sessionID: SessionID.zod,
-    questions: z.array(z.any()),
-    tool: z.object({ messageID: MessageID.zod, callID: z.string() }).optional(),
-  }),
-)
-const BusReplied = BusEvent.define(
-  "question.replied",
-  z.object({
-    sessionID: SessionID.zod,
-    requestID: QuestionID.zod,
-    answers: z.array(z.array(z.string())),
-  }),
-)
-const BusRejected = BusEvent.define(
-  "question.rejected",
-  z.object({ sessionID: SessionID.zod, requestID: QuestionID.zod }),
-)
+// webview-visible event.
+//
+// Reuse the structured Effect schemas (via the `zod` adapter) so the generated
+// `/event` OpenAPI payloads match the SDK's typed shapes rather than `any`.
+//
+// Note: `EventV2Bridge.listen` already forwards these EventV2 events to GlobalBus
+// and `Bus.publish` re-emits to GlobalBus too, so `/global/event` consumers see
+// each question event twice (different top-level ids). This is intentional and
+// verified harmless in-repo — TUI `sync.tsx` reconciles by request id,
+// `notifications.ts` dedupes by id, and the trace consumer ignores question
+// events. External subscribers that don't dedupe are the only residual concern.
+const BusAsked = BusEvent.define("question.asked", zod(Request))
+const BusReplied = BusEvent.define("question.replied", zod(Replied))
+const BusRejected = BusEvent.define("question.rejected", zod(Rejected))
+
+// Best-effort Bus mirror. A fire-and-forget /event notification must NEVER be
+// able to abort core question settlement: `Effect.promise` turns a promise
+// rejection into an unrecoverable fiber defect, and on the Deferred critical
+// path that would skip `Deferred.succeed`/`Deferred.fail` and re-hang the tool
+// on "Thinking…" — the exact failure this PR fixes. Recover any cause to a
+// logged warning so publication can never block settlement.
+const mirror = <D extends BusEvent.Definition>(def: D, properties: z.output<D["properties"]>) =>
+  Effect.promise(() => Bus.publish(def, properties)).pipe(
+    Effect.catchCause((cause) => Effect.logWarning("question bus mirror failed", { type: def.type, cause })),
+  )
 // altimate_change end
 
 export class RejectedError extends Schema.TaggedErrorClass<RejectedError>()("QuestionRejectedError", {}) {
@@ -202,7 +205,11 @@ export const layer = Layer.effect(
       if (!map) return
       pendingByDir.delete(directory)
       for (const { deferred } of map.values()) {
-        await Effect.runPromise(Deferred.fail(deferred, new RejectedError())).catch(() => {})
+        await Effect.runPromise(
+          Deferred.fail(deferred, new RejectedError()).pipe(
+            Effect.catchCause((cause) => Effect.logWarning("question cleanup failed on dispose", { cause })),
+          ),
+        )
       }
     })
     yield* Effect.addFinalizer(() => Effect.sync(off))
@@ -229,11 +236,16 @@ export const layer = Layer.effect(
       }
       pending.set(id, { info, deferred })
       yield* events.publish(Event.Asked, info)
-      // altimate_change start — also publish on the Bus wildcard so the IDE webview
+      // altimate_change start — also mirror on the Bus wildcard so the IDE webview
       // (subscribed to /event) receives question.asked and can answer the card.
-      yield* Effect.promise(() =>
-        Bus.publish(BusAsked, { id, sessionID: input.sessionID, questions: [...input.questions], tool: input.tool }),
-      )
+      // Best-effort: a publish failure must not abort ask() before it registers
+      // the cleanup finalizer below (see `mirror`).
+      yield* mirror(BusAsked, {
+        id,
+        sessionID: input.sessionID,
+        questions: [...input.questions],
+        tool: input.tool,
+      })
       // altimate_change end
 
       return yield* Effect.ensuring(
@@ -263,16 +275,15 @@ export const layer = Layer.effect(
         requestID: existing.info.id,
         answers: input.answers.map((a) => [...a]),
       })
-      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
-      yield* Effect.promise(() =>
-        Bus.publish(BusReplied, {
-          sessionID: existing.info.sessionID,
-          requestID: existing.info.id,
-          answers: input.answers.map((a) => [...a]),
-        }),
-      )
-      // altimate_change end
       yield* Deferred.succeed(existing.deferred, input.answers)
+      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients,
+      // AFTER settling the Deferred and best-effort so a publish failure can't re-hang ask().
+      yield* mirror(BusReplied, {
+        sessionID: existing.info.sessionID,
+        requestID: existing.info.id,
+        answers: input.answers.map((a) => [...a]),
+      })
+      // altimate_change end
     })
 
     const reject = Effect.fn("Question.reject")(function* (requestID: QuestionID) {
@@ -290,12 +301,11 @@ export const layer = Layer.effect(
         sessionID: existing.info.sessionID,
         requestID: existing.info.id,
       })
-      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
-      yield* Effect.promise(() =>
-        Bus.publish(BusRejected, { sessionID: existing.info.sessionID, requestID: existing.info.id }),
-      )
-      // altimate_change end
       yield* Deferred.fail(existing.deferred, new RejectedError())
+      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients,
+      // AFTER settling the Deferred and best-effort so a publish failure can't strand it.
+      yield* mirror(BusRejected, { sessionID: existing.info.sessionID, requestID: existing.info.id })
+      // altimate_change end
     })
 
     const list = Effect.fn("Question.list")(function* () {
--- a/packages/opencode/src/question/index.ts
+++ b/packages/opencode/src/question/index.ts
@@ -7,6 +7,10 @@ import { EventV2Bridge } from "@/event-v2-bridge"
 import { EventV2 } from "@opencode-ai/core/event"
 // altimate_change start — makeRuntime for the restored Promise wrappers (see bottom of file)
 import { makeRuntime } from "@/effect/run-service"
+import { registerDisposer } from "@/effect/instance-registry"
+import { Bus } from "@/bus"
+import { BusEvent } from "@/bus/bus-event"
+import z from "zod"
 // altimate_change end
 
 // Schemas — these are pure data; nothing checks class identity (see PR
@@ -93,6 +97,39 @@ export const Event = {
   Rejected: EventV2.define({ type: "question.rejected", schema: Rejected.fields }),
 }
 
+// altimate_change start — BusEvent mirrors of the question events.
+//
+// The EventV2 `Event` defs above publish to GlobalBus/EventV2 consumers only.
+// The IDE webview subscribes to the `/event` SSE route, which is fed by the Bus
+// *wildcard PubSub* (`Bus.publish`), a different channel. So `question.asked`
+// never reached the webview → `pendingQuestions` stayed empty → the mcp-add
+// question card had no request id to reply with → "submit does nothing".
+// Publish these via `Bus.publish` too so they reach /event like every other
+// webview-visible event. (Schemas are loose — Bus.publish does not re-validate;
+// they exist for the `type` string and typing.)
+const BusAsked = BusEvent.define(
+  "question.asked",
+  z.object({
+    id: QuestionID.zod,
+    sessionID: SessionID.zod,
+    questions: z.array(z.any()),
+    tool: z.object({ messageID: MessageID.zod, callID: z.string() }).optional(),
+  }),
+)
+const BusReplied = BusEvent.define(
+  "question.replied",
+  z.object({
+    sessionID: SessionID.zod,
+    requestID: QuestionID.zod,
+    answers: z.array(z.array(z.string())),
+  }),
+)
+const BusRejected = BusEvent.define(
+  "question.rejected",
+  z.object({ sessionID: SessionID.zod, requestID: QuestionID.zod }),
+)
+// altimate_change end
+
 export class RejectedError extends Schema.TaggedErrorClass<RejectedError>()("QuestionRejectedError", {}) {
   override get message() {
     return "The user dismissed this question"
@@ -108,9 +145,31 @@ interface PendingEntry {
   deferred: Deferred.Deferred<ReadonlyArray<Answer>, RejectedError>
 }
 
-interface State {
-  pending: Map<QuestionID, PendingEntry>
+// altimate_change start — process-global pending registry.
+//
+// The imperative `Question.ask()`/`reply()`/`list()` wrappers (bottom of file)
+// get bundled into MORE THAN ONE module instance (proven: a module-scoped id
+// differs between the tool's ask() and the HTTP route's reply()). Consistent
+// `@/question` imports did NOT dedupe them. Each copy ran its own module-scoped
+// `makeRuntime(...)` runtime with its own `InstanceState` cache, so the question
+// TOOL registered the pending Deferred in one copy's map while the HTTP reply
+// route looked it up in the OTHER copy's empty map — the Deferred never resolved
+// and the `/discover-and-add-mcps` question hung on "Thinking…" after answering.
+//
+// Anchor the registry on `globalThis` so every module copy shares one Map. Keyed
+// by instance directory so `list()` stays per-instance.
+type PendingByDir = Map<string, Map<QuestionID, PendingEntry>>
+const pendingByDir: PendingByDir = ((globalThis as Record<string, unknown>)["__altimateQuestionPending"] ??=
+  new Map<string, Map<QuestionID, PendingEntry>>()) as PendingByDir
+function pendingFor(directory: string): Map<QuestionID, PendingEntry> {
+  let map = pendingByDir.get(directory)
+  if (!map) {
+    map = new Map<QuestionID, PendingEntry>()
+    pendingByDir.set(directory, map)
+  }
+  return map
 }
+// altimate_change end
 
 // Service
 
@@ -134,31 +193,30 @@ export const layer = Layer.effect(
   Service,
   Effect.gen(function* () {
     const events = yield* EventV2Bridge.Service
-    const state = yield* InstanceState.make<State>(
-      Effect.fn("Question.state")(function* () {
-        const state = {
-          pending: new Map<QuestionID, PendingEntry>(),
-        }
 
-        yield* Effect.addFinalizer(() =>
-          Effect.gen(function* () {
-            for (const item of state.pending.values()) {
-              yield* Deferred.fail(item.deferred, new RejectedError())
-            }
-            state.pending.clear()
-          }),
-        )
-
-        return state
-      }),
-    )
+    // altimate_change start — clear a directory's pending questions when its
+    // instance is disposed/reloaded (mirrors the removed InstanceState finalizer)
+    // so entries in the process-global registry don't leak across instances.
+    const off = registerDisposer(async (directory) => {
+      const map = pendingByDir.get(directory)
+      if (!map) return
+      pendingByDir.delete(directory)
+      for (const { deferred } of map.values()) {
+        await Effect.runPromise(Deferred.fail(deferred, new RejectedError())).catch(() => {})
+      }
+    })
+    yield* Effect.addFinalizer(() => Effect.sync(off))
+    // altimate_change end
 
     const ask = Effect.fn("Question.ask")(function* (input: {
       sessionID: SessionID
       questions: ReadonlyArray<Info>
       tool?: Tool
     }) {
-      const pending = (yield* InstanceState.get(state)).pending
+      // altimate_change start — pending map lives in the globalThis registry (see top of file)
+      const directory = yield* InstanceState.directory
+      const pending = pendingFor(directory)
+      // altimate_change end
       const id = QuestionID.ascending()
       yield* Effect.logInfo("asking", { id, questions: input.questions.length })
 
@@ -171,6 +229,12 @@ export const layer = Layer.effect(
       }
       pending.set(id, { info, deferred })
       yield* events.publish(Event.Asked, info)
+      // altimate_change start — also publish on the Bus wildcard so the IDE webview
+      // (subscribed to /event) receives question.asked and can answer the card.
+      yield* Effect.promise(() =>
+        Bus.publish(BusAsked, { id, sessionID: input.sessionID, questions: [...input.questions], tool: input.tool }),
+      )
+      // altimate_change end
 
       return yield* Effect.ensuring(
         Deferred.await(deferred),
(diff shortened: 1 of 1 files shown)


packages/opencode/src/question/index.ts
```diff
@@ -197,11 +263,22 @@ export const layer = Layer.effect(
...
       })
+      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
+      yield* Effect.promise(() =>
+        Bus.publish(BusReplied, {
+          sessionID: existing.info.sessionID,
+          requestID: existing.info.id,
+          answers: input.answers.map((a) => [...a]),
+        }),
+      )
+      // altimate_change end
```
dev-punia-altimate: **[🔴 HIGH]** If `Bus.publish` throws or rejects (e.g., due to bus errors), this Effect will terminate early, and `Deferred.succeed` will never be reached. This will cause the original `ask` operation that awaits this deferred to hang indefinitely.

Consider decoupling side-effects like event mirroring by using a non-blocking fork (e.g., `Effect.fork(Effect.promise(...))`), or placing it after `Deferred.succeed`, or applying explicit error handling (e.g., `Effect.ignoreLogged` / `Effect.catchAllDefect`) so it doesn't block the core business logic.

**Suggested change:**
```suggestion
      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
      yield* Effect.fork(Effect.promise(() =>
        Bus.publish(BusReplied, {
          sessionID: existing.info.sessionID,
          requestID: existing.info.id,
          answers: input.answers.map((a) => [...a]),
        }),
      ))
      // altimate_change end
```

dev-punia-altimate (commented):

- 0 without a line anchor

#### 1. `packages/opencode/src/question/index.ts` (L266-L274)

**[🔴 HIGH]** If `Bus.publish` throws or rejects (e.g., due to bus errors), this Effect will terminate early, and `Deferred.succeed` will never be reached. This will cause the original `ask` operation that awaits this deferred to hang indefinitely.

Consider decoupling side-effects like event mirroring by using a non-blocking fork (e.g., `Effect.fork(Effect.promise(...))`), or placing it after `Deferred.succeed`, or applying explicit error handling (e.g., `Effect.ignoreLogged` / `Effect.catchAllDefect`) so it doesn't block the core business logic.

**Suggested change:**
```suggestion
      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
      yield* Effect.fork(Effect.promise(() =>
        Bus.publish(BusReplied, {
          sessionID: existing.info.sessionID,
          requestID: existing.info.id,
          answers: input.answers.map((a) => [...a]),
        }),
      ))
      // altimate_change end
```

#### 2. `packages/opencode/src/question/index.ts` (L293-L297)

**[🔴 HIGH]** Similarly, if `Bus.publish` throws or rejects here, the Effect will terminate early, preventing `Deferred.fail` from being executed and leaving the pending `ask` operation hanging indefinitely.

Consider placing this in a non-blocking fork or adding explicit error handling to avoid blocking the rejection resolution.

**Suggested change:**
```suggestion
      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
      yield* Effect.fork(Effect.promise(() =>
        Bus.publish(BusRejected, { sessionID: existing.info.sessionID, requestID: existing.info.id }),
      ))
      // altimate_change end
```

#### 3. `packages/opencode/src/question/index.ts` (L110-L118)

**[🟠 MEDIUM]** The use of `any` type (`z.any()`) is strictly prohibited by our review checklist unless explicitly justified. Using `any` bypasses Zod's validation and type inference, potentially allowing unexpected data structures to propagate.

Please use `z.unknown()` or define a stricter Zod schema that matches the expected structure of `Question.Info` to maintain type safety.

**Suggested change:**
```suggestion
const BusAsked = BusEvent.define(
  "question.asked",
  z.object({
    id: QuestionID.zod,
    sessionID: SessionID.zod,
    questions: z.array(z.unknown()),
    tool: z.object({ messageID: MessageID.zod, callID: z.string() }).optional(),
  }),
)
```

#### 4. `packages/opencode/src/question/index.ts` (L204-L206)

**[🟠 MEDIUM]** This promise resolution contains an empty catch block: `.catch(() => {})`. This silently swallows any unexpected errors that might occur when `Effect.runPromise` executes, which violates error-handling best practices and can obscure underlying bugs during instance cleanup.

Consider logging the error explicitly instead of silently ignoring it, e.g., `.catch((e) => log.error(e))`.

**Suggested change:**
```suggestion
      for (const { deferred } of map.values()) {
        await Effect.runPromise(Deferred.fail(deferred, new RejectedError())).catch(console.error)
      }
```

</details>

packages/opencode/src/question/index.ts
```diff
@@ -213,11 +290,18 @@ export const layer = Layer.effect(
         sessionID: existing.info.sessionID,
         requestID: existing.info.id,
       })
+      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
+      yield* Effect.promise(() =>
+        Bus.publish(BusRejected, { sessionID: existing.info.sessionID, requestID: existing.info.id }),
+      )
+      // altimate_change end
```
dev-punia-altimate: **[🔴 HIGH]** Similarly, if `Bus.publish` throws or rejects here, the Effect will terminate early, preventing `Deferred.fail` from being executed and leaving the pending `ask` operation hanging indefinitely.

Consider placing this in a non-blocking fork or adding explicit error handling to avoid blocking the rejection resolution.

**Suggested change:**
```suggestion
      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
      yield* Effect.fork(Effect.promise(() =>
        Bus.publish(BusRejected, { sessionID: existing.info.sessionID, requestID: existing.info.id }),
      ))
      // altimate_change end
```

packages/opencode/src/question/index.ts
```diff
@@ -93,6 +97,39 @@ export const Event = {
...
+// they exist for the `type` string and typing.)
+const BusAsked = BusEvent.define(
+  "question.asked",
+  z.object({
+    id: QuestionID.zod,
+    sessionID: SessionID.zod,
+    questions: z.array(z.any()),
+    tool: z.object({ messageID: MessageID.zod, callID: z.string() }).optional(),
+  }),
+)
```
dev-punia-altimate: **[🟠 MEDIUM]** The use of `any` type (`z.any()`) is strictly prohibited by our review checklist unless explicitly justified. Using `any` bypasses Zod's validation and type inference, potentially allowing unexpected data structures to propagate.

Please use `z.unknown()` or define a stricter Zod schema that matches the expected structure of `Question.Info` to maintain type safety.

**Suggested change:**
```suggestion
const BusAsked = BusEvent.define(
  "question.asked",
  z.object({
    id: QuestionID.zod,
    sessionID: SessionID.zod,
    questions: z.array(z.unknown()),
    tool: z.object({ messageID: MessageID.zod, callID: z.string() }).optional(),
  }),
)
```

packages/opencode/src/question/index.ts
```diff
@@ -134,31 +193,30 @@ export const layer = Layer.effect(
...
+    // altimate_change start — clear a directory's pending questions when its
+    // instance is disposed/reloaded (mirrors the removed InstanceState finalizer)
+    // so entries in the process-global registry don't leak across instances.
+    const off = registerDisposer(async (directory) => {
+      const map = pendingByDir.get(directory)
+      if (!map) return
+      pendingByDir.delete(directory)
+      for (const { deferred } of map.values()) {
+        await Effect.runPromise(Deferred.fail(deferred, new RejectedError())).catch(() => {})
+      }
```
dev-punia-altimate, line 213: **[🟠 MEDIUM]** This promise resolution contains an empty catch block: `.catch(() => {})`. This silently swallows any unexpected errors that might occur when `Effect.runPromise` executes, which violates error-handling best practices and can obscure underlying bugs during instance cleanup.

Consider logging the error explicitly instead of silently ignoring it, e.g., `.catch((e) => log.error(e))`.

**Suggested change:**
```suggestion
      for (const { deferred } of map.values()) {
        await Effect.runPromise(Deferred.fail(deferred, new RejectedError())).catch(console.error)
      }
```

dev-punia-altimate:

No supported files changed.

dev-punia-altimate:

No supported files changed.

packages/opencode/src/question/index.ts
```diff
@@ -171,6 +229,12 @@ export const layer = Layer.effect(
       }
       pending.set(id, { info, deferred })
       yield* events.publish(Event.Asked, info)
+      // altimate_change start — also publish on the Bus wildcard so the IDE webview
+      // (subscribed to /event) receives question.asked and can answer the card.
+      yield* Effect.promise(() =>
+        Bus.publish(BusAsked, { id, sessionID: input.sessionID, questions: [...input.questions], tool: input.tool }),
+      )
```
sahrizvi: **[MAJOR]** Bus mirror on the Deferred critical path — can reintroduce the exact hang this PR fixes.

`Effect.promise` converts a promise **rejection into an unrecoverable defect** that aborts the fiber. This mirror sits **before** the `Effect.ensuring(Deferred.await, pending.delete)` block at lines 239-244 that registers cleanup, so if `Bus.publish` ever rejects:

- the awaiting `Question.ask()` fiber is aborted before it registers the cleanup finalizer;
- the pending entry stays in the global registry until instance disposal (**MINOR #5** — same root cause, folded here);
- and any dependent Deferred is never settled → tool hangs on "Thinking…" — the exact failure class this PR was written to fix.

A fire-and-forget notification must never be able to abort core question settlement.

**Required fix** — make the mirror best-effort (also apply at `reply()`/`reject()`):

```ts
const mirror = <D extends BusEvent.Definition>(def: D, props: z.output<D["properties"]>) =>
  Effect.promise(() => Bus.publish(def, props)).pipe(
    Effect.catchAllCause((cause) =>
      Effect.logWarning("question bus mirror failed", { type: def.type, cause }),
    ),
  )
// ...
yield* mirror(BusAsked, { ... })
```

`.pipe(Effect.ignore)` is the minimal form. Prefer settling the Deferred first, then mirroring.

*Flagged by all 7 reviewers. Probability of a `Bus.publish` rejection is low, but the impact is a permanent hang and the fix is trivial — every model rated this a blocker.*

packages/opencode/src/question/index.ts
```diff
@@ -197,11 +263,22 @@ export const layer = Layer.effect(
...
         answers: input.answers.map((a) => [...a]),
       })
+      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
+      yield* Effect.promise(() =>
+        Bus.publish(BusReplied, {
+          sessionID: existing.info.sessionID,
+          requestID: existing.info.id,
+          answers: input.answers.map((a) => [...a]),
+        }),
+      )
```
sahrizvi: **[MAJOR]** Same Bus-mirror-on-critical-path issue as `ask()` — this one is worse because the mirror runs **after** `pending.delete(input.requestID)` on line 259 but **before** `Deferred.succeed` on line 275.

If `Bus.publish` rejects:

- `Effect.promise` converts the rejection to an unrecoverable fiber defect;
- the Deferred is never succeeded → the awaiter of `Question.ask()` hangs forever;
- the pending entry is already deleted, so a retried reply returns `NotFoundError` — unrecoverable.

**Required fix** — settle the Deferred *first*, then fire the mirror best-effort:

```ts
yield* Deferred.succeed(existing.deferred, input.answers)
yield* mirror(BusReplied, { ... })  // Effect.promise + Effect.catchAllCause
```

See the MAJOR comment on `ask()` for the `mirror` helper.

packages/opencode/src/question/index.ts
```diff
@@ -213,11 +290,18 @@ export const layer = Layer.effect(
         sessionID: existing.info.sessionID,
         requestID: existing.info.id,
       })
+      // altimate_change start — mirror on the Bus wildcard for /event (webview) clients.
+      yield* Effect.promise(() =>
+        Bus.publish(BusRejected, { sessionID: existing.info.sessionID, requestID: existing.info.id }),
+      )
```
sahrizvi: **[MAJOR]** Same critical-path Bus mirror as `ask()`/`reply()`. Mirror runs after `pending.delete(requestID)` on line 287 but before `Deferred.fail` on line 298.

On `Bus.publish` rejection: fiber aborts before `Deferred.fail`, awaiter hangs, pending entry already gone → unrecoverable.

**Required fix** — settle first, then fire the mirror best-effort:

```ts
yield* Deferred.fail(existing.deferred, new RejectedError())
yield* mirror(BusRejected, { ... })
```

sahrizvi (commented):
## 7-model consensus review — verdict: REQUEST CHANGES

**Panel:** Claude + GPT Codex, Kimi K2.5, Qwen 3.6, MiniMax M2.7, GLM-5.1, MiMo V2.5 (Gemini excluded — auth failure). Quorum 6/6 met.
**Tests:** `bun test test/question/question.test.ts` → **14/14 pass** in the PR worktree.

Root-cause diagnosis and the two-part fix (globalThis pending registry + `Bus.publish` mirrors) are correct and unanimously praised. Architecture is sound.

One must-fix (**MAJOR #1**, unanimous across all 7 models) is inline. Other MINOR findings are inline where a diff line applies; the remainder are below.

---

### MINOR #4 — No regression test for the actual Bus-mirror fix

`waitForPending` in `test/question/question.test.ts` listens on the **EventV2** path, not Bus. Nothing asserts `Bus.publish`/`Bus.subscribeAll` delivery — deleting the three mirror lines would leave every test green. Add a test that subscribes via `Bus.subscribeAll` (or `Bus.subscribe(BusAsked)`) and asserts `question.asked/replied/rejected` arrive after `ask/reply/reject`. Ideally also: a test that `reply/reject` still settle the Deferred when `Bus.publish` fails (guards MAJOR #1).

### MINOR #6 — Shutdown/disposal semantics + globalThis anchor asymmetry

The old `InstanceState` finalizer failed all pending Deferreds when the layer scope closed; the new layer finalizer only calls `off()` (unregister) and relies on `registerDisposer` firing per-directory. Closing a `Question.layer` scope *without* going through `disposeInstance(directory)` now leaves pending fibers unresolved and entries in the global map. Negligible on full-process shutdown, but a behaviour delta.

Related: state is anchored on `globalThis` to survive module duplication, but the disposer registers into the **module-scoped** `disposers` Set in `instance-registry.ts` — if that module is ever duplicated the same way, cleanup silently no-ops. This is an inherited `InstanceState` assumption, but the PR's own thesis is that the assumption isn't safe. Consider a per-layer ownership token that drains only its own entries, and/or anchoring `disposers` on `globalThis` too.

### NITs

- `(globalThis as Record<string, unknown>)[...] as PendingByDir` — a non-Map value at that key would crash; a `declare global` augmentation or an `instanceof Map` guard is safer.
- `Effect.runPromise(Deferred.fail(...)).catch(() => {})` in the disposer spins a fresh runtime per deferred and swallows any future requirement; `Effect.runFork(...)` is lighter. `Deferred.fail` is currently `R = never`, so the `.catch` guards a non-event.

### Adjudicated / rejected claims (checked against source)

- **Kimi — "`Deferred.fail` throws if the Deferred is already completed":** false. In Effect, completing an already-settled Deferred returns `false`; it does not throw. The disposer loop is safe.
- **Qwen — "race condition in `pendingFor` on concurrent asks":** false. `pendingFor` is fully synchronous (no `yield`), so two fibers cannot interleave its get-or-create in single-threaded JS.
- **MiniMax — CRITICAL "`reply()`/`reject()` miss the directory lookup":** false and self-refuting — both call `pendingFor(yield* InstanceState.directory)`, identical to `ask()`. Not an issue.

- Root cause precisely diagnosed and documented inline; `altimate_change` markers balanced; change is minimal and confined to one file.
- The globalThis registry keyed by instance directory is the correct, robust fix for the split-module pending map and preserves per-instance `list()` semantics.
- Both delivery channels (EventV2 and Bus wildcard) are genuinely needed; the PR correctly identifies they are separate.
- Existing dispose/reload tests pass, validating the `registerDisposer` cleanup path.

---

*Convergence: all 7 reviewers independently surfaced finding #1 without prompting — no second round needed. Disagreements on severity (#2, #6) and three false-positive claims were adjudicated by reading the source, not by vote.*

packages/opencode/src/question/index.ts
```diff
@@ -93,6 +97,39 @@ export const Event = {
...
+// question card had no request id to reply with → "submit does nothing".
+// Publish these via `Bus.publish` too so they reach /event like every other
+// webview-visible event. (Schemas are loose — Bus.publish does not re-validate;
+// they exist for the `type` string and typing.)
+const BusAsked = BusEvent.define(
+  "question.asked",
+  z.object({
+    id: QuestionID.zod,
+    sessionID: SessionID.zod,
+    questions: z.array(z.any()),
```
sahrizvi: **[MINOR]** `z.any()` weakens the published contract.

`/event`'s OpenAPI is generated from `BusEvent.payloads()`, so the documented `question.asked` payload becomes `questions: any` — diverging from the SDK's typed `Array<QuestionInfo>`. Apply the same to `BusReplied.answers` (`z.array(z.array(z.string()))`) which is looser than `Array<Answer>`.

Reuse the structured schema instead. The `zod` adapter from `util/effect-zod` is already used by the question route:

```ts
const BusAsked = BusEvent.define("question.asked", zod(Request))
const BusReplied = BusEvent.define("question.replied", zod(Replied))
const BusRejected = BusEvent.define("question.rejected", zod(Rejected))
```

Flagged by Kimi, Qwen, GPT, MiniMax, GLM-5, MiMo.

packages/opencode/src/question/index.ts
```diff
@@ -93,6 +97,39 @@ export const Event = {
...
+// *wildcard PubSub* (`Bus.publish`), a different channel. So `question.asked`
+// never reached the webview → `pendingQuestions` stayed empty → the mcp-add
+// question card had no request id to reply with → "submit does nothing".
+// Publish these via `Bus.publish` too so they reach /event like every other
+// webview-visible event. (Schemas are loose — Bus.publish does not re-validate;
+// they exist for the `type` string and typing.)
+const BusAsked = BusEvent.define(
+  "question.asked",
+  z.object({
+    id: QuestionID.zod,
```
sahrizvi: **[MINOR]** Question events are now double-emitted on `GlobalBus`.

`EventV2Bridge.listen` already forwards every EventV2 event (incl. `question.asked/replied/rejected`) to `GlobalBus`, and `Bus.publish` here *also* emits to `GlobalBus`. So every `GlobalBus`/`/global/event` consumer now receives each question event twice (with different top-level ids).

**Verified impact — not the harm GLM-5 suspected:** the TUI trace consumer (`trace-consumer.ts` `handleEvent`) only handles `message.*`/`session.*` and ignores question events entirely, so there is no double trace recording. TUI `sync.tsx` reconciles by `request.id` and `notifications.ts` dedupes via a `Set<id>` — both idempotent. **No in-repo consumer breaks today.**

Residual risk is only for external `/global/event` subscribers / side-effecting plugins that don't dedupe. Making the mirror best-effort (MAJOR #1) plus a one-line comment acknowledging the double is sufficient here; a cleaner long-term option is a local-only Bus publish that doesn't re-emit to `GlobalBus`.

Flagged by Claude, GPT, MiniMax, GLM-5, MiMo.

packages/opencode/src/question/index.ts
saravmajestic: Fixed in `0b5ecd12f`. Added a best-effort `mirror()` helper — `Effect.promise(() => Bus.publish(...))` recovered with `Effect.catchCause` to a logged warning — so a publish failure can never abort the fiber. The Deferred is now settled **first**, then the mirror fires, so settlement can't be skipped. Applied to `ask()`, `reply()`, and `reject()`.

packages/opencode/src/question/index.ts
saravmajestic: Fixed in `0b5ecd12f`. Added a best-effort `mirror()` helper — `Effect.promise(() => Bus.publish(...))` recovered with `Effect.catchCause` to a logged warning — so a publish failure can never abort the fiber. The Deferred is now settled **first**, then the mirror fires, so settlement can't be skipped. Applied to `ask()`, `reply()`, and `reject()`.

packages/opencode/src/question/index.ts
saravmajestic: Fixed in `0b5ecd12f`. Added a best-effort `mirror()` helper — `Effect.promise(() => Bus.publish(...))` recovered with `Effect.catchCause` to a logged warning — so a publish failure can never abort the fiber. The Deferred is now settled **first**, then the mirror fires, so settlement can't be skipped. Applied to `ask()`, `reply()`, and `reject()`.

packages/opencode/src/question/index.ts
saravmajestic: Fixed in `0b5ecd12f`. Added a best-effort `mirror()` helper — `Effect.promise(() => Bus.publish(...))` recovered with `Effect.catchCause` to a logged warning — so a publish failure can never abort the fiber. The Deferred is now settled **first**, then the mirror fires, so settlement can't be skipped. Applied to `ask()`, `reply()`, and `reject()`.

packages/opencode/src/question/index.ts
saravmajestic: Fixed in `0b5ecd12f`. Added a best-effort `mirror()` helper — `Effect.promise(() => Bus.publish(...))` recovered with `Effect.catchCause` to a logged warning — so a publish failure can never abort the fiber. The Deferred is now settled **first**, then the mirror fires, so settlement can't be skipped. Applied to `ask()`, `reply()`, and `reject()`.

packages/opencode/src/question/index.ts
saravmajestic: Fixed in `0b5ecd12f`. Went further than `z.unknown()` per @sahrizvi's related note — the Bus schemas now reuse the structured Effect schemas via the `zod` adapter (`zod(Request)`), so the generated `/event` payload matches the SDK's typed shape.

packages/opencode/src/question/index.ts
saravmajestic: Fixed in `0b5ecd12f`. The dispose cleanup now recovers the cause with `Effect.catchCause` and logs `Effect.logWarning("question cleanup failed on dispose", { cause })` instead of swallowing it — kept in Effect-land rather than `console.error`.

packages/opencode/src/question/index.ts
saravmajestic: Fixed in `0b5ecd12f`. Replaced the hand-written loose schemas with `zod(Request)` / `zod(Replied)` / `zod(Rejected)` from `util/effect-zod`, so the `/event` OpenAPI payloads now match `Array<QuestionInfo>` / `Array<Answer>` instead of `any`.

packages/opencode/src/question/index.ts
saravmajestic: Acknowledged in `0b5ecd12f`. Kept the mirror (best-effort now) and added a comment documenting the intentional GlobalBus double-emit and your verification that no in-repo consumer is affected (TUI reconciles by request id, notifications dedupe by id, trace consumer ignores question events). Left the local-only-publish refactor as a future option.

sahrizvi (approved):
Approving — the question-hang fix is solid. The Bus mirror is now correctly kept off the `Deferred` settlement path (best-effort recovery via `catchCause`, and `reply()`/`reject()` settle before mirroring), the event schemas are structured via `zod(Request/Replied/Rejected)` instead of `z.any()`, and the duplicate emission on `GlobalBus` is documented and harmless for current consumers. Typecheck, the `question` suite, and the `/event` OpenAPI generation all pass.

Two non-blocking follow-ups for a separate PR:

**1. Add a regression test for the Bus mirror.** The fix's core invariant — a publish failure can't re-hang the tool — currently has no automated guard (deleting the three mirror lines leaves every test green). Suggested coverage:
- subscribe via `Bus.subscribeAll` and assert `question.asked` / `question.replied` / `question.rejected` are delivered after `ask` / `reply` / `reject`; and
- inject a `Bus.publish` failure and assert `reply()` / `reject()` still settle the `Deferred`.

**2. Same pattern exists (pre-existing) at `session/session.ts:924`.** `updatePartDelta` publishes via the identical unguarded call:
```ts
yield* Effect.promise(() => Bus.publish(MessageV2.Event.PartDelta, input))
```
`Effect.promise` turns a promise rejection into an unrecoverable fiber defect, so a `Bus.publish` rejection here would abort the streaming fiber rather than no-op — the same class this PR fixes for questions (lower impact: a streaming error rather than a stranded `Deferred`). Suggest routing it through the same best-effort recovery, ideally by extracting the `mirror` helper into a shared Bus utility so every `Effect.promise(() => Bus.publish(...))` site shares one hardened path.

Minor: `mirror` uses `Effect.catchCause`, which also catches interrupts. Consider `catchCauseIf(cause => !Cause.hasInterrupts(cause), …)` so an interrupted fiber isn't logged as "mirror failed" and swallowed — matches the existing `observe()` pattern in `event.ts`.

PR #1014, opened 2026-07-19
test: [de-fork S2] execution-route matrix + sentinel tests

anandgupta42:
PINEAPPLE

## Summary

Stage **S2** of the de-fork spike: the authoritative map of every tool-execution **dispatcher** (where a resolved tool's `.execute()` actually runs), so stage S3's HardPolicy knows every chokepoint it must cover. **Analysis + executable sentinel tests only — zero product-code change.**

Deliverables:
- `docs/internal/defork-route-matrix.md` — ingress surfaces vs execution dispatchers (D1–D8), each classified `active` / `latent` / `out-of-scope` with file:line evidence.
- `packages/opencode/test/altimate/defork/route-sentinels.test.ts` — an executable sentinel per active route, driving a harmless tool through the real dispatch code and asserting execution reaches the documented line.

### Key findings (these shape S3's HardPolicy)
- `SessionTools.resolve` (D3/D4) is **latent** — zero production callers at HEAD (a guard test flips if that changes), so it isn't a current bypass.
- The real bypass is **BatchTool inner dispatch** (`tool/batch.ts:104`) — it skips `Plugin.trigger` and source-stamping.
- **All active dispatchers live in 3 files** (`session/prompt.ts`, `tool/batch.ts`, `session/tools.ts`) → S3's ≤3-file audited-patch budget is feasible.
- Flags a pre-existing security gap (out of this spike's scope): the server-v2 `shell` HTTP route has only a session-existence check, no permission gate.

## Test Plan
- `bun test test/altimate/defork/route-sentinels.test.ts`: **9 pass / 0 fail** (verified in the primary dev worktree).

## Checklist
- [x] Tests added/updated
- [x] Documentation updated (route matrix)
- [ ] CHANGELOG updated — N/A (internal analysis + tests)

Closes #1011

---
## Summary by cubic
Maps all tool-execution dispatchers and adds sentinel tests to pin and exercise each route; commits the S2 spec and clarifies MCP checks and the `BatchTool` permission bypass. Adds an explicit coverage contract and expands latent-guard scanning; no product code changes. Closes #1011.

- **Highlights**
  - Docs: `docs/internal/defork-route-matrix.md` updated with a precise “Coverage contract” (D1/D5 have real-execution sentinels; D2/D6 use documented transitive/citation coverage), plus out-of-scope entries and the v1+v2 `shell` route gap (tracked in #1019); committed `docs/internal/2026-07-18-defork-spike-spec.md`.
[...]

PR #1018, opened 2026-07-19
feat: [de-fork S3] HardPolicy kill gate — non-bypassable DDL denies at every dispatcher

### LINKED ISSUE #1017: De-fork spike S3: HardPolicy kill gate (non-bypassable DDL denies at every dispatcher)
Stage S3 of the de-fork spike — the kill gate. Enforce non-bypassable hard denies (sql_execute DDL: DROP DATABASE/SCHEMA/TRUNCATE; bash DDL) at EVERY model-invoked tool-execution dispatcher, preserving today's behavior (these already block; S3 relocates them to one audited chokepoint).

- NEW \`packages/opencode/src/altimate/policy/hard-policy.ts\` — pure/synchronous/total \`HardPolicy.check()\`; malformed args fail closed (policy_internal_error), never throws/implicit-allows. Emits a structured audit record per call (the oracle — NOT the trace).
- HardPolicy.check inserted before tool.execute at all active + latent dispatchers (D1/D2/D6 in session/prompt.ts, D3/D4 in session/tools.ts, D5 in tool/batch.ts) + fail-closed init in effect/app-runtime.ts.
- Fixes a real bypass: the fork's resolvers discarded \`tool.execute.before\`'s return value, so a before-hook that mutated args (SELECT -> DROP DATABASE) executed the mutation unchecked. S3 captures the post-hook FINAL args and checks THOSE.
- 18 tests (bypass matrix): allow-all config still denies; near-miss controls run; deny is terminal (execute + tool.execute.after skipped); audit oracle asserted.

This is a security boundary + the spike's decision checkpoint. HELD for human go/no-go — do not auto-merge.

anandgupta42:
PINEAPPLE

## Summary

Stage **S3** of the de-fork spike — **the kill gate**. Enforces **non-bypassable hard denies** (`sql_execute` DDL: `DROP DATABASE`/`DROP SCHEMA`/`TRUNCATE`; bash DDL) at **every model-invoked tool-execution dispatcher**, relocating today's scattered per-tool safety checks into one audited chokepoint. **Behavior-preserving** — these operations already block; S3 consolidates and closes a bypass.

- **NEW `packages/opencode/src/altimate/policy/hard-policy.ts`** — pure, synchronous, **total** `HardPolicy.check()`: malformed/missing args return a deny (`policy_internal_error`), never throw, never implicit-allow. Emits a structured **audit record** on every call — this (not the trace) is the enforcement oracle the tests assert against.
- **`HardPolicy.check()` inserted before `tool.execute` at every dispatcher** from the S2 route matrix: D1 (registry loop) + D2 (MCP loop) + D6 (direct Task dispatch) in `session/prompt.ts`, D3/D4 in `session/tools.ts`, D5 (BatchTool inner dispatch) in `tool/batch.ts`. Fail-closed init via `assertInitialized()` at the app-runtime composition seam.
- **Closes a real bypass:** the fork's resolvers were *discarding* `tool.execute.before`'s return value, so a plugin hook that mutated args (e.g. rewrote a benign `SELECT` into `DROP DATABASE`) executed the mutation **unchecked**. Every site now captures the post-hook **final args** and checks *those*. The batch inner dispatch checks the **inner** tool ID + args (not `toolID="batch"`), closing the nested-dispatch gap.
- **On deny:** returns a `hard_policy_denied` tool-error result (or throws in batch) — `execute` and `tool.execute.after` are both skipped.

**Scope preserved:** bash DDL patterns are byte-identical to today's `agent.ts` safety table; the sql check reuses `sql-execute.ts`'s classifier + error string. `rm -rf` / `git push --force` / `git reset --hard` correctly stay **ask-tier** (not promoted to hard deny).

Also includes `route-sentinels.test.ts` made **line-number- and arg-agnostic** (matches `args` or S3's `finalArgs`) so it passes with or without S3 applied — shared identically with PR #1014, mergeable in any order.

### Review
Two independent adversarial reviews (a Claude reviewer + the orchestrator's own trace of every `.execute()` site) found **no bypass** via malformed args, mutated args, or allow-all permission config, and confirmed all 7 safety properties (coverage, final-args, total-function, deny-terminal, config-independence, fail-closed init, behavior-preservation).

## Test Plan
- `bun test test/altimate/defork/hardpolicy.test.ts`: **19 pass / 0 fail** — the bypass matrix across D1–D6: DDL denied + not executed at every route, near-miss controls allowed (proves rules aren't over-broad), allow-all config still denies, malformed args fail closed, deny is terminal, mutated-args checked on the post-hook value (audit `finalArgsDigest` asserted).
- Always-allow red-proof: stubbing `check()` to always-allow fails every dispatcher-integration test.
- `bunx tsc --noEmit -p packages/opencode`: clean for the changed files.

## Checklist
- [x] Tests added/updated
- [x] Documentation updated (spec §S3 + route matrix)
- [ ] CHANGELOG updated — N/A (internal safety guardrail)

Closes #1017

---
_This is a security boundary and the spike's decision checkpoint — **held for human go/no-go; do not auto-merge.**_

---
## Summary by cubic
Adds a centralized kill gate (`HardPolicy`) that enforces non-bypassable denies for destructive SQL/Bash DDL across all tool-execution dispatchers, with audited decisions. Follow-up fixes close whitespace and MCP-prefix bypasses, harden audit hashing, and make tool-call metadata reflect post-hook args. Closes #1017.

- New Features
  - New `packages/opencode/src/altimate/policy/hard-policy.ts` with pure, sync, total `HardPolicy.check()` that emits an audit record on every call.
  - Fail-closed init via `HardPolicy.assertInitialized()` in `packages/opencode/src/effect/app-runtime.ts`.
  - Enforcement added before `execute` in all routes; on deny, return `hard_policy_denied` (or throw in batch) and skip `execute`/`tool.execute.after`.

- Bug Fixes
  - Closed bypasses: MCP-prefixed governed keys now resolve via `_`-suffix matching; bash DDL matcher trims leading/trailing whitespace before checks.
  - Audit hardened: `finalArgsDigest` is a SHA-256 of stable-stringified args; stable stringify now encodes `Map`/`Set`; audit sink wrapped in try/catch and log trimming is amortized O(1).
  - Final-args fidelity: post-hook args are used for `HardPolicy.check()`, `execute`, audit, and `ctx.metadata`; batch checks the inner tool ID + args.

<sup>Written for commit 348e930cf9e5ab5a7ff2dd54bcbb4c5d6b2c0fa8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added centralized, fail-closed hard policy enforcement blocking dangerous SQL and Bash actions (DROP DATABASE/SCHEMA, TRUNCATE) across native, MCP, batch, and task execution.
  * Policy checks now use post-hook arguments, with consistent “Blocked by policy” results when denied.
  * Introduced deterministic audit logging for every decision and an optional audit sink; enforcement initialization is validated at module startup.

* **Tests**
  * Added comprehensive coverage for allow/deny behavior, audit/oracle expectations, post-hook argument handling, MCP/batch/task enforcement, and execution-route sentinels.

### DIFF
--- a/packages/opencode/src/altimate/policy/hard-policy.ts
+++ b/packages/opencode/src/altimate/policy/hard-policy.ts
@@ -21,6 +21,7 @@
 //   absent execute span does NOT prove enforcement. Tests must assert against the audit
 //   log and execute-not-called counters.
 
+import { createHash } from "node:crypto"
 import { Wildcard } from "@/util/wildcard"
 import { classifyAndCheck } from "@/altimate/tools/sql-classify"
 
@@ -191,6 +192,23 @@ export namespace HardPolicy {
   const rulesByTool = buildRulesByTool(RULES)
   const initError = validateRuleTable(RULES)
 
+  // MCP tools are registered with a flattened id `<sanitized-client>_<sanitized-tool>`
+  // (see src/mcp/index.ts's MCP.tools()), but RULES are keyed by the bare governed tool id
+  // (`sql_execute`, `bash`). Without this resolution an MCP server exposing its own
+  // `sql_execute` arrives as e.g. `warehouse_sql_execute`, matches no rule, and is allowed
+  // to run DDL — a silent bypass. For MCP-sourced calls, also resolve a governed key that is
+  // a full `_`-delimited suffix of the flattened id. Native/plugin/batch/task tools keep
+  // strict exact-match (their ids are not client-prefixed), so this only ever ADDS coverage
+  // for the mcp source; it never un-governs a tool that exact-match already caught.
+  function resolveGovernedKey(toolID: string, source: Source): string | undefined {
+    if (rulesByTool.has(toolID)) return toolID
+    if (source !== "mcp") return undefined
+    for (const governed of rulesByTool.keys()) {
+      if (toolID.endsWith(`_${governed}`)) return governed
+    }
+    return undefined
+  }
+
   /**
    * Must be called at the app-runtime composition seam so a broken rule table fails app
    * startup instead of silently allowing everything. Throws (does not process.exit —
@@ -247,12 +265,16 @@ export namespace HardPolicy {
   }
 
   // ---------------------------------------------------------------------------------
-  // Deterministic, bounded digest of the final args snapshot — for audit correlation in
-  // tests (e.g. "was the deny keyed off the post-hook-mutated args?"), not a
-  // cryptographic hash and not for security purposes on its own.
+  // Deterministic, non-reversible digest of the final args snapshot — for audit
+  // correlation in tests (e.g. "was the deny keyed off the post-hook-mutated args?").
+  // It is a SHA-256 over the stable-stringified args, NOT the raw args: tool arguments
+  // routinely carry secrets (a `write` of `.env` contents, a `bash` command with an auth
+  // token). The audit log is retained (up to MAX_AUDIT_LOG records, readable via
+  // getAuditLog() or an installed sink), so storing raw or truncated args would retain
+  // those secrets near the start of the payload. The hash preserves the correlation
+  // property (same args -> same digest, different args -> different digest) without
+  // retaining the plaintext. Tests recompute the expected digest via exported digestArgs().
   // ---------------------------------------------------------------------------------
-
-  const MAX_DIGEST_LENGTH = 2000
 
   function stableStringify(value: unknown): string {
     const seen = new WeakSet<object>()
@@ -277,8 +299,16 @@ export namespace HardPolicy {
   }
 
   function finalArgsDigest(args: unknown): string {
-    const str = stableStringify(args)
-    return str.length > MAX_DIGEST_LENGTH ? str.slice(0, MAX_DIGEST_LENGTH) : str
+    return createHash("sha256").update(stableStringify(args)).digest("hex")
+  }
+
+  /**
+   * Test/forensic support: the exact non-reversible digest check() records for a given
+   * args snapshot. Lets tests assert the audit digest reflects the post-hook (final) args
+   * HardPolicy actually evaluated, without exposing or reconstructing the plaintext.
+   */
+  export function digestArgs(args: unknown): string {
+    return finalArgsDigest(args)
   }
 
   // ---------------------------------------------------------------------------------
@@ -308,7 +338,7 @@ export namespace HardPolicy {
 
       assertInitialized()
 
-      const rules = rulesByTool.get(toolID)
+      const rules = rulesByTool.get(resolveGovernedKey(toolID, source) ?? "")
       if (!rules || rules.length === 0) {
         // Ungoverned toolID — HardPolicy has no rule for it, so it always allows,
         // regardless of args shape. Denying here would be a NEW block, not a relocation
--- a/packages/opencode/src/altimate/policy/hard-policy.ts
+++ b/packages/opencode/src/altimate/policy/hard-policy.ts
@@ -107,7 +107,11 @@ export namespace HardPolicy {
     if (!isRecord(args) || typeof args.command !== "string") {
       throw new Error("hard-policy: bash args missing string 'command' field")
     }
-    const command = args.command
+    // Trim first: Wildcard.match compiles each pattern with a leading `^` anchor, so a raw
+    // `"  DROP DATABASE prod"` (or a leading tab/newline) would slip past `"DROP DATABASE *"`
+    // and bypass the hard deny. Leading/trailing whitespace is never semantically meaningful
+    // to the shell here, so trimming closes that bypass without altering any real command.
+    const command = args.command.trim()
     return BASH_DDL_PATTERNS.some((pattern) => Wildcard.match(command, pattern))
   }
 
@@ -285,6 +289,15 @@ export namespace HardPolicy {
       if (input === null || typeof input !== "object") return input
       if (seen.has(input)) return "[circular]"
       seen.add(input)
+      // Map/Set stringify to `{}` under Object.keys(), so distinct ones would collide to the
+      // same digest (a `tool.execute.before` hook could inject one). Serialize their contents
+      // under a distinguishing tag so different collections produce different digests.
+      if (input instanceof Map) {
+        return { __map__: [...input.entries()].map(([k, v]) => [normalize(k), normalize(v)]) }
+      }
+      if (input instanceof Set) {
+        return { __set__: [...input.values()].map(normalize) }
+      }
       if (Array.isArray(input)) return input.map(normalize)
       const record = input as Record<string, unknown>
       const out: Record<string, unknown> = {}
--- a/packages/opencode/src/altimate/policy/hard-policy.ts
+++ b/packages/opencode/src/altimate/policy/hard-policy.ts
@@ -215,8 +215,20 @@ export namespace HardPolicy {
 
   function emitAudit(record: AuditRecord): void {
     auditLog.push(record)
-    if (auditLog.length > MAX_AUDIT_LOG) auditLog.shift()
-    if (auditSink) auditSink(record)
+    // Batched trim (O(1) amortized) instead of an O(n) shift() on every call:
+    // let the log overshoot, then splice the oldest entries once it grows past
+    // twice the cap. check() runs on the tool-execution hot path.
+    if (auditLog.length > MAX_AUDIT_LOG * 2) auditLog.splice(0, auditLog.length - MAX_AUDIT_LOG)
+    // A faulty externally-installed sink must NEVER propagate out of the
+    // enforcement chokepoint — check() is contractually total (never throws).
+    // The decision is already computed and returned regardless of audit outcome.
+    if (auditSink) {
+      try {
+        auditSink(record)
+      } catch {
+        // Intentionally swallowed: audit-side failure cannot affect enforcement.
+      }
+    }
   }
 
   /** Test support: read the in-memory audit log (oldest first). */
--- a/packages/opencode/src/altimate/policy/hard-policy.ts
+++ b/packages/opencode/src/altimate/policy/hard-policy.ts
@@ -1,0 +1,337 @@
+// altimate_change - HardPolicy: single fork-owned chokepoint for non-bypassable hard denies (S3, de-fork spike)
+//
+// This consolidates two hard blocks that already existed in scattered form:
+//   1. sql_execute DDL block (previously inline in altimate/tools/sql-execute.ts)
+//   2. bash DDL block (previously inline in agent.ts's safetyDenials permission table)
+// into ONE pure, synchronous, TOTAL function so every tool-execution dispatcher can call
+// the same check instead of re-implementing (or forgetting) it.
+//
+// Design contract (see docs/internal/2026-07-18-defork-spike-spec.md §S3):
+// - TOTAL: never throws, never implicitly allows on malformed input. Any internal failure
+//   (bad args shape, broken rule table, classifier exception) resolves to a deny with
+//   ruleID "policy_internal_error" — fail closed, not fail open.
+// - PURE + SYNCHRONOUS: no I/O, no async, safe to call from both Promise-based and
+//   Effect-based dispatchers without adapters.
+// - v1 rules are BEHAVIOR-PRESERVING relocations of existing blocks, not new blocks. Do
+//   NOT add new hard denies here (e.g. rm -rf, git push --force) without a product
+//   decision — those stay ask-tier in the Ruleset, user-overridable.
+// - Every check() call emits a structured audit record. The audit log — not trace
+//   evidence — is the security oracle: TraceSpan.status is only ok|error (no "denied"
+//   state), and denials throw DeniedError without publishing permission events, so an
+//   absent execute span does NOT prove enforcement. Tests must assert against the audit
+//   log and execute-not-called counters.
+
+import { Wildcard } from "@/util/wildcard"
+import { classifyAndCheck } from "@/altimate/tools/sql-classify"
+
+export namespace HardPolicy {
+  // ---------------------------------------------------------------------------------
+  // Public types
+  // ---------------------------------------------------------------------------------
+
+  export type Source = "native" | "plugin" | "mcp" | "batch" | "task"
+
+  export interface Input {
+    toolID: string
+    source: Source
+    args: unknown
+    sessionID: string
+    agentID?: string
+    callID?: string
+  }
+
+  export type Decision = { allow: true } | { allow: false; ruleID: string; safeReason: string }
+
+  export interface AuditRecord {
+    ruleID: string
+    dispatcher: Source
+    toolID: string
+    source: Source
+    finalArgsDigest: string
+    sessionID: string
+    callID?: string
+    decision: Decision
+    timestamp: number
+  }
+
+  // ---------------------------------------------------------------------------------
+  // Rule table (v1) — versioned, each rule is a relocation of a pre-existing hard block.
+  // ---------------------------------------------------------------------------------
+
+  interface Rule {
+    ruleID: string
+    toolID: string
+    // Returns true if the args match this rule's deny condition. MUST throw (not return
+    // false) if `args` is not shaped as this rule expects — that throw is caught by
+    // check() and converted into a policy_internal_error deny. This keeps malformed-args
+    // handling fail-closed for GOVERNED tool IDs (those with an applicable rule) without
+    // extending denial behavior to tool IDs HardPolicy was never meant to touch.
+    match: (args: unknown) => boolean
+    safeReason: string
+    positiveExamples: unknown[]
+    negativeExamples: unknown[]
+  }
+
+  function isRecord(value: unknown): value is Record<string, unknown> {
+    return typeof value === "object" && value !== null
+  }
+
+  // Mirrors altimate/tools/sql-execute.ts:24-28 — same classifier, same denied types
+  // (DROP DATABASE, DROP SCHEMA, TRUNCATE), same "cannot be overridden" behavior.
+  function matchSqlDdl(args: unknown): boolean {
+    if (!isRecord(args) || typeof args.query !== "string") {
+      throw new Error("hard-policy: sql_execute args missing string 'query' field")
+    }
+    return classifyAndCheck(args.query).blocked
+  }
+
+  // Mirrors agent.ts's safetyDenials bash table (9 patterns; UPPER/lowercase/Title-Case
+  // only — NOT fully case-insensitive, because Wildcard.match is case-sensitive on
+  // non-Windows). Whole-command-string match, not per-AST-subcommand — a documented v1
+  // simplification vs. bash.ts's tree-sitter subcommand splitting used for the ask-tier
+  // permission flow.
+  const BASH_DDL_PATTERNS = [
+    "DROP DATABASE *",
+    "DROP SCHEMA *",
+    "TRUNCATE *",
+    "drop database *",
+    "drop schema *",
+    "truncate *",
+    "Drop Database *",
+    "Drop Schema *",
+    "Truncate *",
+  ]
+
+  function matchBashDdl(args: unknown): boolean {
+    if (!isRecord(args) || typeof args.command !== "string") {
+      throw new Error("hard-policy: bash args missing string 'command' field")
+    }
+    const command = args.command
+    return BASH_DDL_PATTERNS.some((pattern) => Wildcard.match(command, pattern))
+  }
+
+  const SQL_DDL_SAFE_REASON =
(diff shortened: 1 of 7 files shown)


packages/opencode/src/session/tools.ts
```diff
@@ -89,12 +92,39 @@ export const resolve = Effect.fn("SessionTools.resolve")(function* (input: {
...
+                output: policyDecision.safeReason,
+                metadata: {
+                  error: policyDecision.safeReason,
+                  hard_policy_denied: true,
+                  code: "hard_policy_denied",
+                  ruleID: policyDecision.ruleID,
+                  success: false,
+                },
+              }
+            }
```
dev-punia-altimate, line 131: **[🟠 MEDIUM]** When `HardPolicy.check` blocks the tool execution and returns early, the `tool.execute.after` plugin hook is completely skipped. 

While it's true the tool itself wasn't executed, any plugins that hooked into `tool.execute.before` (and perhaps started a span, locked a resource, or initialized state) will never receive the corresponding `tool.execute.after` event to clean up or log the blocked outcome. If the intention is to completely short-circuit, this might be fine, but typically a paired `before`/`after` hook design expects `after` to be called with the error or blocked result. Please verify if plugins need to be notified of the hard policy denial.

packages/opencode/src/session/tools.ts
```diff
@@ -89,12 +92,39 @@ export const resolve = Effect.fn("SessionTools.resolve")(function* (input: {
...
-            yield* plugin.trigger(
+            // altimate_change start — HardPolicy enforcement (S3)
+            // upstream_fix: plugin.trigger's returned output was previously discarded, so a
+            // tool.execute.before hook that mutated `args` had no effect on what actually
+            // ran — HardPolicy and execute must see the SAME final, post-hook args.
+            const hookOutput = yield* plugin.trigger(
               "tool.execute.before",
               { tool: item.id, sessionID: ctx.sessionID, callID: ctx.callID },
               { args },
             )
```
dev-punia-altimate: **[🟠 MEDIUM]** You correctly ensured that `HardPolicy.check` and the tool's `execute` function use `finalArgs`. However, `ctx` is initialized *before* the plugin hook using the original `args`. 

```typescript
const ctx = context(args, options)
```
Inside the `context` factory, `args` is captured in a closure and used when the tool updates its metadata (e.g., `input: args` in `updateToolCall`). This means the system state/UI will record the *original* unmutated `args` as the tool's input, even though it actually executed with `finalArgs`.

If the system state should reflect the mutated arguments, you should create `ctx` *after* the hook (using `sessionID: input.session.id` and `callID: options.toolCallId` directly for the hook input), or update the `context` factory to use the latest args.

packages/opencode/src/session/tools.ts
```diff
@@ -135,14 +167,44 @@ export const resolve = Effect.fn("SessionTools.resolve")(function* (input: {
...
-          yield* plugin.trigger(
+          // altimate_change start — HardPolicy enforcement (S3)
+          // upstream_fix: plugin.trigger's returned output was previously discarded, so a
+          // tool.execute.before hook that mutated `args` had no effect on what actually
+          // ran — HardPolicy and execute must see the SAME final, post-hook args.
+          const hookOutput = yield* plugin.trigger(
             "tool.execute.before",
             { tool: key, sessionID: ctx.sessionID, callID: opts.toolCallId },
             { args },
           )
```
dev-punia-altimate: **[🟠 MEDIUM]** Similarly for MCP tools, `ctx` captures the original `args` before the `tool.execute.before` hook can mutate them. Any calls to `ctx.metadata(...)` during execution will save the original `args` to the tool call state rather than `finalArgs`.

Consider moving `const ctx = context(finalArgs, opts)` to after the hook, using `input.session.id` for the hook's `sessionID` parameter.

packages/opencode/src/altimate/policy/hard-policy.ts
```diff
@@ -0,0 +1,337 @@
...
+    "Drop Schema *",
+    "Truncate *",
+  ]
+
+  function matchBashDdl(args: unknown): boolean {
+    if (!isRecord(args) || typeof args.command !== "string") {
+      throw new Error("hard-policy: bash args missing string 'command' field")
+    }
+    const command = args.command
+    return BASH_DDL_PATTERNS.some((pattern) => Wildcard.match(command, pattern))
```
dev-punia-altimate: **[🔴 HIGH]** The `Wildcard.match` function typically expects a strict prefix match since it compiles patterns with a `^` anchor. Because the `BASH_DDL_PATTERNS` start directly with the command (e.g., `"DROP DATABASE *"`), a user could bypass this hard policy by simply prepending whitespace (e.g., `"  DROP DATABASE prod"`).

To prevent this easy bypass, consider trimming the command string before matching.

**Suggested change:**
```suggestion
    const command = args.command.trim()
    return BASH_DDL_PATTERNS.some((pattern) => Wildcard.match(command, pattern))
```

packages/opencode/src/altimate/policy/hard-policy.ts
```diff
@@ -0,0 +1,337 @@
...
+      if (typeof input === "function" || typeof input === "symbol") return String(input)
+      if (typeof input === "bigint") return input.toString()
+      if (input === null || typeof input !== "object") return input
+      if (seen.has(input)) return "[circular]"
+      seen.add(input)
+      if (Array.isArray(input)) return input.map(normalize)
+      const record = input as Record<string, unknown>
+      const out: Record<string, unknown> = {}
+      for (const key of Object.keys(record).sort()) out[key] = normalize(record[key])
+      return out
```
dev-punia-altimate, line 305: **[🟠 MEDIUM]** If `input.args` happens to contain non-plain objects like `Map` or `Set` (e.g., mutated by a `tool.execute.before` hook), `Object.keys()` will return an empty array, and they will be stringified as `{}`. This could lead to identical audit digests for different inputs. 

Although tool inputs are typically plain JSON objects, consider handling `Map` and `Set` or falling back to a more robust serialization if they are expected.

**Suggested change:**
```suggestion
      if (Array.isArray(input)) return input.map(normalize)
      if (input instanceof Map) return normalize(Object.fromEntries(input))
      if (input instanceof Set) return normalize(Array.from(input))
      const record = input as Record<string, unknown>
      const out: Record<string, unknown> = {}
      for (const key of Object.keys(record).sort()) out[key] = normalize(record[key])
      return out
```

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 6 finding(s)

- 6 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `packages/opencode/src/session/tools.ts` (L112-L124)

**[🟠 MEDIUM]** When `HardPolicy.check` blocks the tool execution and returns early, the `tool.execute.after` plugin hook is completely skipped. 

While it's true the tool itself wasn't executed, any plugins that hooked into `tool.execute.before` (and perhaps started a span, locked a resource, or initialized state) will never receive the corresponding `tool.execute.after` event to clean up or log the blocked outcome. If the intention is to completely short-circuit, this might be fine, but typically a paired `before`/`after` hook design expects `after` to be called with the error or blocked result. Please verify if plugins need to be notified of the hard policy denial.

#### 2. `packages/opencode/src/session/tools.ts` (L94-L103)

**[🟠 MEDIUM]** You correctly ensured that `HardPolicy.check` and the tool's `execute` function use `finalArgs`. However, `ctx` is initialized *before* the plugin hook using the original `args`. 

```typescript
const ctx = context(args, options)
```
Inside the `context` factory, `args` is captured in a closure and used when the tool updates its metadata (e.g., `input: args` in `updateToolCall`). This means the system state/UI will record the *original* unmutated `args` as the tool's input, even though it actually executed with `finalArgs`.

If the system state should reflect the mutated arguments, you should create `ctx` *after* the hook (using `sessionID: input.session.id` and `callID: options.toolCallId` directly for the hook input), or update the `context` factory to use the latest args.

#### 3. `packages/opencode/src/session/tools.ts` (L169-L178)

**[🟠 MEDIUM]** Similarly for MCP tools, `ctx` captures the original `args` before the `tool.execute.before` hook can mutate them. Any calls to `ctx.metadata(...)` during execution will save the original `args` to the tool call state rather than `finalArgs`.

Consider moving `const ctx = context(finalArgs, opts)` to after the hook, using `input.session.id` for the hook's `sessionID` parameter.

#### 4. `packages/opencode/src/altimate/policy/hard-policy.ts` (L109-L110)

**[🔴 HIGH]** The `Wildcard.match` function typically expects a strict prefix match since it compiles patterns with a `^` anchor. Because the `BASH_DDL_PATTERNS` start directly with the command (e.g., `"DROP DATABASE *"`), a user could bypass this hard policy by simply prepending whitespace (e.g., `"  DROP DATABASE prod"`).

To prevent this easy bypass, consider trimming the command string before matching.

**Suggested change:**
```suggestion
    const command = args.command.trim()
    return BASH_DDL_PATTERNS.some((pattern) => Wildcard.match(command, pattern))
```

#### 5. `packages/opencode/src/altimate/policy/hard-policy.ts` (L254-L258)

**[🟠 MEDIUM]** If `input.args` happens to contain non-plain objects like `Map` or `Set` (e.g., mutated by a `tool.execute.before` hook), `Object.keys()` will return an empty array, and they will be stringified as `{}`. This could lead to identical audit digests for different inputs. 

Although tool inputs are typically plain JSON objects, consider handling `Map` and `Set` or falling back to a more robust serialization if they are expected.

**Suggested change:**
```suggestion
      if (Array.isArray(input)) return input.map(normalize)
      if (input instanceof Map) return normalize(Object.fromEntries(input))
      if (input instanceof Set) return normalize(Array.from(input))
      const record = input as Record<string, unknown>
      const out: Record<string, unknown> = {}
      for (const key of Object.keys(record).sort()) out[key] = normalize(record[key])
      return out
```

#### 6. `packages/opencode/src/altimate/policy/hard-policy.ts` (L288-L292)

**[🔴 HIGH]** If `input.toolID` is missing or not a string (e.g., due to a caller bug), it defaults to `"unknown"`. Since `"unknown"` is an ungoverned tool ID, the policy will unconditionally allow the execution. This represents a fail-open behavior on malformed top-level input, which contradicts the fail-closed design contract stated in the comments. 

Consider throwing an error here instead of defaulting to `"unknown"` to ensure any such caller bugs result in a secure deny.

**Suggested change:**
```suggestion
      if (!isRecord(input as unknown)) {
        throw new Error("hard-policy: input is not an object")
      }
      if (typeof input.toolID !== "string") {
        throw new Error("hard-policy: input missing string 'toolID'")
      }
      toolID = input.toolID
      source = input.source
```

</details>

packages/opencode/src/altimate/policy/hard-policy.ts
```diff
@@ -0,0 +1,337 @@
...
+    let sessionID = ""
+    let callID: string | undefined
+    let digest = ""
+
+    try {
+      if (!isRecord(input as unknown)) {
+        throw new Error("hard-policy: input is not an object")
+      }
+      toolID = typeof input.toolID === "string" ? input.toolID : "unknown"
+      source = input.source
```
dev-punia-altimate, line 347: **[🔴 HIGH]** If `input.toolID` is missing or not a string (e.g., due to a caller bug), it defaults to `"unknown"`. Since `"unknown"` is an ungoverned tool ID, the policy will unconditionally allow the execution. This represents a fail-open behavior on malformed top-level input, which contradicts the fail-closed design contract stated in the comments. 

Consider throwing an error here instead of defaulting to `"unknown"` to ensure any such caller bugs result in a secure deny.

**Suggested change:**
```suggestion
      if (!isRecord(input as unknown)) {
        throw new Error("hard-policy: input is not an object")
      }
      if (typeof input.toolID !== "string") {
        throw new Error("hard-policy: input missing string 'toolID'")
      }
      toolID = input.toolID
      source = input.source
```

dev-punia-altimate:

No supported files changed.

dev-punia-altimate:

No supported files changed.

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 3 finding(s)

- 3 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `packages/opencode/src/altimate/policy/hard-policy.ts` (L24)

**[🟠 MEDIUM]** This `import` introduces a strict dependency on Node.js/Bun. If `HardPolicy` (or any module importing it) is executed in a non-Node environment like a browser, Edge runtime, or a VS Code web extension, this will cause a runtime crash.

Consider using a cross-platform alternative, such as a pure JS implementation of SHA-256 (e.g., `crypto-js`), using the Web Crypto API (if asynchronous execution is permissible), or injecting the hashing function dynamically based on the environment.

#### 2. `packages/opencode/src/altimate/policy/hard-policy.ts` (L203-L210)

**[🔴 HIGH]** The suffix matching loop here depends on the `Map` insertion order. If there are overlapping suffixes (for instance, rules for both `execute` and `sql_execute`), and `execute` was inserted into `RULES` first, an MCP tool ID like `warehouse_sql_execute` would incorrectly match `execute` instead of `sql_execute`.

To avoid this vulnerability, sort the governed keys by length in descending order, ensuring the most specific suffix matches first.

**Suggested change:**
```suggestion
  function resolveGovernedKey(toolID: string, source: Source): string | undefined {
    if (rulesByTool.has(toolID)) return toolID
    if (source !== "mcp") return undefined
    
    // Sort keys by length descending to ensure the most specific suffix matches first
    const governedKeys = Array.from(rulesByTool.keys()).sort((a, b) => b.length - a.length)
    for (const governed of governedKeys) {
      if (toolID.endsWith(`_${governed}`)) return governed
    }
    return undefined
  }
```

#### 3. `packages/opencode/src/altimate/policy/hard-policy.ts` (L243-L249)

**[🔴 HIGH]** The `try/catch` block only handles synchronous errors. If an externally provided `auditSink` returns a rejected Promise (e.g., if it's an `async` function), the rejection will silently bypass the `catch` block. This leads to an unhandled promise rejection, which can crash the Node.js process.

Safely catch potential Promise rejections to ensure failures on the audit side cannot affect enforcement.

**Suggested change:**
```suggestion
    if (auditSink) {
      try {
        const result = auditSink(record)
        if (result && typeof (result as any).catch === "function") {
          (result as any).catch(() => {})
        }
      } catch {
        // Intentionally swallowed: audit-side failure cannot affect enforcement.
      }
    }
```

</details>

packages/opencode/src/altimate/policy/hard-policy.ts
```diff
@@ -0,0 +1,379 @@
...
+// - v1 rules are BEHAVIOR-PRESERVING relocations of existing blocks, not new blocks. Do
+//   NOT add new hard denies here (e.g. rm -rf, git push --force) without a product
+//   decision — those stay ask-tier in the Ruleset, user-overridable.
+// - Every check() call emits a structured audit record. The audit log — not trace
+//   evidence — is the security oracle: TraceSpan.status is only ok|error (no "denied"
+//   state), and denials throw DeniedError without publishing permission events, so an
+//   absent execute span does NOT prove enforcement. Tests must assert against the audit
+//   log and execute-not-called counters.
+
+import { createHash } from "node:crypto"
```
dev-punia-altimate, line 24: **[🟠 MEDIUM]** This `import` introduces a strict dependency on Node.js/Bun. If `HardPolicy` (or any module importing it) is executed in a non-Node environment like a browser, Edge runtime, or a VS Code web extension, this will cause a runtime crash.

Consider using a cross-platform alternative, such as a pure JS implementation of SHA-256 (e.g., `crypto-js`), using the Web Crypto API (if asynchronous execution is permissible), or injecting the hashing function dynamically based on the environment.

packages/opencode/src/altimate/policy/hard-policy.ts
```diff
@@ -0,0 +1,379 @@
...
+  // strict exact-match (their ids are not client-prefixed), so this only ever ADDS coverage
+  // for the mcp source; it never un-governs a tool that exact-match already caught.
+  function resolveGovernedKey(toolID: string, source: Source): string | undefined {
+    if (rulesByTool.has(toolID)) return toolID
+    if (source !== "mcp") return undefined
+    for (const governed of rulesByTool.keys()) {
+      if (toolID.endsWith(`_${governed}`)) return governed
+    }
+    return undefined
+  }
```
dev-punia-altimate, line 214: **[🔴 HIGH]** The suffix matching loop here depends on the `Map` insertion order. If there are overlapping suffixes (for instance, rules for both `execute` and `sql_execute`), and `execute` was inserted into `RULES` first, an MCP tool ID like `warehouse_sql_execute` would incorrectly match `execute` instead of `sql_execute`.

To avoid this vulnerability, sort the governed keys by length in descending order, ensuring the most specific suffix matches first.

**Suggested change:**
```suggestion
  function resolveGovernedKey(toolID: string, source: Source): string | undefined {
    if (rulesByTool.has(toolID)) return toolID
    if (source !== "mcp") return undefined
    
    // Sort keys by length descending to ensure the most specific suffix matches first
    const governedKeys = Array.from(rulesByTool.keys()).sort((a, b) => b.length - a.length)
    for (const governed of governedKeys) {
      if (toolID.endsWith(`_${governed}`)) return governed
    }
    return undefined
  }
```

packages/opencode/src/altimate/policy/hard-policy.ts
```diff
@@ -0,0 +1,379 @@
...
+    // A faulty externally-installed sink must NEVER propagate out of the
+    // enforcement chokepoint — check() is contractually total (never throws).
+    // The decision is already computed and returned regardless of audit outcome.
+    if (auditSink) {
+      try {
+        auditSink(record)
+      } catch {
+        // Intentionally swallowed: audit-side failure cannot affect enforcement.
+      }
+    }
```
dev-punia-altimate, line 253: **[🔴 HIGH]** The `try/catch` block only handles synchronous errors. If an externally provided `auditSink` returns a rejected Promise (e.g., if it's an `async` function), the rejection will silently bypass the `catch` block. This leads to an unhandled promise rejection, which can crash the Node.js process.

Safely catch potential Promise rejections to ensure failures on the audit side cannot affect enforcement.

**Suggested change:**
```suggestion
    if (auditSink) {
      try {
        const result = auditSink(record)
        if (result && typeof (result as any).catch === "function") {
          (result as any).catch(() => {})
        }
      } catch {
        // Intentionally swallowed: audit-side failure cannot affect enforcement.
      }
    }
```

anandgupta42:
## Review comments addressed (2 rounds)

Thanks to @dev-punia-altimate, @chatgpt-codex-connector, and @cubic-dev-ai.

**Round 2 — `3ac7b68324`**
- **P1 — MCP-prefix DDL bypass**: rules keyed `sql_execute`/`bash`, but MCP tools arrive flattened as `<client>_<tool>` (e.g. `warehouse_sql_execute`) → no rule matched → DDL ran. Added `resolveGovernedKey()` (mcp-source only, full `_`-delimited suffix; native keeps exact-match).
- **P1 — audit arg retention**: `finalArgsDigest` stored raw truncated args (could hold secrets); now a SHA-256 hash, correlation preserved.

**Round 3 — `348e930cf9`**
- **HIGH** (@dev-punia-altimate — bash whitespace bypass): `Wildcard.match` anchors at `^`, so `"  DROP DATABASE prod"` slipped past `"DROP DATABASE *"`. `matchBashDdl` now trims first. +test.
- **MEDIUM — Map/Set digest collision**: `Object.keys(new Map())` is `[]`, so distinct Maps/Sets both stringified to `{}` and collided; now serialized distinctly. +test.
- **MEDIUM — tool-call metadata recorded pre-hook args**: both dispatch sites now build `ctx` from the post-hook `finalArgs` (passing `input.session.id`/`toolCallId` directly to the hook), so the UI/state reflects what actually ran.
- **MEDIUM — `tool.execute.after` on hard deny**: documented that skipping it is intentional (the tool never executed; the security invariant holds regardless).

**Deferred (with rationale)**: the `export namespace` → flat-export convention (P1). No test monkeypatches HardPolicy, so it's *safe* to convert — but `export namespace` is the established pattern across 9 sibling `src/altimate/` modules, so converting this one file alone breaks local consistency for a marginal tree-shaking gain on an always-fully-used module. Better as a subtree-wide follow-up.

24 pass, typecheck clean, markers clean.

dev-punia-altimate:

No supported files changed.

PR #1030, opened 2026-07-22
feat: dbt_project_health tool — reuse altimate-core dbt health checks

mdesmet:
## Summary

Wires the new Rust dbt **health checks** (`altimate_core::dbt::health`, added in AltimateAI/altimate-core-internal#762) into altimate-code, so dbt governance / modelling / documentation / test checks run **directly against a dbt project's files** — no compiled `manifest.json` required — through the `@altimateai/altimate-core` napi binding.

## Changes

- **Bridge + dispatcher**: new `altimate_core.dbt_project_health` bridge-method type and a `Dispatcher.register` handler calling `core.dbtProjectHealth(projectDir, configJson?, catalogPath?)`.
- **Tool**: `DbtProjectHealthTool` (`dbt_project_health`) — takes a `project_dir` (+ optional `config_path`, `catalog_path`), runs the checks, and formats findings grouped by severity. Registered in the tool registry and the altimate barrel export.
- **Dep**: bumps `@altimateai/altimate-core` **0.5.1 → 0.8.0** (the version that exports `dbtProjectHealth`).

## Dependency

Requires `@altimateai/altimate-core` **0.8.0**, published from AltimateAI/altimate-core-internal#762 (stacked on #761). Merge/release that first.

## Verification

TypeScript typecheck is clean on all changed files against the 0.8.0 core types (verified by overlaying the freshly-built core `index.d.ts`). End-to-end, the underlying core returns findings for a sample project (25 findings across 13 check codes on the fixture).

---
## Summary by cubic
Adds dbt project health checks that run directly on project files and a deterministic config inference flow, powered by `@altimateai/altimate-core` 0.8.0. You can now generate a per-project `.altimate/dbt-health.yml` and have the health tool pick it up automatically.

- **New Features**
  - Bridge + dispatcher for `altimate_core.dbt_project_health` and `altimate_core.dbt_health_infer_config`.
  - `dbt_project_health` tool: runs checks on files; accepts `project_dir` with optional `config_path`/`catalog_path`; auto-discovers `.altimate/dbt-health.{yml,yaml,json}`; outputs findings with severity counts.
  - `dbt_health_config` tool: infers and writes `.altimate/dbt-health.yml` per project; adds `/dbt-health-config` command to infer then run health; tools registered and exported.

- **Dependencies**
  - Bumps `@altimateai/altimate-core` from `0.5.1` to `0.8.0`.
  - Requires `@altimateai/altimate-core` `0.8.0` to be published first.

<sup>Written for commit fefe41f2868954fba4cab50df78b95da3d921c1e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1030?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added dbt project health tooling to analyze projects and report findings, with support for optional configuration and catalog inputs.
  * Findings are normalized, sorted by severity and code, and summarized with per-severity counts; each finding includes messages and recommendations when available (including file/identifier context when present).
  * Added a dbt health-config capability and command to infer and write `.altimate/dbt-health.yml`, returning the written path(s) and resulting severity counts.

### DIFF
--- a/packages/opencode/package.json
+++ b/packages/opencode/package.json
@@ -78,7 +78,7 @@
     "@ai-sdk/togetherai": "2.0.41",
     "@ai-sdk/vercel": "2.0.39",
     "@ai-sdk/xai": "3.0.82",
-    "@altimateai/altimate-core": "0.5.1",
+    "@altimateai/altimate-core": "0.8.0",
     "@altimateai/drivers": "workspace:*",
     "@aws-sdk/credential-providers": "3.1057.0",
     "@clack/prompts": "1.0.0-alpha.1",
--- a/packages/opencode/src/altimate/native/altimate-core.ts
+++ b/packages/opencode/src/altimate/native/altimate-core.ts
@@ -276,6 +276,14 @@ export function registerAll(): void {
       return fail(e)
     }
   })
+  // Deterministically infer a dbt health-check config (YAML) from a project's conventions.
+  register("altimate_core.dbt_health_infer_config", async (params) => {
+    try {
+      return ok(true, { config: core.dbtHealthInferConfig(params.project_dir) })
+    } catch (e) {
+      return fail(e)
+    }
+  })
   // altimate_change end
 
   // 7. altimate_core.fix
--- a/packages/opencode/src/altimate/native/altimate-core.ts
+++ b/packages/opencode/src/altimate/native/altimate-core.ts
@@ -267,6 +267,15 @@ export function registerAll(): void {
       return fail(e)
     }
   })
+  // dbt project health checks — reads the dbt project files directly (no manifest required).
+  register("altimate_core.dbt_project_health", async (params) => {
+    try {
+      const json = core.dbtProjectHealth(params.project_dir, params.config_json, params.catalog_path)
+      return ok(true, { findings: JSON.parse(json) })
+    } catch (e) {
+      return fail(e)
+    }
+  })
   // altimate_change end
 
   // 7. altimate_core.fix
--- a/packages/opencode/src/altimate/tools/dbt-health-config.ts
+++ b/packages/opencode/src/altimate/tools/dbt-health-config.ts
@@ -1,0 +1,55 @@
+import { mkdir } from "node:fs/promises"
+import { dirname } from "node:path"
+import z from "zod"
+import { Tool } from "../../tool/tool"
+import { Dispatcher } from "../native"
+
+/** Relative path (from the dbt project root) where the inferred config is written. */
+export const DBT_HEALTH_CONFIG_RELPATH = ".altimate/dbt-health.yml"
+
+export const DbtHealthConfigTool = Tool.define("dbt_health_config", {
+  description:
+    "Deterministically infer a dbt health-check config from a dbt project's EXISTING conventions (modal tags/meta keys/tests, configured schemas & databases, naming contracts, distribution-ratcheted thresholds) and write it to <project_dir>/.altimate/dbt-health.yml. This activates the config-driven governance checks (which are no-ops until configured). Per-project: each dbt project gets its own config file. Reproducible — no LLM, same project always yields the same file.",
+  parameters: z.object({
+    project_dir: z.string().describe("Path to the dbt project root (the directory containing dbt_project.yml)"),
+    write: z
+      .boolean()
+      .optional()
+      .describe("Write the config to <project_dir>/.altimate/dbt-health.yml (default true). If false, only return the YAML."),
+  }),
+  async execute(args, _ctx) {
+    const result = await Dispatcher.call("altimate_core.dbt_health_infer_config", {
+      project_dir: args.project_dir,
+    })
+
+    if (!result.success) {
+      const error = result.error ?? "unknown error"
+      return {
+        title: "dbt health config: ERROR",
+        metadata: { error },
+        output: `Failed to infer dbt health config: ${error}`,
+      }
+    }
+
+    const yaml = String((result.data as { config?: unknown }).config ?? "")
+    const outPath = `${args.project_dir.replace(/\/+$/, "")}/${DBT_HEALTH_CONFIG_RELPATH}`
+
+    let written = false
+    if (args.write !== false) {
+      await mkdir(dirname(outPath), { recursive: true })
+      await Bun.write(outPath, yaml)
+      written = true
+    }
+
+    const checkCount = (yaml.match(/^ {2}\S/gm) ?? []).length
+    return {
+      title: written
+        ? `dbt health config written (${checkCount} checks tuned) → ${outPath}`
+        : `dbt health config inferred (${checkCount} checks tuned)`,
+      metadata: { path: outPath, written, project_dir: args.project_dir },
+      output: written
+        ? `Wrote inferred config to ${outPath}:\n\n${yaml}`
+        : `Inferred config (not written):\n\n${yaml}`,
+    }
+  },
+})
--- a/packages/opencode/src/altimate/tools/dbt-project-health.ts
+++ b/packages/opencode/src/altimate/tools/dbt-project-health.ts
@@ -33,10 +33,22 @@ export const DbtProjectHealthTool = Tool.define("dbt_project_health", {
       .describe("Optional path to a dbt catalog.json to power column-aware checks"),
   }),
   async execute(args, _ctx) {
+    // Explicit config_path wins; otherwise auto-discover a per-project inferred config
+    // at <project_dir>/.altimate/dbt-health.{yml,yaml,json} (YAML or JSON both parse).
     let config_json: string | undefined
-    if (args.config_path) {
-      const file = Bun.file(args.config_path)
-      if (await file.exists()) config_json = await file.text()
+    const candidates = args.config_path
+      ? [args.config_path]
+      : [
+          `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.yml`,
+          `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.yaml`,
+          `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.json`,
+        ]
+    for (const candidate of candidates) {
+      const file = Bun.file(candidate)
+      if (await file.exists()) {
+        config_json = await file.text()
+        break
+      }
     }
 
     const result = await Dispatcher.call("altimate_core.dbt_project_health", {
--- a/packages/opencode/src/altimate/tools/dbt-project-health.ts
+++ b/packages/opencode/src/altimate/tools/dbt-project-health.ts
@@ -1,0 +1,88 @@
+import z from "zod"
+import { Tool } from "../../tool/tool"
+import { Dispatcher } from "../native"
+
+/** A single finding returned by `altimate_core.dbt_project_health` (Rust `HealthFinding`). */
+interface HealthFinding {
+  code: string
+  alias: string
+  resource_type: "model" | "source" | "exposure" | "macro" | "project"
+  unique_id?: string
+  file?: string
+  severity: "error" | "warning" | "info"
+  message: string
+  recommendation?: string
+  reason_to_flag?: string
+  metadata?: unknown
+}
+
+const SEVERITY_ORDER: Record<string, number> = { error: 0, warning: 1, info: 2 }
+
+export const DbtProjectHealthTool = Tool.define("dbt_project_health", {
+  description:
+    "Run dbt project health checks (governance, modelling, documentation, tests, sources) directly against a dbt project's files — no compiled manifest.json required. Reads .sql models (via the dbt Jinja AST parser), schema.yml properties and dbt_project.yml. Optionally uses a catalog.json for column-level checks; otherwise columns are inferred from the SQL parser.",
+  parameters: z.object({
+    project_dir: z.string().describe("Path to the dbt project root (the directory containing dbt_project.yml)"),
+    config_path: z
+      .string()
+      .optional()
+      .describe("Optional path to a JSON health-check config (disabled checks, severity overrides, options)"),
+    catalog_path: z
+      .string()
+      .optional()
+      .describe("Optional path to a dbt catalog.json to power column-aware checks"),
+  }),
+  async execute(args, _ctx) {
+    let config_json: string | undefined
+    if (args.config_path) {
+      const file = Bun.file(args.config_path)
+      if (await file.exists()) config_json = await file.text()
+    }
+
+    const result = await Dispatcher.call("altimate_core.dbt_project_health", {
+      project_dir: args.project_dir,
+      config_json,
+      catalog_path: args.catalog_path,
+    })
+
+    if (!result.success) {
+      const error = result.error ?? "unknown error"
+      return {
+        title: "dbt health: ERROR",
+        metadata: { error },
+        output: `Failed to run dbt health checks: ${error}`,
+      }
+    }
+
+    const findings = ((result.data.findings as HealthFinding[] | undefined) ?? []).slice()
+    findings.sort(
+      (a, b) => (SEVERITY_ORDER[a.severity] ?? 3) - (SEVERITY_ORDER[b.severity] ?? 3) || a.code.localeCompare(b.code),
+    )
+
+    const counts = { error: 0, warning: 0, info: 0 } as Record<string, number>
+    for (const f of findings) counts[f.severity] = (counts[f.severity] ?? 0) + 1
+
+    return {
+      title: `dbt health: ${findings.length} finding(s) — ${counts.error} error, ${counts.warning} warning, ${counts.info} info`,
+      metadata: {
+        finding_count: findings.length,
+        error_count: counts.error,
+        warning_count: counts.warning,
+        info_count: counts.info,
+      },
+      output: formatFindings(findings),
+    }
+  },
+})
+
+function formatFindings(findings: HealthFinding[]): string {
+  if (findings.length === 0) return "✓ No dbt health issues found."
+  const lines: string[] = []
+  for (const f of findings) {
+    const where = f.file ? ` (${f.file})` : f.unique_id ? ` (${f.unique_id})` : ""
+    lines.push(`[${f.severity.toUpperCase()}] ${f.code} ${f.alias}${where}`)
+    lines.push(`  ${f.message}`)
+    if (f.recommendation) lines.push(`  → ${f.recommendation}`)
+  }
+  return lines.join("\n")
+}
--- a/.opencode/command/dbt-health-config.md
+++ b/.opencode/command/dbt-health-config.md
@@ -1,0 +1,25 @@
+---
+description: infer a dbt health-check config from the project's own conventions and write .altimate/dbt-health.yml
+agent: build
+subtask: true
+---
+
+Generate a dbt health-check configuration by inferring it from a dbt project's **existing conventions** (deterministic — no guessing).
+
+## Target project
+
+Use `$ARGUMENTS` as the dbt project directory if provided. Otherwise, locate the dbt project in the current worktree by finding the directory that contains a `dbt_project.yml` (run `project_scan` if you need to discover it). If there are **multiple** dbt projects, do this for **each** one — every project gets its own `.altimate/dbt-health.yml`.
+
+## Steps
+
+1. For each target project directory, call the **`dbt_health_config`** tool:
+   `dbt_health_config({ project_dir: "<dir>" })`
+   This deterministically infers the config from the project (modal tags / meta keys / tests, configured schemas & databases, naming contracts, distribution-ratcheted thresholds) and writes it to `<dir>/.altimate/dbt-health.yml`.
+
+2. Then run **`dbt_project_health`** on the same directory:
+   `dbt_project_health({ project_dir: "<dir>" })`
+   It auto-discovers the freshly written `.altimate/dbt-health.yml`, so the report now reflects the inferred rules.
+
+3. Summarize for the user: the path(s) written, which config-driven checks were activated (tags, meta keys, tests, naming contracts, parent schema/database whitelists, thresholds), and the resulting finding counts by severity. Tell them the file is committed with the project and can be hand-edited — re-running this command regenerates it deterministically.
+
+Do not fabricate values; everything comes from the two tools above.
--- a/packages/opencode/src/altimate/index.ts
+++ b/packages/opencode/src/altimate/index.ts
@@ -40,6 +40,7 @@ export * from "./tools/altimate-core-validate"
 export * from "./tools/dbt-lineage"
 export * from "./tools/dbt-manifest"
 export * from "./tools/dbt-profiles"
+export * from "./tools/dbt-project-health"
 
 export * from "./tools/finops-analyze-credits"
 export * from "./tools/finops-expensive-queries"
--- a/packages/opencode/src/altimate/index.ts
+++ b/packages/opencode/src/altimate/index.ts
@@ -41,6 +41,7 @@ export * from "./tools/dbt-lineage"
 export * from "./tools/dbt-manifest"
 export * from "./tools/dbt-profiles"
 export * from "./tools/dbt-project-health"
+export * from "./tools/dbt-health-config"
 
 export * from "./tools/finops-analyze-credits"
 export * from "./tools/finops-expensive-queries"
--- a/packages/opencode/src/altimate/native/types.ts
+++ b/packages/opencode/src/altimate/native/types.ts
@@ -1244,6 +1244,10 @@ export const BridgeMethods = {
     params: { base_sql: string; head_sql: string }
     result: AltimateCoreResult
   },
+  "altimate_core.dbt_project_health": {} as {
+    params: { project_dir: string; config_json?: string; catalog_path?: string }
+    result: AltimateCoreResult
+  },
   // altimate_change end
   // --- altimate-core Phase 1 (P0) ---
   "altimate_core.fix": {} as { params: AltimateCoreFixParams; result: AltimateCoreResult },
--- a/packages/opencode/src/altimate/native/types.ts
+++ b/packages/opencode/src/altimate/native/types.ts
@@ -1248,6 +1248,10 @@ export const BridgeMethods = {
     params: { project_dir: string; config_json?: string; catalog_path?: string }
     result: AltimateCoreResult
   },
+  "altimate_core.dbt_health_infer_config": {} as {
+    params: { project_dir: string }
+    result: AltimateCoreResult
+  },
   // altimate_change end
   // --- altimate-core Phase 1 (P0) ---
   "altimate_core.fix": {} as { params: AltimateCoreFixParams; result: AltimateCoreResult },
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -70,6 +70,7 @@ import { McpDiscoverTool } from "../altimate/tools/mcp-discover"
 
 import { DbtManifestTool } from "../altimate/tools/dbt-manifest"
 import { DbtProjectHealthTool } from "../altimate/tools/dbt-project-health"
+import { DbtHealthConfigTool } from "../altimate/tools/dbt-health-config"
 // altimate_change start - import dbt unit test generation tool
 import { DbtUnitTestGenTool } from "../altimate/tools/dbt-unit-test-gen"
 // altimate_change end
@@ -407,6 +408,7 @@ export namespace ToolRegistry {
       // altimate_change end
       DbtProfilesTool,
       DbtProjectHealthTool,
+      DbtHealthConfigTool,
       DbtLineageTool,
       SchemaIndexTool,
       SchemaSearchTool,
--- a/packages/opencode/src/tool/registry.ts
+++ b/packages/opencode/src/tool/registry.ts
@@ -69,6 +69,7 @@ import { WarehouseDiscoverTool } from "../altimate/tools/warehouse-discover"
 import { McpDiscoverTool } from "../altimate/tools/mcp-discover"
 
 import { DbtManifestTool } from "../altimate/tools/dbt-manifest"
+import { DbtProjectHealthTool } from "../altimate/tools/dbt-project-health"
 // altimate_change start - import dbt unit test generation tool
 import { DbtUnitTestGenTool } from "../altimate/tools/dbt-unit-test-gen"
 // altimate_change end
@@ -405,6 +406,7 @@ export namespace ToolRegistry {
       DbtUnitTestGenTool,
       // altimate_change end
       DbtProfilesTool,
+      DbtProjectHealthTool,
       DbtLineageTool,
       SchemaIndexTool,
       SchemaSearchTool,


packages/opencode/src/altimate/tools/dbt-project-health.ts
```diff
@@ -0,0 +1,100 @@
...
+          `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.yaml`,
+          `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.json`,
+        ]
+    for (const candidate of candidates) {
+      const file = Bun.file(candidate)
+      if (await file.exists()) {
+        config_json = await file.text()
+        break
+      }
+    }
```
dev-punia-altimate, line 52: **[🟠 MEDIUM]** If `args.config_path` is explicitly provided but the file does not exist, this logic will silently proceed with `config_json = undefined`. It is recommended to throw an error or return an error response when an explicitly provided configuration file is not found, to avoid running health checks with unintended settings.

**Suggested change:**
```suggestion
    let config_json: string | undefined
    if (args.config_path) {
      const file = Bun.file(args.config_path)
      if (!(await file.exists())) {
        return {
          title: "dbt health: ERROR",
          metadata: { error: `Config file not found: ${args.config_path}` },
          output: `Failed to find config file at ${args.config_path}`,
        }
      }
      config_json = await file.text()
    } else {
      const candidates = [
        `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.yml`,
        `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.yaml`,
        `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.json`,
      ]
      for (const candidate of candidates) {
        const file = Bun.file(candidate)
        if (await file.exists()) {
          config_json = await file.text()
          break
        }
      }
    }
```

packages/opencode/src/altimate/tools/dbt-project-health.ts
```diff
@@ -0,0 +1,100 @@
...
+  },
+})
+
+function formatFindings(findings: HealthFinding[]): string {
+  if (findings.length === 0) return "✓ No dbt health issues found."
+  const lines: string[] = []
+  for (const f of findings) {
+    const where = f.file ? ` (${f.file})` : f.unique_id ? ` (${f.unique_id})` : ""
+    lines.push(`[${f.severity.toUpperCase()}] ${f.code} ${f.alias}${where}`)
```
dev-punia-altimate, line 95: **[🔵 LOW]** According to the review checklist, nested ternary expressions are not allowed. Please refactor this to use an `if...else` statement or logical operators to improve readability.

**Suggested change:**
```suggestion
  for (const f of findings) {
    let where = ""
    if (f.file) {
      where = ` (${f.file})`
    } else if (f.unique_id) {
      where = ` (${f.unique_id})`
    }
    lines.push(`[${f.severity.toUpperCase()}] ${f.code} ${f.alias}${where}`)
```

packages/opencode/src/altimate/tools/dbt-health-config.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (!result.success) {
+      const error = result.error ?? "unknown error"
+      return {
+        title: "dbt health config: ERROR",
+        metadata: { error },
+        output: `Failed to infer dbt health config: ${error}`,
+      }
+    }
+
+    const yaml = String((result.data as { config?: unknown }).config ?? "")
```
dev-punia-altimate, line 34: **[🔴 HIGH]** Potential runtime `TypeError`. Accessing `.config` directly on the typecasted `result.data` will throw a `TypeError: Cannot read properties of undefined` if `result.data` is `null` or `undefined` (even when `result.success` is true). Please use optional chaining (`?.config`) to ensure safe access.

**Suggested change:**
```suggestion
    const yaml = String((result.data as { config?: unknown })?.config ?? "")
```

packages/opencode/src/altimate/tools/dbt-health-config.ts
```diff
@@ -0,0 +1,55 @@
...
+    const outPath = `${args.project_dir.replace(/\/+$/, "")}/${DBT_HEALTH_CONFIG_RELPATH}`
```
dev-punia-altimate, line 35: **[🔴 HIGH]** Unsafe path construction and potential directory traversal. The output path is constructed using simple string concatenation. If the AI agent provides an empty string, `/`, or `../` for `project_dir`, it could attempt to write to unexpected system locations outside the intended workspace. 

Consider importing `join` from `node:path`, using `join(args.project_dir, DBT_HEALTH_CONFIG_RELPATH)`, and adding a check to ensure the resolved path remains within the allowed workspace context.

**Suggested change:**
```suggestion
    // Ensure `join` is imported from `node:path`
    const outPath = join(args.project_dir, DBT_HEALTH_CONFIG_RELPATH)
    // Recommended: add validation to ensure `outPath` is within the workspace root
```

packages/opencode/src/altimate/tools/dbt-health-config.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    let written = false
+    if (args.write !== false) {
+      await mkdir(dirname(outPath), { recursive: true })
+      await Bun.write(outPath, yaml)
+      written = true
+    }
```
dev-punia-altimate, line 42: **[🔴 HIGH]** Unhandled exceptions in file system operations and inconsistent FS API usage.

1. **Unhandled Exceptions**: `mkdir` and file write operations are not enclosed in a `try...catch` block. If the tool encounters permission issues or invalid paths, these async calls will throw exceptions that could crash the execution. They should be caught to gracefully return an error payload (similar to `!result.success`).
2. **Inconsistent API**: The code imports `mkdir` from standard `node:fs/promises` but uses the runtime-specific `Bun.write`. For consistency within the file and to avoid potential `ReferenceError`s outside of Bun, it's recommended to use `writeFile` from `node:fs/promises`.

**Suggested change:**
```suggestion
    let written = false
    if (args.write !== false) {
      try {
        await mkdir(dirname(outPath), { recursive: true })
        // Ensure `writeFile` is imported from `node:fs/promises` alongside `mkdir`
        await writeFile(outPath, yaml)
        written = true
      } catch (err) {
        return {
          title: "dbt health config: ERROR",
          metadata: { error: String(err) },
          output: `Failed to write config file to ${outPath}: ${err}`,
        }
      }
    }
```

dev-punia-altimate (commented):
## 🤖 Code Review — OpenCodeReview (Gemini) — 5 finding(s)

- 5 anchored to a line (posted inline when the comment stream is on)
- 0 without a line anchor

#### 1. `packages/opencode/src/altimate/tools/dbt-project-health.ts` (L38-L52)

**[🟠 MEDIUM]** If `args.config_path` is explicitly provided but the file does not exist, this logic will silently proceed with `config_json = undefined`. It is recommended to throw an error or return an error response when an explicitly provided configuration file is not found, to avoid running health checks with unintended settings.

**Suggested change:**
```suggestion
    let config_json: string | undefined
    if (args.config_path) {
      const file = Bun.file(args.config_path)
      if (!(await file.exists())) {
        return {
          title: "dbt health: ERROR",
          metadata: { error: `Config file not found: ${args.config_path}` },
          output: `Failed to find config file at ${args.config_path}`,
        }
      }
      config_json = await file.text()
    } else {
      const candidates = [
        `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.yml`,
        `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.yaml`,
        `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.json`,
      ]
      for (const candidate of candidates) {
        const file = Bun.file(candidate)
        if (await file.exists()) {
          config_json = await file.text()
          break
        }
      }
    }
```

#### 2. `packages/opencode/src/altimate/tools/dbt-project-health.ts` (L93-L95)

**[🔵 LOW]** According to the review checklist, nested ternary expressions are not allowed. Please refactor this to use an `if...else` statement or logical operators to improve readability.

**Suggested change:**
```suggestion
  for (const f of findings) {
    let where = ""
    if (f.file) {
      where = ` (${f.file})`
    } else if (f.unique_id) {
      where = ` (${f.unique_id})`
    }
    lines.push(`[${f.severity.toUpperCase()}] ${f.code} ${f.alias}${where}`)
```

#### 3. `packages/opencode/src/altimate/tools/dbt-health-config.ts` (L34)

**[🔴 HIGH]** Potential runtime `TypeError`. Accessing `.config` directly on the typecasted `result.data` will throw a `TypeError: Cannot read properties of undefined` if `result.data` is `null` or `undefined` (even when `result.success` is true). Please use optional chaining (`?.config`) to ensure safe access.

**Suggested change:**
```suggestion
    const yaml = String((result.data as { config?: unknown })?.config ?? "")
```

#### 4. `packages/opencode/src/altimate/tools/dbt-health-config.ts` (L35)

**[🔴 HIGH]** Unsafe path construction and potential directory traversal. The output path is constructed using simple string concatenation. If the AI agent provides an empty string, `/`, or `../` for `project_dir`, it could attempt to write to unexpected system locations outside the intended workspace. 

Consider importing `join` from `node:path`, using `join(args.project_dir, DBT_HEALTH_CONFIG_RELPATH)`, and adding a check to ensure the resolved path remains within the allowed workspace context.

**Suggested change:**
```suggestion
    // Ensure `join` is imported from `node:path`
    const outPath = join(args.project_dir, DBT_HEALTH_CONFIG_RELPATH)
    // Recommended: add validation to ensure `outPath` is within the workspace root
```

#### 5. `packages/opencode/src/altimate/tools/dbt-health-config.ts` (L37-L42)

**[🔴 HIGH]** Unhandled exceptions in file system operations and inconsistent FS API usage.

1. **Unhandled Exceptions**: `mkdir` and file write operations are not enclosed in a `try...catch` block. If the tool encounters permission issues or invalid paths, these async calls will throw exceptions that could crash the execution. They should be caught to gracefully return an error payload (similar to `!result.success`).
2. **Inconsistent API**: The code imports `mkdir` from standard `node:fs/promises` but uses the runtime-specific `Bun.write`. For consistency within the file and to avoid potential `ReferenceError`s outside of Bun, it's recommended to use `writeFile` from `node:fs/promises`.

**Suggested change:**
```suggestion
    let written = false
    if (args.write !== false) {
      try {
        await mkdir(dirname(outPath), { recursive: true })
        // Ensure `writeFile` is imported from `node:fs/promises` alongside `mkdir`
        await writeFile(outPath, yaml)
        written = true
      } catch (err) {
        return {
          title: "dbt health config: ERROR",
          metadata: { error: String(err) },
          output: `Failed to write config file to ${outPath}: ${err}`,
        }
      }
    }
```

</details>

dev-punia-altimate:

No supported files changed.

packages/opencode/package.json
```diff
@@ -78,7 +78,7 @@
     "@ai-sdk/togetherai": "2.0.41",
     "@ai-sdk/vercel": "2.0.39",
     "@ai-sdk/xai": "3.0.82",
-    "@altimateai/altimate-core": "0.5.1",
+    "@altimateai/altimate-core": "0.8.0",
```
sahrizvi, line 81: **CRITICAL — `0.8.0` is not published, and `bun.lock` still pins `0.5.1`**

```
npm view @altimateai/altimate-core versions
→ … "0.5.1", "0.6.0", "0.7.0"     dist-tags.latest = 0.7.0
```

`0.8.0` does not exist on npm today. Separately, `bun.lock` still resolves `0.5.1` at lines 273 and 642 plus all five platform binaries, and the lockfile isn't part of this PR (`git log -1 -- bun.lock` points at a pre-existing commit).

Every CI job runs plain `bun install` (`.github/workflows/ci.yml:100,108,287,348,408,465,505`), and Bun applies frozen-lockfile behaviour when `CI` is set — so the manifest/lock mismatch fails the install step. If it did install from the current lock it would fetch `0.5.1`, which exports neither `dbtProjectHealth` nor `dbtHealthInferConfig`, and both tools would return `TypeError: core.dbtProjectHealth is not a function` on every call.

The PR description already flags the ordering dependency, so this is partly known — noting it because the branch is red until that release lands.

**Fix**: publish `0.8.0`, then `bun install` and commit the regenerated `bun.lock` with the root entry and all five platform optional deps at `0.8.0`.

One wrinkle: `bunfig.toml` sets `minimumReleaseAge = 259200` (3 days) and this package is **not** in `minimumReleaseAgeExcludes`, so a freshly published `0.8.0` will still be refused for three days after publish. Add it to the excludes list or plan the timing.

packages/opencode/src/altimate/native/altimate-core.ts
```diff
@@ -267,6 +267,23 @@ export function registerAll(): void {
...
+    }
+  })
+  // Deterministically infer a dbt health-check config (YAML) from a project's conventions.
+  register("altimate_core.dbt_health_infer_config", async (params) => {
+    try {
+      return ok(true, { config: core.dbtHealthInferConfig(params.project_dir) })
+    } catch (e) {
+      return fail(e)
+    }
+  })
```
sahrizvi, line 286: **CRITICAL — these two registrations break `altimate-core-native.test.ts`**

`packages/opencode/test/altimate/altimate-core-native.test.ts:176-231` keeps a hand-maintained `ALL_METHODS` array (43 entries) and asserts an exact count:

```ts
const coreCount = registered.filter((m) => m.startsWith("altimate_core.")).length
expect(coreCount).toBe(ALL_METHODS.length)
```

Registering two more methods without adding them to the array makes it 45 vs 43. This is a deterministic CI failure, not a coverage gap.

**Fix**: add `"altimate_core.dbt_project_health"` and `"altimate_core.dbt_health_infer_config"` to `ALL_METHODS`.

packages/opencode/src/altimate/native/altimate-core.ts
```diff
@@ -267,6 +267,23 @@ export function registerAll(): void {
       return fail(e)
     }
   })
+  // dbt project health checks — reads the dbt project files directly (no manifest required).
+  register("altimate_core.dbt_project_health", async (params) => {
+    try {
+      const json = core.dbtProjectHealth(params.project_dir, params.config_json, params.catalog_path)
```
sahrizvi, line 273: **MINOR — no version guard on the new exports**

This file already establishes the guard pattern for exactly this situation at line 176:

```ts
params.base_sql && typeof core.lintDiff === "function"
```

added because `lintDiff` postdates `0.5.1`. The surrounding `try/catch` does convert a missing-export `TypeError` into the normal failure envelope, so this is diagnostics quality rather than a crash — but the operator sees a raw `TypeError` string instead of an actionable message.

**Fix**: guard with `typeof core.dbtProjectHealth === "function"` and fail with `"dbt_project_health requires @altimateai/altimate-core >= 0.8.0"`. Same for `dbtHealthInferConfig` below.

packages/opencode/src/altimate/tools/dbt-health-config.ts
sahrizvi, line 42: **CRITICAL — this write has none of the repo's write guards**

`project_dir` comes straight from the model. There is no `path.resolve` against `Instance.directory`, no containment check, no symlink resolution, and no permission prompt. `WriteTool` (`src/tool/write.ts:64-70`) resolves against `Instance.directory` and then gates every write behind `assertExternalDirectoryEffect` and `assertSensitiveWriteEffect`.

This is the first write-capable tool under `src/altimate/tools/` — no other file there calls `Bun.write`/`writeFile` — so it establishes a path around the guard layer.

Three concrete failure modes:
- an absolute or `../`-containing `project_dir` creates directories and writes files outside the workspace with no prompt;
- a `.altimate` directory or `.altimate/dbt-health.yml` **symlink** inside an otherwise legitimate project is followed and its external target truncated;
- `mkdir(…, { recursive: true })` creates the whole tree on the way there.

**Fix**: resolve `project_dir` against `Instance.directory`, build the path with `path.join`, canonicalize and verify containment, then gate the write. This tool uses the Promise/Zod adapter, so call the Promise-style helpers — `assertExternalDirectory(ctx, target, options)` / `assertExternalDirectoryLegacy(...)` / `assertSensitiveWrite(ctx, target)` (`src/tool/external-directory.ts:54,89,113`) — not the Effect-only variants. Re-validate symlinked path components immediately before writing.

packages/opencode/src/altimate/tools/dbt-health-config.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export const DbtHealthConfigTool = Tool.define("dbt_health_config", {
+  description:
+    "Deterministically infer a dbt health-check config from a dbt project's EXISTING conventions (modal tags/meta keys/tests, configured schemas & databases, naming contracts, distribution-ratcheted thresholds) and write it to <project_dir>/.altimate/dbt-health.yml. This activates the config-driven governance checks (which are no-ops until configured). Per-project: each dbt project gets its own config file. Reproducible — no LLM, same project always yields the same file.",
+  parameters: z.object({
+    project_dir: z.string().describe("Path to the dbt project root (the directory containing dbt_project.yml)"),
+    write: z
+      .boolean()
+      .optional()
+      .describe("Write the config to <project_dir>/.altimate/dbt-health.yml (default true). If false, only return the YAML."),
```
sahrizvi, line 18: **MAJOR — hand-edited config is silently destroyed**

`write` defaults to true and `Bun.write` at line 40 unconditionally replaces an existing `.altimate/dbt-health.yml` — no diff, no backup, no `overwrite` argument, and no mention in the tool description.

`.opencode/command/dbt-health-config.md:23` explicitly tells users the file "is committed with the project and can be hand-edited", so the advertised workflow is precisely the one that erases their edits.

**Fix**: add `overwrite: boolean` defaulting to `false`, or open with `wx` and fail when the file exists. At minimum, report that the file already existed and what changed.

packages/opencode/src/altimate/tools/dbt-health-config.ts
```diff
@@ -0,0 +1,55 @@
...
+    const outPath = `${args.project_dir.replace(/\/+$/, "")}/${DBT_HEALTH_CONFIG_RELPATH}`
+
+    let written = false
+    if (args.write !== false) {
+      await mkdir(dirname(outPath), { recursive: true })
+      await Bun.write(outPath, yaml)
+      written = true
+    }
+
+    const checkCount = (yaml.match(/^ {2}\S/gm) ?? []).length
```
sahrizvi, line 44: **MINOR — `checkCount` is a fragile heuristic**

`(yaml.match(/^ {2}\S/gm) ?? []).length` counts every key indented exactly two spaces at any nesting depth, not tuned checks. It may match the current emitter's output, but any nesting or indentation change on the core side silently turns the title's `(N checks tuned)` into a wrong number that nobody will notice.

**Fix**: have the core return the count, parse the YAML, or drop the claim from the title.

packages/opencode/src/altimate/tools/dbt-health-config.ts
```diff
@@ -0,0 +1,55 @@
+import { mkdir } from "node:fs/promises"
+import { dirname } from "node:path"
+import z from "zod"
+import { Tool } from "../../tool/tool"
+import { Dispatcher } from "../native"
+
+/** Relative path (from the dbt project root) where the inferred config is written. */
+export const DBT_HEALTH_CONFIG_RELPATH = ".altimate/dbt-health.yml"
```
sahrizvi, line 8: **MINOR — this constant isn't used by the reader**

`DBT_HEALTH_CONFIG_RELPATH` is exported here, but `dbt-project-health.ts:42-44` rebuilds `.altimate/dbt-health.*` independently as three string literals. One rename desynchronises write and discovery, and the failure is silent: the config just stops being found.

**Fix**: export the directory and basename (or a `candidates(projectDir)` helper) from one place and import it in the reader.

packages/opencode/src/altimate/tools/dbt-project-health.ts
sahrizvi, line 52: **MAJOR — an explicit `config_path` that doesn't exist is silently ignored**

When `config_path` is supplied it becomes the sole candidate, and if `file.exists()` is false the loop just falls through leaving `config_json` undefined. The checks then run fully unconfigured and the report gives no hint. The auto-discovery fallback is skipped entirely in that branch too.

For a governance tool this is the worst failure mode: a typo'd path yields `✓ No dbt health issues found.` for a project whose policy never loaded. Anyone reading that output concludes the project is clean.

**Fix**: if `config_path` was passed explicitly and is missing or unreadable, return an error naming the path. Either way, report the loaded config path (or `"none"`) in `metadata` so the result is self-describing.

packages/opencode/src/altimate/tools/dbt-project-health.ts
```diff
@@ -0,0 +1,100 @@
...
+        config_json = await file.text()
+        break
+      }
+    }
+
+    const result = await Dispatcher.call("altimate_core.dbt_project_health", {
+      project_dir: args.project_dir,
+      config_json,
+      catalog_path: args.catalog_path,
+    })
```
sahrizvi, line 58: **MAJOR — relative paths resolve against `process.cwd()`, not the active project**

`project_dir`, `config_path`, and `catalog_path` are passed straight to Bun and to Rust. This codebase models the current project as `Instance.directory` — the pattern used by `write.ts:65-67`, `ls.ts`, `glob.ts`, and `bash.ts` — which can differ from `process.cwd()` in server, desktop, attach, and multi-instance execution.

Users will reasonably expect `config_path` / `catalog_path` to be project-relative; today they're process-relative, and neither the parameter descriptions nor the returned metadata say which.

**Fix**:
```ts
const projectDir = path.resolve(Instance.directory, args.project_dir)
const configPath = args.config_path ? path.resolve(projectDir, args.config_path) : undefined
const catalogPath = args.catalog_path ? path.resolve(projectDir, args.catalog_path) : undefined
```
Pass only normalized paths to the dispatcher and report them in metadata.

packages/opencode/src/altimate/tools/dbt-project-health.ts
```diff
@@ -0,0 +1,100 @@
...
+    config_path: z
+      .string()
+      .optional()
+      .describe("Optional path to a JSON health-check config (disabled checks, severity overrides, options)"),
+    catalog_path: z
+      .string()
+      .optional()
+      .describe("Optional path to a dbt catalog.json to power column-aware checks"),
+  }),
+  async execute(args, _ctx) {
```
sahrizvi, line 35: **MINOR — `execute` isn't wrapped in try/catch**

`Bun.file(candidate).text()` throws on an unreadable config, and `Dispatcher.call` throws `No native handler for …` (`native/dispatcher.ts:48-50`) if lazy registration failed. In `dbt-health-config.ts`, `mkdir` throws `EACCES`/`ENOTDIR` and `Bun.write` throws if the target is a directory.

The framework's `Effect.onError` hook does record and surface these, so nothing is lost — but they bypass the structured `{title, metadata, output}` error shape both tools already implement for the `!result.success` path. Sibling `dbt-manifest.ts:13-28` wraps its whole body.

Also: the soft-error returns omit `metadata.success: false`, which `dbt-profiles.ts:27,50` sets — so telemetry records these as successful calls despite the `ERROR` title.

**Fix**: wrap each `execute` body and return the same error shape with `success: false` in metadata.

packages/opencode/src/altimate/tools/dbt-project-health.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export const DbtProjectHealthTool = Tool.define("dbt_project_health", {
+  description:
+    "Run dbt project health checks (governance, modelling, documentation, tests, sources) directly against a dbt project's files — no compiled manifest.json required. Reads .sql models (via the dbt Jinja AST parser), schema.yml properties and dbt_project.yml. Optionally uses a catalog.json for column-level checks; otherwise columns are inferred from the SQL parser.",
+  parameters: z.object({
+    project_dir: z.string().describe("Path to the dbt project root (the directory containing dbt_project.yml)"),
+    config_path: z
+      .string()
+      .optional()
+      .describe("Optional path to a JSON health-check config (disabled checks, severity overrides, options)"),
```
sahrizvi, line 29: **MINOR — this description is inaccurate and incomplete**

It says "path to a JSON health-check config", but auto-discovery below accepts `.yml`/`.yaml`/`.json` and the sibling `dbt_health_config` tool writes **YAML**. As written, the model is steered away from passing the very file it just generated.

The description also never mentions that a project-local config is auto-discovered when this is omitted, so the model can't reason about when to pass it.

**Fix**: `"Optional path to a health-check config (YAML or JSON). When omitted, auto-discovers <project_dir>/.altimate/dbt-health.{yml,yaml,json}."`

packages/opencode/src/altimate/tools/dbt-project-health.ts
```diff
@@ -0,0 +1,100 @@
...
+  async execute(args, _ctx) {
+    // Explicit config_path wins; otherwise auto-discover a per-project inferred config
+    // at <project_dir>/.altimate/dbt-health.{yml,yaml,json} (YAML or JSON both parse).
+    let config_json: string | undefined
+    const candidates = args.config_path
+      ? [args.config_path]
+      : [
+          `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.yml`,
+          `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.yaml`,
+          `${args.project_dir.replace(/\/+$/, "")}/.altimate/dbt-health.json`,
```
sahrizvi, line 44: **MINOR — build these with `node:path.join`**

`${args.project_dir.replace(/\/+$/, "")}/…` hand-rolls path joining in three places here and once more in `dbt-health-config.ts:35`. It doesn't normalize `..` segments and assumes POSIX separators.

**Fix**: `path.join(projectDir, ".altimate", "dbt-health.yml")`. `dbt-health-config.ts` already imports `dirname` from `node:path`.

packages/opencode/src/altimate/tools/dbt-project-health.ts
```diff
@@ -0,0 +1,100 @@
...
+  alias: string
+  resource_type: "model" | "source" | "exposure" | "macro" | "project"
+  unique_id?: string
+  file?: string
+  severity: "error" | "warning" | "info"
+  message: string
+  recommendation?: string
+  reason_to_flag?: string
+  metadata?: unknown
+}
```
sahrizvi, line 17: **MINOR — `HealthFinding` is an unvalidated local guess**

This interface is hand-declared here with no link to the Rust `HealthFinding` struct and no runtime validation — the shape is only asserted via a cast at line 69. Every other bridge method has a typed result interface in `native/types.ts`. If the core renames or drops a field, TypeScript won't catch it and the formatter silently emits `undefined` text.

**Fix**: move the type into `native/types.ts` as part of a `DbtProjectHealthResult`, or parse the JSON through a `z.object()` schema in the handler.

packages/opencode/src/altimate/tools/dbt-project-health.ts
```diff
@@ -0,0 +1,100 @@
...
+  code: string
+  alias: string
+  resource_type: "model" | "source" | "exposure" | "macro" | "project"
+  unique_id?: string
+  file?: string
+  severity: "error" | "warning" | "info"
+  message: string
+  recommendation?: string
+  reason_to_flag?: string
+  metadata?: unknown
```
sahrizvi, line 16: **MINOR — these two fields are declared but never rendered**

`formatFindings` (lines 90-100) emits `code`, `alias`, `file`/`unique_id`, `message`, and `recommendation` — `reason_to_flag` and `metadata` never reach the output. `reason_to_flag` is arguably the single most useful field in a governance report, since it's what tells the reader *why* a rule fired.

**Fix**: render `reason_to_flag` (it's what makes a finding actionable), or drop both from the interface.

packages/opencode/src/altimate/tools/dbt-project-health.ts
```diff
@@ -0,0 +1,100 @@
...
+  if (findings.length === 0) return "✓ No dbt health issues found."
+  const lines: string[] = []
+  for (const f of findings) {
+    const where = f.file ? ` (${f.file})` : f.unique_id ? ` (${f.unique_id})` : ""
+    lines.push(`[${f.severity.toUpperCase()}] ${f.code} ${f.alias}${where}`)
+    lines.push(`  ${f.message}`)
+    if (f.recommendation) lines.push(`  → ${f.recommendation}`)
+  }
+  return lines.join("\n")
+}
```
sahrizvi, line 100: **MINOR — no domain-level cap on output**

`formatFindings` emits 2–3 lines per finding with no limit. This is bounded in practice: the shared `Tool.define` wrapper applies `Truncate.Service` at `MAX_LINES = 2000` / `MAX_BYTES = 50 * 1024` (`src/tool/truncate.ts:15-16`) and spills the full text to a file, and because findings are sorted error-first the head truncation keeps the important ones while `metadata` counts stay accurate. That's a good outcome largely by accident of the sort order.

What's missing is that the model is never told it's seeing a partial list.

**Fix**: cap at a domain-appropriate number and append `"… N more findings omitted"`, so the truncation is explicit rather than inferred.

sahrizvi (commented):
## Review: 3 critical, 4 major, 10 minor

The design is sound and both tools follow the established `Tool.define` + `Dispatcher.call` + `{title, metadata, output}` shape closely. Three things block merge: a dependency that isn't published yet, a test this PR deterministically breaks, and a file write that sidesteps the repo's permission layer.

Inline comments below cover the critical/major/minor findings. Issues that don't map to a specific line — plus nits, one item needing verification against the core, and what's done well — are in a follow-up comment.

sahrizvi:
## Remaining findings — not line-attributable, plus nits

### MAJOR — no tests for either tool

Neither `dbt_project_health` nor `dbt_health_config` has a test file, in a package with a substantial `test/altimate/` suite where every sibling bridge method is covered. Untested paths:

- config auto-discovery precedence (`.yml` → `.yaml` → `.json`) and explicit-path-wins
- the `write: false` branch, and overwrite behaviour when `.altimate/dbt-health.yml` already exists
- severity sorting, the empty-findings path (`✓ No dbt health issues found.`), unknown severities
- a non-dbt `project_dir`, and a dispatcher failure (`success: false`)
- both bridge handlers' exact argument order and JSON decoding

Worth adding beyond the basics: absolute and `..` `project_dir`; a symlinked `.altimate` directory and output file; an explicit config that is missing, unreadable, or malformed; `mkdir`/write failure; a large finding set exercising truncation; and an install smoke test proving `0.8.0` and all five platform binaries resolve.

### Needs verification against the core (couldn't confirm from this repo)

The bridge passes **YAML** file contents in a parameter named `config_json` (`dbt-project-health.ts:39-52` → `types.ts:1248`). The inline comment asserts "YAML or JSON both parse". If the core instead falls back to a default config when it can't parse the input — rather than returning an error — this compounds the `config_path` issue above: a malformed or wrong-format config produces a clean, empty report rather than a failure.

Two things to confirm on the core side:
1. Does `dbt_project_health` actually parse YAML for `config_json`? If so, rename the parameter to `config` — the current name will keep misleading callers.
2. Does an unparseable config error, or silently degrade to defaults? If it degrades, surface the parse error through the envelope.

### NITs

- **Non-dbt `project_dir` produces a Rust-worded error.** The native parser does validate the project and the bridge returns it via `result.success === false`, so this is message quality rather than a crash — but `project-scan.ts` already has `detectDbtProject()` if a friendlier pre-check is wanted.
- **Registration ordering.** In `registry.ts:72-73` the two new imports sit between `DbtManifestTool` and the `altimate_change` marker for `DbtUnitTestGenTool`, without markers of their own; `altimate/index.ts:43-44` and the registry list both put `dbt-project-health` before `dbt-health-config`, breaking the otherwise-alphabetical grouping.
- **`catalog_path` isn't reachable from the command flow.** `dbt_health_config` has no `catalog_path` passthrough, so step 2 of `/dbt-health-config` can't offer column-aware checks in the same invocation.

### Two things worth explicitly retiring

- `result.data.findings` needs **no** null guard — `AltimateCoreResult.data` is non-optional (`types.ts:831-835`) and always populated by both `ok()` and `fail()` (`altimate-core.ts:36-42`).
- The stale `"all 34 altimate_core.* bridge methods"` module comment at `native/altimate-core.ts:2` is pre-existing (already wrong at the base commit), not introduced here — though this PR does make it wronger, if someone wants to fix it in passing.

### What's done well

- Both tools follow the established `Tool.define` + `Dispatcher.call` + `{title, metadata, output}` shape used across ~40 sibling altimate tools. Nothing bespoke, nothing to relearn.
- The bridge handlers reuse the `ok()`/`fail()` envelope correctly and preserve the "success = handler didn't throw; semantics live in `data`" contract, so a project with findings isn't misreported as a tool failure.
- Deterministic severity sort with a `localeCompare` tiebreak gives stable, diffable output — and it's what makes the framework's head-truncation land on the right findings.
- `types.ts` entries match the handler param shapes exactly, with `config_json` / `catalog_path` correctly modelled as optional.
- Registration is complete and consistent across dispatcher, tool registry, and barrel export — easy to get partially wired, and this isn't.
- Config discovery has a sensible deterministic precedence, and `/dbt-health-config` chains infer-then-run coherently, handles multi-project repos, and explicitly forbids fabricating values.

PR #1046, opened 2026-07-28
feat: first-run activation menu + bundled dbt sample

### LINKED ISSUE #1045: First-run activation menu + bundled dbt sample project
## Problem

After the first-run auth + scan-gate flow, a user who dismisses the
scan or lands on an empty environment ends up staring at an
unqualified chat prompt. There's no guided next action, no way to try
Altimate without connecting a real warehouse, and no built-in demo
project to reason against.

## What we want

1. **Guided activation** — after every branch of `/onboard-connect`, the
   agent emits a short JTBD menu (see downstream / review a SQL PR / try
   the sample project / describe your own). Warehouse-connected users
   get a personalized variant with cost-analysis.
2. **Bundled sample dbt project** — a shipped jaffle-shop DuckDB project
   the user can launch without any credentials. Idempotent materialize
   to the user's home, versioned marker so future CLI upgrades can
   detect stale copies.

## Acceptance

- Scan-gate "No" leads to a menu, not an empty prompt.
- "Try the sample project" materializes a working dbt project offline.
- Sample-mode follow-up menu points at concrete jobs (`dbt build`,
  downstream impact, SQL review) that work against the sample.
- No credentials required until the user picks "Connect your own".

sahrizvi:
### Issue for this PR

Closes #1045

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

> **Draft** — depends on the base branch landing first. Codex review addressed (see checklist).

Two coupled features on top of the scan-gate work from the base branch:

1. **Activation menu** — after every `/onboard-connect` scan branch (all 4), the agent appends a JTBD-worded menu (see-downstream / review-SQL / try-sample / describe-your-own). Warehouse-connected users get a personalized variant with cost-analysis. Scan-gate "No" now dispatches `/onboard-connect skip` so the template's skip branch runs and the menu appears in the chat.
2. **Bundled jaffle-shop DuckDB sample** — under `packages/opencode/sample-projects/`, materialized to the user's home by a new `sample_setup` tool. Idempotent, non-destructive (suffix variants on collision), HOME-safety-guarded (rejects `/root`, `/tmp/*`, `/`), versioned marker for a future upgrade flow. Ships with a pre-compiled `target/manifest.json` so `/discover` and `/review` work offline out of the box.

**Files touched:**
- `packages/opencode/src/command/template/onboard-connect.txt` — activation menu appended after every scan branch, with warehouse-connected + no-data variants, and a sample-mode follow-up menu with `dbt build` + `sql_execute` routing that handles missing `dbt-duckdb` gracefully.
- `packages/opencode/src/altimate/tools/sample-setup.ts` — LLM-facing tool.
- `packages/opencode/src/altimate/onboarding/{marker,materialize,sample-source-resolver,tool-detection}.ts` — supporting logic.
- `packages/opencode/sample-projects/jaffle-shop-duckdb/` — bundled sample (2 seeds, 4 models, schema tests, pre-compiled manifest).
- `packages/opencode/script/publish.ts` — copies the sample tree into the wrapper package at publish time.
- `packages/tui/src/app.tsx` — scan-gate "No" now dispatches `/onboard-connect skip` (reverts the later close-only change on the base branch).

- **Typecheck:** `bun turbo typecheck` → 13/13 packages pass.
- **Unit tests:** 76 bun:test cases covering marker, materialize (including 19 path-traversal adversarial tests + 2 orphan-sweep tests + happy-path/collision/upgrade), sample-source-resolver, tool-detection (via real subprocess PATH-override, not copied constants), and the `sample_setup` contract — all pass.
- **End-to-end:** verified the full flow against the Altimate LLM Gateway on all three scan-gate branches (skip, scan-empty, scan-detected against a real dbt+DuckDB fixture cwd) plus the sample-materialize + follow-up-menu path.

### Screenshots / recordings

Local MP4 recordings captured for each branch (dialog navigation via a mocked auth path, plus real-gateway LLM output for skip / scan-empty / scan-detected / sample flow). Not committed to the repo — they're regeneratable from the VHS tape scripts in the reviewer's local environment.

### Checklist

---

### Codex review — addressed

- [x] **P0 security**: path traversal in `sample_setup` — regex allowlist on Zod schema + defense-in-depth check in `materializeSample` + post-`path.resolve` containment assertion + 19 adversarial tests
- [x] **P0 hygiene**: `.github/meta/merge-commit.txt` verified pre-existing on base branch (not our commit)
- [x] **P1 bug**: `sample_setup` return-state branching — 5 explicit template branches for error / version-conflict / plain-reuse / suffix-collision / fresh, with prompt-user language for the "different version" case
- [x] **P1 bug**: atomic materialize via `.tmp-<hex>` staging + `fs.renameSync` + orphan-sweep on next run; 2 new tests
- [x] **P2 test**: `tool-detection.test.ts` rewritten to invoke `detectDbtRuntime` via PATH-override with a bash-script `dbt` stub; 8 real subprocess tests
- [x] **P2 test**: `sample-source-resolver.test.ts` no longer returns early on missing source (`expect(location).toBeDefined()`)
- [x] **P2 template**: skip-branch conflict reconciled — opener + menu together in one reply, free-text vs numbered handled explicitly
- [x] **P3 fork**: `.gitignore` marker — verified as non-issue; repo convention doesn't wrap `.gitignore` in `altimate_change` markers (grep for markers in `.gitignore` returns 0 across the tree)


sahrizvi:
## Review status — consensus + codex sweep

Ready for bot review. Local branch is at `ae316980b1`.

### Fixed on this branch

**23 findings from the multi-model consensus review** (safety, correctness, telemetry, tests) — see commits `ff35d934a1` and `5f672088ee`. Highlights: LLM-visible `status: ok|error` tool prefix, safe-parent hardening, atomic staging + `Flock` serialization, install-alongside upgrade path, sample-manifest sanitizer, publish-parity guard, and template-contract tests.

**Codex sweep of the 4 rejected-claim classes** — commits `ea6ec07737` (canonicalize HOME before tmp checks, `lstatSync` in orphan sweep) and `ae316980b1` (Windows `dbt.cmd` fallback in the dbt probe).

### Deferred — on this branch's surface, chose not to fix now

Bots may still want to flag these; posting here so the intent is transparent.

- `SAFE_TARGET_NAME_RE` doesn't guard Windows reserved names (`CON`, `NUL`, `COM1`) or trailing-dot/space normalization. Preferred name comes only from the LLM tool today, which won't generate those, but a defense-in-depth allowlist is a plausible follow-up.
- Case-insensitive filesystem containment (`/Users/Foo` vs `/users/foo` on macOS/Windows). The `startsWith(parent + path.sep)` check compares case-sensitively.
- Symlinked `targetParent` bypasses the lexical containment check. Same class as the `rejectUnsafeHome` canonicalization fix in `ea6ec07737`, applied to a different guard.
- `.altimate-sample.json.tmp-<pid>` marker temp files aren't age-swept directly. They live inside the staging dir which IS age-swept, so they're cleaned up indirectly.
- `sampleName` doesn't run through an allowlist in `resolveSampleSource`. No caller feeds a user-controlled sample name today (the LLM tool hard-codes `jaffle-shop-duckdb`).

### Pre-existing — not on this branch's surface

Codex flagged these while auditing the rejection classes end-to-end. They live in `script/postinstall.mjs`, `script/publish.ts`, and `sample-projects/regenerate.sh` — files this PR does not modify. Listing them for triage as separate work, not as blockers for this PR:

- `script/postinstall.mjs` — `rm -rf ~/.altimate/builtin` before recopying skills, no ownership marker, no symlink check. Same class as the sample-target hardening in this PR, applied to a different install path.
- `script/postinstall.mjs` — the same wipe-then-copy is not staged+renamed, so a crash between the two leaves the user with no built-in skills.
- `script/postinstall.mjs` — bin/link/shim writes (`.altimate-code`, `altimate-dbt`) are not atomic.
- `script/publish.ts` — dist package dirs are not cleaned up on crash; only overwritten on next release.
- `sample-projects/regenerate.sh` — maintainer script; `rm -rf target dbt_packages` inside the sample dir has no symlink guard.
- `Flock` stale-lock detection uses wall-clock age (60s default), not PID liveness. Documented behavior of the `Flock` contract, not a bug in our usage.

### Follow-up test gap the PR doesn't cover

- No Windows CI runs the new `dbt.cmd` fallback in `tool-detection.ts`. Behavior is documented in the code comment; manual verification on a Windows box would confirm.

sahrizvi:
## Review round 2 — cubic + follow-up codex sweep

Branch is at `6c8b4452c9`.

### Just fixed (`6c8b4452c9`)

- **cubic P1-1** (`marker.ts:121`) — `classifyTarget` now gates on `sampleName` in addition to version, so a future second bundled sample can't be silently reused/upgraded through an existing first-sample dir. Threaded through `findSafeTarget` and the reverify-before-`rmSync` call. Added marker tests for wrong-sampleName, symlinked, and unreadable-dir branches.
- **cubic P1-2** (`onboard-connect.txt`) — template's "user pastes a dbt path" branch is no longer interpolated raw into bash. Refuses shell-metacharacter paths, refuses non-executable files, single-quotes the path on every subsequent use, verifies via `'<path>' --version` (not `grep -q`).
- **cubic P1-3** (`onboard-connect.txt`) — DuckDB profile path (`target/jaffle.duckdb`, relative) is now absolutized against the sample path before `warehouse_add` so `sql_execute` connects to the DB dbt just built, not an empty file at the CLI's cwd.
- **cubic P2 #4** (`materialize.ts`) — `copySampleTree` now distinguishes required vs optional entries. A missing required file throws with a reinstall pointer instead of silently marker-stamping an incomplete copy as reused-forever.
- **codex sweep NEW-10** — `verify-freshness.test.ts` now walks `models/*.sql` and `seeds/*.csv` on disk and asserts every file has a manifest node, so a maintainer adding a new model without re-running `regenerate.sh` gets caught (the per-node hash test alone couldn't catch new files with no manifest entry).

### Cubic P2/P3 items deferred on this branch

Small, chose not to bundle into this round — happy to pick these up if bots re-flag:

- P2: `sample-manifest.json` `requires` field has no consumer (comment as informational or wire into probe).
- P2: `regenerate.sh` docstring claim "every source edit is caught" — extend hash to include schema.yml, or narrow the claim.
- P2: `sample-source-resolver.test.ts` `delete process.env` isn't restored (env leak across parallel tests).
- P2: `verify-freshness.test.ts` docstring contradiction (block comment vs JSDoc on strip-one-`\n`).
- P2: `orders/schema.yml` missing `customer_name`, `order_date`, `amount` columns.
- P3: `README.md` file tree omits `.gitignore`.
- P3: `regenerate.sh` `created_at` on 2026-07-25 vs metadata pinned 2026-07-24 (timeline mismatch).
- P3: `profiles.yml` comment about dbt-duckdb path resolution — cubic disputes the "process CWD" wording; needs a doc verification pass.

### Codex sweep — pre-existing surface (out of this PR's scope)

Class-scoped re-sweep of the four cubic finding classes surfaced these siblings in code THIS PR doesn't touch. Listing for follow-up triage:

- **`/review` template** — `$ARGUMENTS` interpolated into `git show`/`git diff`/`gh pr view` without `--` or shell quoting.
- **`/configure-claude` and `/configure-codex` templates** — `which altimate` output is written into future bash snippets without quoting the binary path.
- **`build.ts` embedded-skill packaging** — asserts only `length > 0`, so a partial skill set can ship cleanly.
- **`build.ts:508–510`** — platform-package alias copy uses `.nothrow()`; a package without `bin/altimate-code` can be produced without failing the build.
- **`postinstall.mjs`** — missing bundled `altimate-dbt` and missing/failed skill copies are treated as non-fatal install success.
- **`skill/index.ts:316`** — a single filesystem SKILL.md match suppresses the embedded-skill fallback; partial postinstall → skills gone.
- **Non-DuckDB dbt profile paths** — `keyfile`, `credentials_path`, `private_key_path` also relative; template doesn't absolutize them for other warehouse types when routed through `warehouse_add`.
- **dbt manifest tools** — `impact-analysis`, `dbt-lineage`, `dbt-unit-test-gen` default to cwd-relative `target/manifest.json`; templates operating on discovered project paths should pass absolute or `cd` first.

saravmajestic (approved):

packages/opencode/src/altimate/onboarding/tool-detection.ts
```diff
@@ -0,0 +1,117 @@
...
+ * materialization (to decide which workflow entries to show in the first
+ * place). `dbt --version` output includes the plugin list on dbt 1.x, so
+ * scan for the "duckdb" plugin there rather than shelling out again.
+ *
+ * If probing fails for any reason (timeout, ENOENT, garbage output), we
+ * treat that as "tool not usable" — the caller falls back to look-only
+ * workflows. Never throws; always returns a defined result.
+ */
+
+export interface DbtRuntime {
```
mdesmet, line 46: Dont we already have a way to detect dbt with dbt integration library? why not reuse it?

mdesmet (commented):
LGTM except maybe dbt detection?

mdesmet (approved):

PR #1069, opened 2026-08-04
fix(tui): make the welcome panel responsive to terminal size (#1067)

### LINKED ISSUE #1067: home_logo slot always visible and cannot be dismissed in TUI
home_logo slot always showing on new update. takes up half the tui and is not dismissable

---

### Metadata

| Field | Value |
|-------|-------|
| CLI Version | 0.9.4 |
| Platform | darwin |
| Architecture | arm64 |
| OS Release | 25.5.0 |
| Category | bug |

saravmajestic:
Closes #1067.

### What & why

The welcome panel (`home_logo` slot) rendered a fixed two-column boot box — block `ALTIMATE CODE` wordmark on the left (~65 cols), "What is Altimate Code" on the right — that never shrank. On a typical terminal it took ~40–50% of the screen and couldn't be reduced, which is what #1067 reports.

### What this does

It now scales with terminal size, reactively (adapts live on resize), using the repo's existing breakpoint idiom (`useTerminalDimensions` + `createMemo`, as in `routes/session/permission.tsx` and `component/upgrade-indicator.tsx`). Both axes gate each step — the wordmark is wide **and** tall:

| Variant | When | What shows | ~height |
|---|---|---|---|
| **full** | `≥110w & ≥44h` | the two-column wordmark boot box (large windows only) | ~13 rows |
| **medium** | in between (the common case) | title + one condensed line + CTA, **no wordmark** | ~6 rows |
| **compact** | `<60w or <18h` | a single line; the border title already carries the version | ~3 rows |

On an everyday laptop terminal (e.g. 106×31, 80×24) you now get **medium**; the full branded panel returns only when there's genuinely room for it.

### Notes

- The breakpoint decision is a **pure function** in `welcome-panel-utils.ts` (mirroring `upgrade-indicator-utils.ts`), so it's unit-testable without the render/sync context. 4 tests cover the thresholds and that the smaller dimension wins.
- **Scope: responsiveness only.** No dismiss / collapse-once-connected behavior change — that's separable and can be a follow-up if wanted.
- Thresholds are four named constants in `welcome-panel-utils.ts`, easy to tune.

### Verified

`tsgo --noEmit` clean · 4 unit tests pass · existing `home-onboarding` render test still passes · oxlint 0/0 · prettier.

---
## Summary by cubic
Make the TUI welcome panel responsive to its available space, so it no longer overtakes small terminals or sidebar layouts. Addresses AI-1067 by switching between full, medium, and compact layouts, updating live on resize and sidebar toggle.

- **Bug Fixes**
  - Moved breakpoint logic to `welcome-panel-utils` as `welcomePanelVariant(width, height)` using available size and `PANEL_VERTICAL_RESERVE`; thresholds are floors (`MEDIUM_MIN_*`, `FULL_MIN_*` with height floor 36).
  - `WelcomePanel` now accepts `availableWidth`/`availableHeight`; `home` passes `width-4` and `height-PANEL_VERTICAL_RESERVE`, `session` passes `contentWidth()` and `height-PANEL_VERTICAL_RESERVE`.
  - Simplified render logic with `<Switch>/<Match>`, deduped the CTA, trimmed the full left column to 54 cols, and shortened the medium copy; no change to CTA/dismiss behavior.
  - Expanded unit tests to prove both width/height gates, exact boundaries, and sidebar/short-terminal cases.

<sup>Written for commit 4db83eaa7acfe7eec484a78daff8c3ed5e98f613. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1069?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * The welcome panel now adapts its layout to the available terminal size.
  * Added full, medium, and compact layouts for different terminal dimensions.
  * Compact layouts provide condensed information and streamlined readiness actions.
  * The panel now accounts for available content space across home and session views.

* **Tests**
  * Added coverage for layout selection across terminal size thresholds, boundary conditions, and constrained dimensions.

### DIFF
--- a/packages/tui/src/component/welcome-panel-utils.ts
+++ b/packages/tui/src/component/welcome-panel-utils.ts
@@ -1,29 +1,46 @@
 // Pure breakpoint logic for WelcomePanel, split out (like upgrade-indicator-utils)
 // so it's unit-testable without the component's render/context dependencies.
 //
-// The full two-column boot box (block wordmark + description) is ~65 cols wide
-// and ~8 rows tall, which eats half a small terminal (issue #1067). We scale it
-// down by terminal size. Both axes matter — the wordmark is wide AND tall — so
-// each threshold is a min across width and height. Values follow the repo's
-// breakpoint idiom (routes/session/permission.tsx uses width < 80; upgrade-
-// indicator uses width < 100).
+// The full two-column boot box (block wordmark + description) is a ~13-row
+// bordered box, so on a typical terminal it eats ~40% of the screen with no way
+// to shrink it (issue #1067). We scale it down by the space the panel ACTUALLY
+// has on both axes:
+//   - width:  the terminal minus the caller's padding and any sibling sidebar.
+//   - height: the terminal minus the fixed chrome that always shares the column
+//             with the panel (top spacer + prompt + footer), so a big panel is
+//             chosen only when it won't crowd the prompt off a short terminal.
+// `medium` (no wordmark) is the common case; `full` only when there's real room.
+//
+// Naming: these are the MINIMUMS a tier requires, matched with strict `<`
+// (`width < FULL_MIN_WIDTH` → not full). MEDIUM_MIN_* is the floor for medium
+// (below → compact); FULL_MIN_* is the floor for full (below → medium). All in
+// terms of AVAILABLE (usable) size, not the raw terminal.
 
 export type WelcomePanelVariant = "full" | "medium" | "compact"
 
-// The full panel is ~13 rows tall, so it's reserved for genuinely large windows
-// — on a typical laptop terminal (80–110 wide, 24–35 tall) it would still eat
-// ~40% of the screen, which is the problem #1067 is about. So `medium` (no
-// wordmark, ~6 rows) is the common case; `full` only when there's real room.
-/** Below these, drop to the single-line compact panel. */
-export const COMPACT_MAX_WIDTH = 60
-export const COMPACT_MAX_HEIGHT = 18
-/** At/above these, show the full two-column wordmark panel; below → medium. */
-export const MEDIUM_MAX_WIDTH = 110
-export const MEDIUM_MAX_HEIGHT = 44
+/**
+ * Rows the panel must leave for the always-present chrome below/around it (the
+ * prompt, the footer, and the home top spacer). Callers subtract this from the
+ * terminal height to get the panel's usable height. An estimate — the prompt can
+ * grow with multi-line input, but at rest this is the fixed cost.
+ */
+export const PANEL_VERTICAL_RESERVE = 8
 
-/** Choose the WelcomePanel layout for a given terminal size. */
+/** Minimum usable size for the medium panel; below either → compact (one line). */
+export const MEDIUM_MIN_WIDTH = 60
+export const MEDIUM_MIN_HEIGHT = 16
+/** Minimum usable size for the full wordmark panel; below either → medium. */
+export const FULL_MIN_WIDTH = 110
+export const FULL_MIN_HEIGHT = 36
+
+/**
+ * Choose the WelcomePanel layout from the panel's AVAILABLE size — width already
+ * minus padding/sidebar, height already minus PANEL_VERTICAL_RESERVE. Not the
+ * raw terminal (that's the #1067 bug: a sidebar-narrowed column, or a short
+ * terminal, would still pick `full`).
+ */
 export function welcomePanelVariant(width: number, height: number): WelcomePanelVariant {
-  if (width < COMPACT_MAX_WIDTH || height < COMPACT_MAX_HEIGHT) return "compact"
-  if (width < MEDIUM_MAX_WIDTH || height < MEDIUM_MAX_HEIGHT) return "medium"
+  if (width < MEDIUM_MIN_WIDTH || height < MEDIUM_MIN_HEIGHT) return "compact"
+  if (width < FULL_MIN_WIDTH || height < FULL_MIN_HEIGHT) return "medium"
   return "full"
 }
--- a/packages/tui/src/component/welcome-panel-utils.ts
+++ b/packages/tui/src/component/welcome-panel-utils.ts
@@ -1,0 +1,29 @@
+// Pure breakpoint logic for WelcomePanel, split out (like upgrade-indicator-utils)
+// so it's unit-testable without the component's render/context dependencies.
+//
+// The full two-column boot box (block wordmark + description) is ~65 cols wide
+// and ~8 rows tall, which eats half a small terminal (issue #1067). We scale it
+// down by terminal size. Both axes matter — the wordmark is wide AND tall — so
+// each threshold is a min across width and height. Values follow the repo's
+// breakpoint idiom (routes/session/permission.tsx uses width < 80; upgrade-
+// indicator uses width < 100).
+
+export type WelcomePanelVariant = "full" | "medium" | "compact"
+
+// The full panel is ~13 rows tall, so it's reserved for genuinely large windows
+// — on a typical laptop terminal (80–110 wide, 24–35 tall) it would still eat
+// ~40% of the screen, which is the problem #1067 is about. So `medium` (no
+// wordmark, ~6 rows) is the common case; `full` only when there's real room.
+/** Below these, drop to the single-line compact panel. */
+export const COMPACT_MAX_WIDTH = 60
+export const COMPACT_MAX_HEIGHT = 18
+/** At/above these, show the full two-column wordmark panel; below → medium. */
+export const MEDIUM_MAX_WIDTH = 110
+export const MEDIUM_MAX_HEIGHT = 44
+
+/** Choose the WelcomePanel layout for a given terminal size. */
+export function welcomePanelVariant(width: number, height: number): WelcomePanelVariant {
+  if (width < COMPACT_MAX_WIDTH || height < COMPACT_MAX_HEIGHT) return "compact"
+  if (width < MEDIUM_MAX_WIDTH || height < MEDIUM_MAX_HEIGHT) return "medium"
+  return "full"
+}
--- a/packages/tui/src/component/welcome-panel.tsx
+++ b/packages/tui/src/component/welcome-panel.tsx
@@ -1,4 +1,4 @@
-import { Show, createMemo } from "solid-js"
+import { Match, Show, Switch, createMemo } from "solid-js"
 import { TextAttributes } from "@opentui/core"
 import { useTerminalDimensions } from "@opentui/solid"
 import { useTheme } from "../context/theme"
@@ -7,30 +7,36 @@ import { InstallationVersion } from "@opencode-ai/core/installation/version"
 import { useReady } from "./altimate-onboarding"
 import { welcomePanelVariant } from "./welcome-panel-utils"
 
+const CONNECT_CTA = "Connect your AI model to start."
+
 // altimate_change — Claude-Code-style boot box: "What is Altimate Code" plus the
 // readiness-aware CTA. Shared between the home route and the session view so the
 // header stays consistent when a command (e.g. /discover) starts a session.
 // zIndex keeps it above transient top toasts (update/MCP) that would otherwise
 // blank its top rows.
 //
-// Responsive (issue #1067): the full two-column boot box with the block wordmark
-// is ~65 cols wide and ~8 rows tall, which eats half a small terminal. So it
-// scales down by terminal size, following the repo's breakpoint idiom
-// (createMemo on useTerminalDimensions, e.g. routes/session/permission.tsx,
-// component/upgrade-indicator.tsx). Both axes matter — the wordmark is wide AND
-// tall:
-//   full   — wordmark + full description (large terminals)
-//   medium — title + one condensed line, no wordmark
-//   compact — a single line; the border title already shows the version
-export function WelcomePanel() {
+// Responsive (issue #1067): the full two-column boot box is a ~13-row bordered
+// box that ate ~40% of the screen. It now scales down by AVAILABLE size,
+// following the repo's breakpoint idiom (createMemo over useTerminalDimensions,
+// cf. routes/session/permission.tsx:450, component/upgrade-indicator.tsx:14):
+//   full   — wordmark + full description (large windows only)
+//   medium — title + one condensed line, no wordmark (the common case)
+//   compact — a short line; the border title already carries the version
+//
+// `availableWidth` / `availableHeight` are the space the panel actually gets, not
+// the whole terminal — the caller subtracts its padding, any sibling sidebar
+// (session's contentWidth), and the fixed prompt/footer chrome
+// (PANEL_VERTICAL_RESERVE). Using the raw terminal would keep `full` selected in
+// a sidebar-narrowed column or a short window and swell the panel back up — the
+// bug #1067 is about. Both fall back to the terminal dimension when omitted.
+export function WelcomePanel(props: { availableWidth?: number; availableHeight?: number }) {
   const { theme } = useTheme()
   const ready = useReady()
   const dimensions = useTerminalDimensions()
 
-  const variant = createMemo(() => welcomePanelVariant(dimensions().width, dimensions().height))
-  const compact = createMemo(() => variant() === "compact")
-  const medium = createMemo(() => variant() === "medium")
-  const full = createMemo(() => variant() === "full")
+  const variant = createMemo(() =>
+    welcomePanelVariant(props.availableWidth ?? dimensions().width, props.availableHeight ?? dimensions().height),
+  )
 
   const title = InstallationVersion === "local" ? " Altimate Code " : ` Altimate Code v${InstallationVersion} `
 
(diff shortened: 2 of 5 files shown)


packages/tui/src/component/welcome-panel.tsx
```diff
@@ -1,78 +1,131 @@
...
+// tall:
+//   full   — wordmark + full description (large terminals)
+//   medium — title + one condensed line, no wordmark
+//   compact — a single line; the border title already shows the version
 export function WelcomePanel() {
   const { theme } = useTheme()
   const ready = useReady()
+  const dimensions = useTerminalDimensions()
+
+  const variant = createMemo(() => welcomePanelVariant(dimensions().width, dimensions().height))
```
sahrizvi: **MAJOR (Bug / Logic Error) — the breakpoint reads terminal width, but the panel never gets terminal width.**

`dimensions().width` is the whole terminal. This panel is never that wide:

- `routes/home.tsx:104` wraps it in `<box … paddingLeft={2} paddingRight={2}>` → `width - 4`.
- `routes/session/index.tsx:1181-1189` — same padding, **and** it sits in the content column of a `flexDirection="row"` layout whose sibling is a 42-column `Sidebar` (`routes/session/sidebar.tsx:30`), auto-shown whenever `width > 120` (`session/index.tsx:267-272`).

The correct quantity already exists four lines from the call site and is unused here:

```ts
// session/index.tsx:274
const contentWidth = createMemo(() => dimensions().width - (sidebarVisible() ? 42 : 0) - 4)
```

The `full` branch has a hard `width={65} flexShrink={0}` left column (line 82), so it needs ~105 usable columns:

| Terminal | Sidebar | Panel width | Variant chosen | Right-column text |
|---|---|---|---|---|
| 110×44 (home) | n/a | 106 | full | ~33 cols — fine |
| 130×50 (session) | visible (auto, >120) | 84 | **full** | **~11 cols** |
| 140×50 (session) | visible | 94 | **full** | **~21 cols** |

At 140×50 the two description paragraphs wrap to ~13 rows and the panel lands at ~22 rows — **44% of the screen**, which is the complaint in #1067. Every terminal roughly 121–151 columns wide and ≥44 rows, in a session with the sidebar visible, reproduces the original bug.

Not a regression (before this change `full` rendered at *every* size), but the fix misses its stated goal in a configuration a user on a large monitor will hit.

**Fix:**

```tsx
export function WelcomePanel(props: { availableWidth?: number } = {}) {
  const dimensions = useTerminalDimensions()
  const width = createMemo(() => props.availableWidth ?? dimensions().width - 4)
  const variant = createMemo(() => welcomePanelVariant(width(), dimensions().height))
```

```tsx
// session/index.tsx
<WelcomePanel availableWidth={contentWidth()} />
```

`contentWidth` is already reactive to `sidebarVisible()`, so toggling the sidebar re-picks the variant for free.

Belt-and-braces regardless: give the `full` left column `flexShrink={1}` or a `maxWidth`, or a `minWidth` on the right column, so a mis-sized `full` degrades instead of collapsing to a text gutter.

packages/tui/test/component/welcome-panel-utils.test.ts
```diff
@@ -0,0 +1,32 @@
...
+} from "../../src/component/welcome-panel-utils"
+
+test("large terminals get the full two-column boot box", () => {
+  expect(welcomePanelVariant(MEDIUM_MAX_WIDTH + 20, MEDIUM_MAX_HEIGHT + 10)).toBe("full")
+  expect(welcomePanelVariant(MEDIUM_MAX_WIDTH, MEDIUM_MAX_HEIGHT)).toBe("full") // at the threshold
+})
+
+test("a narrow OR short terminal drops the wordmark (medium)", () => {
+  expect(welcomePanelVariant(80, 24)).toBe("medium")
+  expect(welcomePanelVariant(MEDIUM_MAX_WIDTH - 1, 40)).toBe("medium") // narrow but tall
```
sahrizvi: **MAJOR (Testing) — these tests don't prove what the PR description claims they prove.**

The description says *"4 tests cover the thresholds and that the smaller dimension wins."* Two separate problems mean they don't.

**(a) The `full`-tier width gate is entirely unexercised.** Mutation-test the pure function — delete the width gate from the `medium` branch:

```ts
if (width < MEDIUM_MAX_WIDTH || height < MEDIUM_MAX_HEIGHT) return "medium"
// becomes
if (height < MEDIUM_MAX_HEIGHT) return "medium"
```

…and **all four tests still pass.** The other three gates (`full` height, `compact` width, `compact` height) are each caught by an equivalent mutant, so this is the one real hole.

This line is the culprit: the comment says "narrow but tall", but `40 < MEDIUM_MAX_HEIGHT (44)`, so *height* forces the `medium` result, not width. "Tall" isn't tall enough to isolate the width axis.

**(b) The compact→medium boundary is untested, and test 4 (lines 27-31) duplicates test 3.** Line 12 does an "at the threshold" check for the `full` tier; the `compact` tier has none, so a `<` → `<=` slip on either compact gate ships silently. And test 4's `(200, 17)` / `(59, 200)` take the same two branches as test 3's `(120, 17)` / `(59, 40)` — only the non-deciding dimension differs, and it stays on the same side of every threshold. No mutant is caught by test 4 that test 3 doesn't already catch.

**Fix** — repair this assertion and replace test 4 with the boundaries that are actually missing:

```ts
expect(welcomePanelVariant(MEDIUM_MAX_WIDTH - 1, MEDIUM_MAX_HEIGHT)).toBe("medium") // narrow, tall enough
expect(welcomePanelVariant(COMPACT_MAX_WIDTH, COMPACT_MAX_HEIGHT)).toBe("medium")
expect(welcomePanelVariant(COMPACT_MAX_WIDTH, 40)).toBe("medium")
expect(welcomePanelVariant(120, COMPACT_MAX_HEIGHT)).toBe("medium")
expect(welcomePanelVariant(0, 0)).toBe("compact")
```

packages/tui/src/component/welcome-panel.tsx
```diff
@@ -1,78 +1,131 @@
...
-        paddingBottom={1}
-        gap={1}
-      >
-        <box gap={0}>
-          <text fg={theme.accent} attributes={TextAttributes.BOLD}>
-            What is Altimate Code
+      {/* compact — one line; the border title already carries the version */}
+      <Show when={compact()}>
+        <box paddingLeft={2} paddingRight={2} width="100%">
+          <text fg={ready() ? theme.text : theme.primary} wrapMode="word" width="100%">
```
sahrizvi: **MINOR (Bug) — `wrapMode="word"` means the "single line" invariant doesn't actually hold.**

The compact string is 31 characters ("Connect your AI model to start."). Subtract 2 border + 4 panel padding + 4 home-route gutter and it wraps at roughly 40 columns; near 20 columns it can reach ~5 rows. `compact` is reachable at any width below 60, so the "a single line; ~3 rows" claim on line 24 is aspirational rather than enforced.

**Fix** — enforce it structurally:

```tsx
<text fg={ready() ? theme.text : theme.primary} wrapMode="none" truncate width="100%">
```

packages/tui/src/component/welcome-panel.tsx
```diff
@@ -1,78 +1,131 @@
...
+            A data-engineering harness that gives your AI real context — column-level lineage, SQL analysis, dbt, and
+            live warehouse metadata.
           </text>
-          {/* CTA only until a model is connected — stale afterwards */}
           <Show when={!ready()}>
             <text fg={theme.primary} wrapMode="word" width="100%">
               Connect your AI model to start.
             </text>
           </Show>
         </box>
```
sahrizvi: **MINOR (Design) — `medium` is ~9 rows at its own lower bound, not the advertised ~6.**

At the smallest terminal that still selects `medium` (60×18, home route → 56-wide panel, 50-column text area):

| Element | Rows |
|---|---|
| border | 2 |
| `paddingTop` / `paddingBottom` | 2 |
| title | 1 |
| description (~130 chars at 50 cols) | 3 |
| CTA (disconnected) | 1 |
| **total** | **~9** |

That's **50% of an 18-row terminal** — the same ratio #1067 complains about, in a different layout. Even at minimum intrinsic size the disconnected `medium` panel needs 7 rows, so the "~6 rows" figure in `welcome-panel-utils.ts:16` isn't reachable while the CTA is present.

Raise `COMPACT_MAX_HEIGHT`, drop the vertical padding here, or shorten the description so it fits one or two lines at 50 columns.

packages/tui/src/component/welcome-panel-utils.ts
```diff
@@ -0,0 +1,29 @@
...
+// The full panel is ~13 rows tall, so it's reserved for genuinely large windows
+// — on a typical laptop terminal (80–110 wide, 24–35 tall) it would still eat
+// ~40% of the screen, which is the problem #1067 is about. So `medium` (no
+// wordmark, ~6 rows) is the common case; `full` only when there's real room.
+/** Below these, drop to the single-line compact panel. */
+export const COMPACT_MAX_WIDTH = 60
+export const COMPACT_MAX_HEIGHT = 18
+/** At/above these, show the full two-column wordmark panel; below → medium. */
+export const MEDIUM_MAX_WIDTH = 110
+export const MEDIUM_MAX_HEIGHT = 44
```
sahrizvi: **MINOR (Code Quality / Documentation) — the `*_MAX_*` names are actually minimums for the tier above.**

Every comparison in `welcomePanelVariant` is a strict `<`, so:

- `COMPACT_MAX_WIDTH = 60` → a 60-wide terminal is **medium**, not compact. 60 is the *minimum* width for `medium`.
- `MEDIUM_MAX_WIDTH = 110` → 110 is the *minimum* width for `full`.

The doc comment on line 20 ("At/above these, show the full…") already describes minimum semantics and contradicts the name on line 21. These constants are exported and consumed by the tests, so the wrong name propagates — and it's the direct cause of several "is this an off-by-one?" readings that turn out to be wrong.

Suggest `MEDIUM_MIN_WIDTH` / `MEDIUM_MIN_HEIGHT` / `FULL_MIN_WIDTH` / `FULL_MIN_HEIGHT`.

sahrizvi (commented):
## Review — responsiveness change works, but two things block

The direction is right and the decomposition is clean. Splitting the breakpoint decision into a pure, exported function mirroring `upgrade-indicator-utils.ts` is the right call, the `createMemo` + `useTerminalDimensions` wiring matches the idiom in `upgrade-indicator.tsx:14` and `permission.tsx:450` (genuinely reactive, no destructured props), and the `full` branch is a faithful move of the original markup so the large-terminal path is low-risk. Theme tokens are unchanged, there's no dead code, and the scope discipline — explicitly declining to bundle dismiss/collapse — is good.

Two blockers, both left inline:

1. **The breakpoint is computed from terminal width, but the panel never gets terminal width.** In the session view with the sidebar open, `full` is still selected when the panel only has ~84–94 columns, and the panel swells back to ~44% of the screen — the exact bug #1067 reports. (`welcome-panel.tsx:30`)
2. **The tests don't prove the width threshold the change is built around.** Deleting `width < MEDIUM_MAX_WIDTH` from the source leaves all four tests green. (`welcome-panel-utils.test.ts:17`)

Plus three minors inline: the compact variant's `wrapMode="word"` breaks its own single-line invariant, `medium` is ~9 rows at its lower bound rather than ~6, and the `*_MAX_*` constants are really minimums for the tier above.

---

### MINOR — no render test for the component, though the repo has the harness

The pure function is tested; the part that can actually break isn't — that `useTerminalDimensions` is wired up, that the memo survives a resize, that each variant fits its row budget. `test/feature-plugins/home-onboarding.test.tsx` shows `testRender` takes an explicit `{ width, height }`, so a three-case render test is cheap.

"existing `home-onboarding` render test still passes" is true but says nothing about this component — that test renders at 120×4 and never mounts `WelcomePanel`. A render test at session-like widths with the sidebar visible is also the only thing that would have caught blocker #1.

### NIT — contradictory height figures inside one file

`welcome-panel-utils.ts:4-5` says the full box is "~8 rows tall"; lines 13-16 say "~13 rows tall". Both are defensible (inner content vs. bordered box) but the reader can't tell which is which. Pick one framing.

### NIT — structural cleanups worth folding in

- `welcome-panel.tsx:31-33` — `compact()`, `medium()`, `full()` are three memos over one memo feeding three mutually-exclusive `<Show>` blocks (lines 51, 60, 78). `routes/session/permission.tsx` and `session/index.tsx:1349-1358` both use `<Switch><Match when={…}>` for exactly this; switching makes the exclusivity structural rather than incidental.
- The fixed `width={65}` left column (line 82) is ~15 columns wider than its content — `src/logo.ts` is 32 + 1 gap + 17 = 50 columns, and the title line is 24. Trimming to ~52 would let `full` engage on a smaller terminal and would soften blocker #1.
- "Connect your AI model to start." is duplicated verbatim at lines 54, 71 and 122 — worth one module constant. (The three *descriptions* are deliberately different copy, not duplication.)

### Open design question, not a defect

`MEDIUM_MAX_HEIGHT = 44` means the wordmark is effectively never seen on a laptop — it needs an external monitor or a full-screen tall window. That may well be intentional ("full only when there's real room"), but it's worth an explicit product decision, since it makes the branded panel dead code for most users. Note the tension with blocker #1: lowering the threshold widens the range where the sidebar case misfires, so fix that first.

### Missing tests, collected

- Compact→medium boundary at exactly `(60, 18)`, `(60, 40)`, `(120, 18)` — the entire lower boundary.
- `(109, 44)` — the only case that isolates the `full` width gate.
- Degenerate sizes `(0, 0)`, `(1, 1)` — currently return `compact`, which is correct, but nothing pins it.
- Render tests per variant, with height-budget assertions and a resize transition.
- Session layout with the sidebar visible — where blocker #1 lives.

saravmajestic:
Thanks — both blockers were spot on. Fixed in `4db83eaa7a`.

### Blocking 1 — breakpoint read the whole terminal, not the panel's width
Right: the home slot pads the panel and the session view sits it beside a 42-col sidebar, so `full` was picked at ~84 usable cols. `WelcomePanel` now takes `availableWidth`; **home** passes `dimensions().width - 4`, **session** passes the existing `contentWidth()` (already `width - sidebar - 4`). It's reactive to resize **and** sidebar toggle (contentWidth depends on `sidebarVisible`). Concretely: 130w session with the sidebar auto-shown now resolves to `medium` (84 usable cols), where it used to be `full`.

I also gave **height** the same treatment while here — it had the analogous problem (the panel shares the column with the prompt/footer). Callers now pass `availableHeight = dimensions().height - PANEL_VERTICAL_RESERVE` (8 rows of chrome), so a short window drops to a smaller variant instead of a too-tall panel. Thresholds re-expressed in usable terms (`MEDIUM_MIN_HEIGHT 16`, `FULL_MIN_HEIGHT 36`) so `full` still engages at a ~44-row terminal and ≤~23 rows → `compact`.

### Blocking 2 — tests didn't prove the width gate
Fixed. Each gate is now isolated on one axis so deleting its `<` check fails a test — e.g. `(FULL_MIN_WIDTH-1, tall) -> medium` (dies if `width < FULL_MIN_WIDTH` is removed), `(MEDIUM_MIN_WIDTH-1, tall) -> compact`, and the mirror height cases. Added exact boundaries, the sidebar case, a short-terminal height case, and degenerate `(0,0)/(1,1)`.

### Minors / nits
- **Naming** (you + cubic): `*_MAX_*` → `MEDIUM_MIN_*` / `FULL_MIN_*` — they're the floor a tier needs, matched with strict `<`.
- **`<Switch>/<Match>`** instead of three memos + three `<Show>` — exclusivity is now structural.
- **Left column 65 → 54** (logo is ~50 cols) so `full` engages sooner.
- **Medium copy** shortened to one line (was ~9 rows at its lower bound).
- **CTA** deduped into `CONNECT_CTA`.
- **Height figures** unified to "~13-row bordered box"; dropped the strict single-line claim on `compact` (it wraps only on an extremely narrow terminal — better than truncating the CTA mid-word).

### Open question — `FULL_MIN_HEIGHT`
Kept high on purpose: the wordmark shows only on large windows (the product ask for #1067 was "don't dominate the screen"). Happy to lower it if you'd rather see the branded panel more often — note it trades against the short-terminal case.

### Render tests
Added the full boundary/gate/degenerate coverage on the pure function. I did **not** add a session-with-sidebar integration render test: a `WelcomePanel`-only render (as you noted) wouldn't catch blocker 1 — that needs the session view mounted with the sidebar, which is a heavier integration test. Wiring is verified by inspection + `tsgo`. Glad to add that integration test as a follow-up if you want the regression guard.

`tsgo` clean · 9 unit + 11 render/adjacent tests pass · oxlint 0/0.

sahrizvi:
## Round 2 review — `4db83eaa7a`

Both blockers are genuinely fixed, and the test work went past what I asked for. Approving. What's left is one MINOR worth a quick fix and a few nits.

### Blocker 1 — width source: fixed

`props.availableWidth ?? dimensions().width` read inside the memo (not destructured), so reactivity survives the prop boundary. Both call sites check out: home's `- 4` matches its own `paddingLeft/Right={2}`, and session reusing `contentWidth()` means it's reactive to the sidebar toggle as well as resize. The reported case resolves — 130-col session with the sidebar open now yields 84 available → `medium`, and it's pinned by a test.

The `65 → 54` trim is right: `logo.ts` measures 32 + 1 gap + 17 = **exactly 50 columns**, so 54 with `paddingLeft={1}` fits with 3 to spare. At `FULL_MIN_WIDTH` the right column gets 110 − 2 − 54 − 1 − 4 = **49 columns** of text, so `full` fits at its own floor.

### Blocker 2 — tests: fixed, beyond the ask

I re-ran the mutation exercise with eight mutants — deleting each of the four gates, and flipping each `<` to `<=`:

| Mutant | Result |
|---|---|
| drop `width < FULL_MIN_WIDTH` | killed |
| drop `height < FULL_MIN_HEIGHT` | killed |
| drop `width < MEDIUM_MIN_WIDTH` | killed |
| drop `height < MEDIUM_MIN_HEIGHT` | killed |
| `width <= FULL_MIN_WIDTH` | killed |
| `height <= FULL_MIN_HEIGHT` | killed |
| `width <= MEDIUM_MIN_WIDTH` | killed |
| `height <= MEDIUM_MIN_HEIGHT` | killed |

All eight die — every gate proven in both directions. I only asked for the one isolated width case; deriving `TALL`/`WIDE` from the floor constants makes the isolation structural rather than something a future threshold retune can silently break. Nice.

### Prior findings

| # | Finding | Status |
|---|---|---|
| 1 | Breakpoint used raw terminal width | **Fixed** |
| 2 | Tests didn't prove the width gate | **Fixed** |
| 3 | `compact` `wrapMode="word"` vs the single-line claim | **Resolved** — agreed, softening the claim beats truncating a CTA mid-word |
| 4 | `medium` ~9 rows at its lower bound | **Fixed in substance** — 8 rows of a raw 24-row terminal (33%, was 50%) |
| 5 | `*_MAX_*` names were minimums | **Fixed** — no stale references anywhere |
| 6 | No component render test | **Deferred** — your reasoning is right; see the MINOR below for why I'd still want it eventually |
| 7 | NITs (height figures, `<Switch>`, left column, CTA dedupe) | **Fixed** |

---

### MINOR — `PANEL_VERTICAL_RESERVE = 8` fits session, under-reserves home

The docstring calls it "the prompt, the footer, and the home top spacer", but that's home's layout and the number fits session's. Measuring the pieces:

- **Home:** 2 (top spacer, `home.tsx:110`) + 1 (prompt wrapper `paddingTop`) + ~4-5 (prompt: input + separator row + status row) + 3 (footer — `feature-plugins/home/footer.tsx` is `paddingTop` 1 + content 1 + `paddingBottom` 1) ≈ **10-11 rows**. Reserve 8 under-reserves by ~2-3.
- **Session:** gap 1 + gap 1 + `paddingBottom` 1 + ~4-5 (prompt) ≈ **7-8 rows** — no top spacer, no footer in that column. Reserve 8 is about right.

Practical impact is small: at `full`'s 44-row floor the panel is ~14 rows, leaving ~20 rows of slack, so nothing overflows. But the constant is a heuristic that matches neither route as documented. Either split it into `HOME_VERTICAL_RESERVE` / `SESSION_VERTICAL_RESERVE`, or reword the docstring so it doesn't present one route's chrome as both.

### MINOR — the optional props default to raw terminal dimensions

`props.availableWidth ?? dimensions().width` means a future call site that forgets a prop silently gets exactly the pre-fix behavior — and nothing tests the call sites, so it'd be silent. Given the whole point of this commit is that raw terminal dimensions are the wrong input, consider making both props required. That also converts finding 6 from "we should render-test this" into "the type system covers the regression".

### NIT — `medium`'s "one condensed line" is two lines at `medium`'s own minimum width

The string is ~96 chars; at the minimum `medium` width the text area is 60 − 2 border − 4 padding = 54 columns, so it wraps to 2. It's one line above ~100 available columns. Wording only — the 8-row budget already assumes 2.

### NIT — the width retune wasn't disclosed alongside the height retune

The commit message calls out the height change (raw ≤23 → compact, was ≤18). The width floor moved too: measured against available width, home's raw compact threshold went 60 → 64 columns. Same class of intentional change, worth the same one-line note.

### NIT — the classic 80×24 terminal sits exactly on the medium floor

80×24 → available (76, 16), and `MEDIUM_MIN_HEIGHT` is 16 — dead on the boundary, with 80×23 flipping to a single line. The test at line 49 pins it, so it's clearly considered rather than accidental. Just worth knowing the most common legacy size is on a cliff edge.

---

### Checked and fine — no action needed

- **Reactivity through the prop boundary.** Props are getters, so reading `props.x` inside `createMemo` tracks; only destructuring would break it. Sidebar toggle → `contentWidth` recomputes → `variant` recomputes. Verified.
- **`<Switch>/<Match>` conversion** kept the `Show when={!ready()}` CTA gating in both `medium` and `full`.
- **Zero/negative available dimensions** return `compact` correctly, and the degenerate tests pin it.
- **`contentWidth` subtracts 42 even when the sidebar is an absolute overlay** at `!wide()` — pre-existing, shared with the consumer at line 1167, and it errs conservatively, so nothing to do.

### On `FULL_MIN_HEIGHT`

Your call, and the reasoning holds — #1067 was "don't dominate the screen", so erring toward the wordmark being rare is the right default. Leaving it.

sahrizvi (approved):

PR #1071, opened 2026-08-05
fix(tui): per-route vertical reserve + required panel props (#1069 round-2 follow-up)

saravmajestic:
Follow-up to the merged #1069, addressing @sahrizvi's round-2 review. Also passed a 2-model consensus panel (Claude + Gemini) — both **SOUND**.

### What changed (from the round-2 comments)

**MINOR — reserve fit only session, under-reserved home.** Split into per-route constants that match each route's actual chrome:
- `HOME_VERTICAL_RESERVE = 10` (top spacer 2 + prompt wrapper 1 + prompt ~4 + footer ~3)
- `SESSION_VERTICAL_RESERVE = 7` (gaps 2 + paddingBottom 1 + prompt ~4; no top spacer/footer in that column)

Thresholds re-expressed in usable terms (`MEDIUM_MIN_HEIGHT 13`, `FULL_MIN_HEIGHT 34`) so `full` still engages at a ~44-row terminal on home, and the classic **80×24 clears the medium floor with margin** (the cliff the review flagged now sits at 23, off the common size).

**MINOR — optional props silently defaulted to raw terminal.** Made `availableWidth`/`availableHeight` **required** and dropped the component's own `useTerminalDimensions` + fallback. A call site that forgets to scale is now a **compile error**, not a silent regression — this is the type-level guard for the "no render test" finding.

**NITs**
- Medium copy wording clarified (one line on a wide terminal, two at the narrow end).
- Height now uses the same available-space model as width, so short windows shrink instead of crowding the prompt.
- Per-route note: `full` engages ~3 rows sooner on session (≥41) than home (≥44) — intentional (session has less chrome, so more usable height).

### From the consensus panel (both models endorsed)
- **Centralized `PANEL_HORIZONTAL_PADDING = 4`** — was a bare `-4` in the home call site + test helper (vertical reserves were already named constants).
- **Added the sidebar-toggle transition test**: `session(150,50,false) → full` vs `session(150,50,true) → medium` — pins the exact #1067 case directly.

### Effective terminal → variant (verified)
| Terminal | Route | Variant |
|---|---|---|
| 120×44 | home | full (the chosen "full at 44") |
| 106×31 / 80×24 | home | medium |
| 120×22 | home | compact (short-terminal height fix) |
| 130×50 + sidebar | session | medium (blocker-1) |
| 130×50 no sidebar | session | full |

### Verification
`tsgo` clean · **12 unit + 11 render/adjacent tests** pass · oxlint 0/0 · prettier. Gate-isolation tests still kill mutants (deleting or `<`→`<=`-flipping any gate fails a test).

Deferred (noted by both reviewers): a session-with-sidebar **render** test as the true call-site regression guard — happy to add as a follow-up.

---
## Summary by cubic
Fixes welcome panel sizing so variants are chosen by available space per route, not raw terminal size. Updates reserves, thresholds, and shared route math to prevent “full” in the session sidebar and on short terminals.

- **Bug Fixes**
  - Per-route reserves corrected: `HOME_VERTICAL_RESERVE = 15`, `SESSION_VERTICAL_RESERVE = 9`.
  - Thresholds retuned on available size: `MEDIUM_MIN_HEIGHT = 8`, `FULL_MIN_HEIGHT = 30`. Results: 80×24 → medium; home full at ≥45 rows; session full at ≥39; 120×22 → compact.
  - Centralized arithmetic: `homeAvailable` / `sessionAvailable`, shared `PANEL_HORIZONTAL_PADDING = 4` and `SIDEBAR_WIDTH = 42`; session `contentWidth` derives from `panelAvailable().width`.
  - Session sizing follows the content column and narrows with the sidebar (including the overlay state); on 150×50, toggling the sidebar flips full → medium.
  - Added tests for reserves, overlay narrowing, boundaries and degenerate sizes, plus a render test that asserts each variant’s content.

- **Migration**
  - `WelcomePanel` now requires `availableWidth` and `availableHeight`. Pass values from `homeAvailable(...)` / `sessionAvailable(...)`.

<sup>Written for commit a5e1b50e9fbc3348880248a2e1272604c92e3b0a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1071?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Improvements**
  * Improved welcome panel sizing across Home and Session views.
  * Responsive layouts now account for each view’s available space, including surrounding interface elements and sidebars.
  * Updated height thresholds provide more accurate compact, medium, and full panel layouts.
  * Panel dimensions now adjust more reliably as available terminal space changes.
  * Improved panel content and connection prompts across different layout sizes.

### DIFF
--- a/packages/tui/src/component/welcome-panel-utils.ts
+++ b/packages/tui/src/component/welcome-panel-utils.ts
@@ -5,43 +5,58 @@
 // bordered box, so on a typical terminal it eats ~40% of the screen with no way
 // to shrink it (issue #1067). We scale it down by the space the panel ACTUALLY
 // has on both axes:
-//   - width:  the terminal minus the caller's padding and any sibling sidebar.
+//   - width:  the terminal minus the caller's padding and any sibling sidebar
+//             that consumes layout width.
 //   - height: the terminal minus the fixed chrome that always shares the column
-//             with the panel (see the per-route reserves below), so a big panel
-//             is chosen only when it won't crowd the prompt off a short terminal.
+//             with the panel (the per-route reserves below), so a big panel is
+//             chosen only when it won't crowd the prompt off a short terminal.
 // `medium` (no wordmark) is the common case; `full` only when there's real room.
 //
-// Naming: these are the MINIMUMS a tier requires, matched with strict `<`
-// (`width < FULL_MIN_WIDTH` -> not full). MEDIUM_MIN_* is the floor for medium
-// (below -> compact); FULL_MIN_* is the floor for full (below -> medium). All in
-// terms of AVAILABLE (usable) size, not the raw terminal.
+// Route arithmetic lives here (homeAvailable / sessionAvailable) so the routes
+// and the tests share ONE definition — a test that maps a terminal size to a
+// variant then exercises the real call-site math, not a copy of it.
 
 export type WelcomePanelVariant = "full" | "medium" | "compact"
 
-// Rows the panel must leave for the always-present chrome that shares its column.
-// The two routes have different chrome, so they reserve different amounts; the
-// shared thresholds below then transition ~3 rows of terminal height apart
-// between the routes (session, with less chrome, shows a bigger variant sooner).
-// Estimates — the prompt can grow with multi-line input; this is the at-rest cost.
+// --- Chrome reserves (rows the panel must leave for always-present siblings) ---
 //
-// home (routes/home.tsx): top spacer (2) + prompt wrapper paddingTop (1) + prompt
-//   (~4) + footer (~3, feature-plugins/home/footer.tsx) ≈ 10.
-export const HOME_VERTICAL_RESERVE = 10
-// session (routes/session/index.tsx): two column gaps (2) + paddingBottom (1) +
-//   prompt (~4); no top spacer, no footer in that column ≈ 7.
-export const SESSION_VERTICAL_RESERVE = 7
+// Prompt tree at rest (component/prompt/index.tsx): top rule 1 + inner paddingTop
+// 1 + textarea 1 + separator 1 + agent/model meta row 1 + idle hint row 1 = 6.
+// Measured against the rendered tree, not estimated — the earlier "~4" undercounted
+// it (the meta and hint rows are always in flow).
+const PROMPT_REST_HEIGHT = 6
 
-// Columns the home slot spends on its own left/right padding (2 + 2); the caller
-// subtracts this to get the panel's usable width. (Session's contentWidth applies
-// the same 4 as part of its own content-column math.)
+// home (routes/home.tsx): top spacer 2 (`<box height={2}>`) + prompt wrapper
+// paddingTop 1 + prompt 6 + home_bottom slot 3 (feature-plugins/home/tips.tsx
+// paddingTop, rendered unconditionally — flexbox shrinks content, not padding) +
+// footer 3 (feature-plugins/home/footer.tsx) = 15.
+export const HOME_VERTICAL_RESERVE = 2 + 1 + PROMPT_REST_HEIGHT + 3 + 3
+// session (routes/session/index.tsx): two column gaps 2 + paddingBottom 1 +
+// prompt 6 = 9. No top spacer and no footer share this column.
+export const SESSION_VERTICAL_RESERVE = 2 + 1 + PROMPT_REST_HEIGHT
+
+// Columns the home slot spends on its own left/right padding (2 + 2). Session
+// subtracts the same via sessionAvailable(); both routes import this constant so
+// the value has a single source of truth.
 export const PANEL_HORIZONTAL_PADDING = 4
+// Width the session sidebar occupies WHEN it consumes layout width (i.e. it is
+// rendered in-flow, not as an overlay). Shared with the route + tests so they
+// can't drift — the drift that produced #1067.
+export const SIDEBAR_WIDTH = 42
 
-/** Minimum usable size for the medium panel; below either -> compact (one line). */
+// --- Thresholds ---
+// MEDIUM_MIN_* is a real FIT requirement: below it the medium panel (~8 rows /
+// ~a title + one/two-line description) would not fit, so drop to the one-line
+// compact. FULL_MIN_* is a product BREAKPOINT, not a fit minimum: the full panel
+// is only ~13 rows, but we require far more available space so the branded
+// wordmark appears only on a genuinely large terminal and never dominates a
+// small one (the #1067 ask). All in AVAILABLE (usable) terms, not the raw
+// terminal, matched with strict `<`.
 export const MEDIUM_MIN_WIDTH = 60
-export const MEDIUM_MIN_HEIGHT = 13
-/** Minimum usable size for the full wordmark panel; below either -> medium. */
+export const MEDIUM_MIN_HEIGHT = 8
 export const FULL_MIN_WIDTH = 110
-export const FULL_MIN_HEIGHT = 34
+// ~45-row home terminal / ~39-row session terminal after the reserves above.
+export const FULL_MIN_HEIGHT = 30
 
 /**
  * Choose the WelcomePanel layout from the panel's AVAILABLE size — width already
@@ -53,4 +68,33 @@ export function welcomePanelVariant(width: number, height: number): WelcomePanel
   if (width < MEDIUM_MIN_WIDTH || height < MEDIUM_MIN_HEIGHT) return "compact"
   if (width < FULL_MIN_WIDTH || height < FULL_MIN_HEIGHT) return "medium"
   return "full"
+}
+
+/** Available panel size on the home route for a given terminal size. */
+export function homeAvailable(terminalWidth: number, terminalHeight: number): { width: number; height: number } {
+  return {
+    width: terminalWidth - PANEL_HORIZONTAL_PADDING,
+    height: terminalHeight - HOME_VERTICAL_RESERVE,
+  }
+}
+
+/**
+ * Available panel size on the session route. Subtracts SIDEBAR_WIDTH whenever the
+ * sidebar is open (`sidebarVisible`) — the panel shares the same content-column
+ * basis as the messages (session's `contentWidth`), so the two stay aligned.
+ *
+ * On narrow terminals the sidebar renders as a dimmed full-area overlay rather
+ * than in-flow; the content column still narrows uniformly (panel + messages)
+ * and both restore to full width when it closes, so we deliberately size to the
+ * narrowed column rather than the transient obscured width.
+ */
+export function sessionAvailable(
+  terminalWidth: number,
+  terminalHeight: number,
+  sidebarVisible: boolean,
+): { width: number; height: number } {
+  return {
+    width: terminalWidth - (sidebarVisible ? SIDEBAR_WIDTH : 0) - PANEL_HORIZONTAL_PADDING,
+    height: terminalHeight - SESSION_VERTICAL_RESERVE,
+  }
 }
--- a/packages/tui/src/component/welcome-panel-utils.ts
+++ b/packages/tui/src/component/welcome-panel-utils.ts
@@ -7,36 +7,46 @@
 // has on both axes:
 //   - width:  the terminal minus the caller's padding and any sibling sidebar.
 //   - height: the terminal minus the fixed chrome that always shares the column
-//             with the panel (top spacer + prompt + footer), so a big panel is
-//             chosen only when it won't crowd the prompt off a short terminal.
+//             with the panel (see the per-route reserves below), so a big panel
+//             is chosen only when it won't crowd the prompt off a short terminal.
 // `medium` (no wordmark) is the common case; `full` only when there's real room.
 //
 // Naming: these are the MINIMUMS a tier requires, matched with strict `<`
-// (`width < FULL_MIN_WIDTH` → not full). MEDIUM_MIN_* is the floor for medium
-// (below → compact); FULL_MIN_* is the floor for full (below → medium). All in
+// (`width < FULL_MIN_WIDTH` -> not full). MEDIUM_MIN_* is the floor for medium
+// (below -> compact); FULL_MIN_* is the floor for full (below -> medium). All in
 // terms of AVAILABLE (usable) size, not the raw terminal.
 
 export type WelcomePanelVariant = "full" | "medium" | "compact"
 
-/**
- * Rows the panel must leave for the always-present chrome below/around it (the
- * prompt, the footer, and the home top spacer). Callers subtract this from the
- * terminal height to get the panel's usable height. An estimate — the prompt can
- * grow with multi-line input, but at rest this is the fixed cost.
- */
-export const PANEL_VERTICAL_RESERVE = 8
+// Rows the panel must leave for the always-present chrome that shares its column.
+// The two routes have different chrome, so they reserve different amounts; the
+// shared thresholds below then transition ~3 rows of terminal height apart
+// between the routes (session, with less chrome, shows a bigger variant sooner).
+// Estimates — the prompt can grow with multi-line input; this is the at-rest cost.
+//
+// home (routes/home.tsx): top spacer (2) + prompt wrapper paddingTop (1) + prompt
+//   (~4) + footer (~3, feature-plugins/home/footer.tsx) ≈ 10.
+export const HOME_VERTICAL_RESERVE = 10
+// session (routes/session/index.tsx): two column gaps (2) + paddingBottom (1) +
+//   prompt (~4); no top spacer, no footer in that column ≈ 7.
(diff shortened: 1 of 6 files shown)


sahrizvi:
@saravmajestic CI is failing on this PR.

sahrizvi (commented):
## Review — round 3

The refactor is the right shape and structurally answers the round-2 comments: per-route reserves are the correct fix, and making the props required is the right way to close the silent-fallback path. But the central quantitative claim — the reserve arithmetic, and the "80×24 clears the medium floor with margin" conclusion drawn from it — doesn't survive checking against the rendered layout.

### Major — both vertical reserves undercount the real chrome

`welcome-panel-utils.ts:27-32` budgets `prompt (~4)`. Counting the actual `Prompt` tree at rest (`component/prompt/index.tsx:1479-1611`):

1. `<box border={["top"]}>` — 1 row (top rule, L1479)
2. inner `paddingTop={1}` — 1 row (L1483)
3. `<textarea minHeight={1}>` — 1 row (L1500)
4. `<box height={1} border={["top"]} />` — 1 row (separator, L1569)
5. agent/model meta row — 1 row, always rendered (`fallback={<box height={1} />}`, L1572)
6. status/hint row — **0 or 1**. The wrapping `<box>` at L1611 is unconditional, but at idle the `<Switch>` falls through to `<Match when={true}>{props.hint ?? <text />}</Match>` (L1752), and neither route passes `hint` — so whether that empty row occupies a line is renderer-dependent.

**The prompt is 5-6 rows at rest, not ~4** — 5 are certain, the 6th needs measuring.

Home is undercounted a second time: the reserve omits the `home_bottom` slot entirely (`routes/home.tsx:128`). Its built-in provider always returns a box with `paddingTop={3}` (`feature-plugins/home/tips.tsx:34`, outside the `<Show when={props.show}>`), and flexbox shrinks content, not padding — so those 3 rows are unconditional whether or not tips render.

```
home:    top spacer 2 + wrapper 1 + prompt 5-6 + home_bottom 3+ + footer 3 = 14-15+   (constant says 10)
session: gaps 2 + paddingBottom 1 + prompt 5-6                            = 8-9      (constant says 7)
```

**Why this matters rather than being a comment nit.** The round-2 comment this PR answers was *"80×24 sits exactly on the medium floor"*, and the fix claims the cliff moved to 23. Against `MEDIUM_MIN_HEIGHT = 13`:

| home reserve | source | `home(80,24)` available | variant |
|---|---|---|---|
| 10 | as merged | 14 | medium, margin 1 |
| 11 | 5-row prompt, ignoring `home_bottom` (most conservative) | 13 | medium, **margin 0 — exactly on the floor** |
| 12 | 6-row prompt, ignoring `home_bottom` | 12 | **compact** |
| 14-15 | prompt + `home_bottom` | 9-10 | **compact** |

Every corrected reading destroys the claimed margin. The most conservative one puts 80×24 back *exactly on the medium floor* — verbatim the round-2 complaint — and the rest drop it to compact. The margin exists only at reserve 10, and `expect(home(80,24)).toBe("medium")` is pinned to that constant rather than to the rendered layout.

To be precise about the consequence: at 80×24 the unconnected medium panel plus a one-line tip fits almost exactly, so this isn't guaranteed overflow — it clips once a tip wraps to a second line. The defect is the absent margin, not certain breakage.

**Suggested fix** — derive the reserves from the real tree and re-tune the thresholds separately, so the two stop compensating for each other:

```ts
// Measure once against a real render and fix the value; 5 is the floor,
// 6 if the idle hint row occupies a line.
const PROMPT_REST_HEIGHT = 6

export const HOME_VERTICAL_RESERVE =
  2 +                    // top spacer (home.tsx:109)
  1 +                    // prompt wrapper paddingTop (home.tsx:123)
  PROMPT_REST_HEIGHT +
  3 +                    // home_bottom paddingTop (tips.tsx:34)
  3                      // footer (feature-plugins/home/footer.tsx:64-82)
// 15

export const SESSION_VERTICAL_RESERVE = 2 + 1 + PROMPT_REST_HEIGHT  // 9
```

Then choose `MEDIUM_MIN_HEIGHT` / `FULL_MIN_HEIGHT` against the panel's actual heights (medium ≈ 6-8 rows, full ≈ 13) plus whatever product breakpoints you want. If 80×24 must stay `medium`, that becomes an explicit threshold decision instead of a hidden one.

### Minor

**1. `PANEL_HORIZONTAL_PADDING` isn't actually centralized** — `welcome-panel-utils.ts:34-37`, `routes/session/index.tsx:275`

The constant's comment says session "applies the same 4 as part of its own content-column math", but `contentWidth` is still `dimensions().width - (sidebarVisible() ? 42 : 0) - 4` — literal `4`, literal `42`, neither imported. Only `home.tsx` uses the constant. Either import it at `session/index.tsx:275` (and add a `SIDEBAR_WIDTH` alongside), or drop the claim from the comment.

**2. Sidebar width `42` is duplicated between production and the test helper** — `test/component/welcome-panel-utils.test.ts:22`, `routes/session/index.tsx:275`

If the sidebar width changes, production and tests drift silently in the same direction that produced #1067.

**3. Test helpers re-implement the call-site arithmetic instead of exercising it** — `test/component/welcome-panel-utils.test.ts:19-22`

`home()` and `session()` duplicate exactly what `home.tsx:116-117` and `session/index.tsx:1194-1195` compute, so the "verified terminal → variant" table proves a property of the helpers, not of the routes — which is precisely why the Major finding above passed CI green. Required props catch a *missing* prop, never a *wrong* one, so they don't cover this. Cheap fix short of the deferred render test: export `homeAvailable(w,h)` / `sessionAvailable(w,h,sidebar)` from the utils module and call them from both the routes and the tests, so there's one definition of the arithmetic.

**4. Worth confirming intent: session sizes against the sidebar even when it's an overlay** — `routes/session/index.tsx:267, 275, 1194`

`contentWidth` subtracts 42 whenever `sidebarVisible()`, but the sidebar only consumes column width when `wide()` (`width > 120`). Below that it renders `position="absolute"` (`session/index.tsx:1360-1371`) and takes no layout width, so on a 100-col terminal with the sidebar opened `contentWidth()` reports 54 while the panel has ~96 columns of layout space, and the panel drops to `compact`.

Raising this as a question, not a defect — the overlay does visually cover roughly those columns behind a dimmed backdrop, so sizing to the obscured width may well be deliberate, and it can't produce the opposite (wrongly-`full`) error since `full` needs ≥156 raw cols where `wide()` is always true. Worth a line of comment recording which behaviour is intended, now that panel sizing routes through `contentWidth`.

### Nit

**5. The height thresholds read as fit requirements but aren't** — `welcome-panel-utils.ts:39-44`

`MEDIUM_MIN_HEIGHT = 13` / `FULL_MIN_HEIGHT = 34` are documented as "minimum usable size for the … panel", but the medium panel is ~6-8 rows and the full panel ~13 (`welcome-panel.tsx:88-138`). They're breathing-room heuristics several times the actual need, and nothing records why 13 and 34 — relevant once the Major finding forces a re-tune.

### What's done well

- Splitting `PANEL_VERTICAL_RESERVE` per route is the right response to the round-2 finding, and the comments walk the JSX rather than asserting a number. The home footer (3 rows: paddingTop 1 + content 1 + paddingBottom 1) and the `<box height={2}>` spacer are both counted correctly, as is the session gap count — 3 in-flow children with `gap={1}` gives exactly 2 gaps, since `Toast` is `position="absolute"` and out of flow.
- Making the props required and deleting the component's internal `useTerminalDimensions` removes the silent-fallback path that would have reintroduced #1067. Both call sites are in-repo, so there's no external breakage.
- The added sidebar-toggle test (`session(150,50,false) → full` vs `session(150,50,true) → medium`) pins the actual reported bug rather than a proxy — the most valuable test in the file.
- Gate isolation is real: `TALL`/`WIDE` derive from the constants, so each test fails if its own `<` is deleted or flipped to `<=`.
- The reactivity note at `welcome-panel.tsx:40-41` is correct — reading `props.x` inside `createMemo` tracks, so the variant recomputes on resize and sidebar toggle.

### Missing tests

- A render test of either call site — the acknowledged gap, and the one that would have caught the Major finding and Minor 3. Given this is the second round on the same bug, I'd treat it as the remaining open item rather than a follow-up.
- A test tying the reserves to the real chrome (`HOME_VERTICAL_RESERVE === SPACER + WRAPPER + PROMPT_REST_HEIGHT + HOME_BOTTOM + FOOTER` from named constants), which would make the Major finding impossible to reintroduce silently.
- The overlay-sidebar case from Minor 4: sidebar open on a ≤120-col terminal.
- Negative available dimensions (an 80×5 terminal yields height `-5` on home); only `(0,0)` and `(1,1)` are covered today.

---

One caveat on the numbers above: I read the layout tree rather than running it, so `PROMPT_REST_HEIGHT` is 5-or-6 pending one measurement against a real render. The conclusion holds either way — even the most conservative count puts 80×24 back on the floor.

saravmajestic:
Thanks for the round-3 pass — the reserve-arithmetic catch was exactly right. All items addressed in `9d4c36cd79` (fixes) + `2762efb97c` (render test).

## Major — reserves undercount the real chrome
Fixed. Reserves are now derived from the counted tree, not estimated:
- `PROMPT_REST_HEIGHT = 6` (top rule 1 + paddingTop 1 + textarea 1 + separator 1 + meta row 1 + idle hint row 1).
- `HOME_VERTICAL_RESERVE = 2 + 1 + 6 + 3 + 3 = 15` (spacer + wrapper + prompt + `home_bottom` + footer).
- `SESSION_VERTICAL_RESERVE = 2 + 1 + 6 = 9`.

And the thresholds are re-tuned separately so they stop compensating for each other:
- `MEDIUM_MIN_HEIGHT: 13 → 8` — the real medium-panel height, so 80×24 is a *genuine* medium fit (home available 24−15 = 9 ≥ 8), not a pinned margin.
- `FULL_MIN_HEIGHT: 34 → 30` — kept explicitly as a product breakpoint (the full box is ~13 rows; we require far more so the wordmark never dominates), documented as such rather than a fit minimum.

## Minor
1. **`PANEL_HORIZONTAL_PADDING` not centralized / sidebar `42` duplicated (also cubic):** `PANEL_HORIZONTAL_PADDING` and a new `SIDEBAR_WIDTH = 42` are exported and used by `session/index.tsx`'s `contentWidth` and by `sessionAvailable`; the test no longer references `42` at all.
2. **Test helpers re-implement the route math:** exported `homeAvailable(w,h)` / `sessionAvailable(w,h,sidebarVisible)` — **both the routes and the tests now call these**, so a wrong reserve/threshold fails a test instead of being mirrored in a private copy. Added a test pinning the reserves to the counted chrome.
3. **Session sizes against the overlay sidebar:** documented as deliberate. The panel shares `contentWidth`'s basis (narrows whenever the sidebar is open, including the dimmed full-area overlay) so the panel and the messages stay aligned and both restore to full width when it closes — sizing to the narrowed column, not the transient obscured width. Test + comment record this.

## Nit
The height thresholds now read for what they are: `MEDIUM_MIN_*` a real fit requirement, `FULL_MIN_*` a product breakpoint (comments explain the ~13-row panel vs the larger required space).

## Missing tests
- **Reserve-tied-to-chrome, overlay narrowing, negative dimensions** — all added.
- **Render test — added, component-level** (`test/component/welcome-panel.test.tsx`): mounts the real `WelcomePanel` through the app context stack and asserts each variant renders its distinct content (full "What is Altimate Code"; medium condensed line; compact CTA), covering the `.tsx` Switch the pure function can't.
- **Full Home/Session route render test — deferred, honestly.** `Home` renders `<Prompt>`, whose dependency graph is ~28 contexts; mounting it is ~2× the heaviest existing test and brittle, with no precedent in the suite. Since the arithmetic is now shared between the routes and the tests (Minor 2), the remaining route surface is a one-line `<WelcomePanel availableWidth={panelAvailable().width} …/>`. Happy to build the full-route harness if you'd still like the end-to-end guard.

## CI
The red "TypeScript" job is a pre-existing unrelated flake — the only failing test was `config HttpApi > serves config with active provider model status` (`httpapi-config.test.ts`, untouched by this PR); it passes locally (2/0). The fresh push should re-run it.

These changes were also cross-checked with a two-model consensus review (both APPROVE).

sahrizvi (approved):
## Review — round 4 (`a5e1b50e`) — approving

The Major finding is genuinely fixed, not papered over. This round I could run the code rather than reason about it from the tree, so the numbers below are executed.

### Verified

**Reserves derive from a named, counted constant.** `PROMPT_REST_HEIGHT = 6`, `HOME_VERTICAL_RESERVE = 2+1+6+3+3 = 15`, `SESSION_VERTICAL_RESERVE = 2+1+6 = 9`. Matches my count of the rendered tree exactly, `home_bottom`'s 3 rows included.

**`MEDIUM_MIN_HEIGHT = 8` is a real fit, not a tuned number.** Counting the medium branch (`welcome-panel.tsx:71-85`): border 2 + padding 2 + title 1 + description 1-2 + CTA 0-1 = **max 8 rows**. At 80×24, home gets 9 available against a panel needing 8 — a genuine one-row margin, with the whole column landing at 23 of 24 rows. Worth calling out that this went the right direction: lowering the threshold to the panel's real height, rather than inflating the reserve to preserve a predetermined answer.

**`FULL_MIN_HEIGHT = 30` is now labelled for what it is** — a product breakpoint, with the ~13-row panel height stated next to it so it can't be misread as a fit minimum.

**The shared arithmetic is real, not cosmetic.** Both routes call `homeAvailable()` / `sessionAvailable()`, and so do the tests — a wrong reserve now fails a test instead of being mirrored in a private copy. `contentWidth` derives from `panelAvailable()`, `SIDEBAR_WIDTH` is exported, and the literal `42` is gone from the codebase. This was the finding most likely to be closed superficially; it wasn't.

**Tests.** 18/18 pass across the two panel files. Full `packages/tui` suite: **220 pass / 9 fail** against **211 pass / 9 fail** on `origin/main` — identical failures, all in sync/hydration files this PR doesn't touch, so they're pre-existing and the 9 added tests all pass. I also re-derived every threshold assertion by hand; all correct.

### Please fix before merge

**Unused imports** — `routes/session/index.tsx:29` still imports `PANEL_HORIZONTAL_PADDING` and `SIDEBAR_WIDTH`, but `contentWidth` now derives from `panelAvailable()`, so neither is referenced. `oxlint` reports both as `no-unused-vars`. One-line fix; flagging it because the previous round advertised oxlint 0/0.

### Follow-ups — non-blocking

**1. The "stay aligned with the messages" rationale doesn't hold in the overlay case** — `welcome-panel-utils.ts:85-89`, `session/index.tsx:276-281`

Worth being precise here, since this is the justification used to close the overlay question from the last round. Messages consume `ctx.width` (= `contentWidth`) as an *explicit* width, but the panel's outer box is `width="100%"` (`welcome-panel.tsx:54`) of a column the overlay does **not** narrow. So on a 100-col terminal with the overlay open, messages render at 54 columns while the panel box still renders ~96 wide and merely selects the `compact` variant — misaligned, and under-filling a full-width box.

The decision itself is defensible (size conservatively while obscured, restore on close) and the effect is cosmetic, but the comment asserts an alignment that doesn't occur. Either narrow the panel wrapper to `contentWidth()` so the claim becomes true, or restate the reason.

**2. `reserves stay pinned to the counted chrome` doesn't do what its name says** — `welcome-panel-utils.test.ts:33-38`

It asserts `HOME_VERTICAL_RESERVE === 15`, a literal restatement of the constant. If the prompt grows a row, the constant stays 15, the real chrome becomes 16, and this test still passes while claiming to pin the reserve to the chrome. Useful as a change-detector against accidental edits, but not the invariant the name promises — and `PROMPT_REST_HEIGHT` isn't exported, so the test can't decompose the sum. Either export the pieces and assert `HOME_VERTICAL_RESERVE === SPACER + WRAPPER + PROMPT_REST_HEIGHT + HOME_BOTTOM + FOOTER`, or rename it to what it actually guards.

### Residual risk — noted, not blocking

The render test is component-level, as called out in the PR description. The route wiring stays unguarded: `availableWidth={panelAvailable().width}` / `availableHeight={panelAvailable().height}` — swapping those two compiles and passes every test. The risk is low now that both routes go through one shared function and the reserve/threshold logic is unit-tested, so I agree with deferring the full-route harness rather than mounting `Prompt`'s ~28-context dependency graph for it.

---

Approving. The reserve arithmetic is honest, the threshold re-tune is derived from the real panel height instead of reverse-engineered from a desired outcome, and the shared-arithmetic change fixes the root cause rather than the symptom. Just drop the two unused imports on the way in.

PR #1074, opened 2026-08-05
fix: Windows grep outage and test-run telemetry pollution

### LINKED ISSUE #1072: Windows: grep broken for 99 machines — ripgrep zip extraction shells out to PowerShell
## Description

`grep` fails on a large share of Windows users. Azure telemetry for 2026-07-22 → 2026-08-05 shows **328 `core_failure` events across 99 distinct Windows machines** (of 617 Windows machines active in the window), all with:

```
? is not recognized as an internal or external command, operable program or batch file.
```

plus the German, French, Spanish and Portuguese translations of the same cmd.exe message. Present on released 0.9.2, 0.9.3 and 0.9.4. The `?` is our own PII masker — `Telemetry.maskString` replaces single-quoted text with `?`, and cmd.exe quotes the command name, so the identity of the failing binary was destroyed before it reached telemetry.

## Root cause

ripgrep's Windows release is a zip, and `packages/core/src/ripgrep/binary.ts` extracts it by shelling out:

```ts
const shell = which("powershell.exe") ?? which("pwsh.exe") ?? "powershell.exe"
```

When neither resolves, that literal string reaches `cross-spawn`. `parseNonShell()` sets `needsShell = true` whenever `resolveCommand()` returns undefined and re-spawns through `cmd.exe /d /s /c`, so cmd.exe answers with the "not recognized" message. `throw new Error(result.stderr.trim())` then makes that string the error verbatim.

Because `RipgrepBinary.filepath` is wrapped in `Effect.cached`, a single failed extraction breaks `grep` for the rest of the session.

Note cross-spawn resolves against `process.env.PATH`, not our augmented PATH — so the affected machines are ones where PowerShell is missing **from PATH**, not necessarily absent from the machine.

## Blast radius

Wider than the `grep` tool: `@opencode-ai/core/ripgrep` also backs the HTTP-API file handlers (`server/routes/instance/httpapi/handlers/file.ts`) and `cli/cmd/debug/ripgrep.ts`.

`glob`, `ls` and `skill` are unaffected — they use the `packages/opencode/src/file/ripgrep.ts` shim, which already unzips in-process.

## Upstream

Upstream carries the same fragility. anomalyco/opencode#24291 is open, reporting `Expand-Archive` unusable when spawned from the Bun-compiled binary and affecting `grep`/`glob`/`skill`. Their #23457 fix only corrected how paths were passed to PowerShell (the `$args` → inlined-and-escaped form we already carry); it did not remove the dependency on PowerShell resolving. `upstream/dev` still has the same code.

## Expected

Extracting ripgrep should not depend on an external shell being resolvable.

### LINKED ISSUE #1073: Telemetry: test runs ship to the production sink, inflating machine counts by ~33%
## Description

Our own test runs ship telemetry to the production Application Insights resource, inflating every install and active-machine metric.

In the 2026-07-22 → 2026-08-05 window, **1,020 of 3,135 distinct machine ids** emitted `provider_id="test"` and `cli_version="local"` — i.e. test processes, which regenerate their machine id on every run. That is roughly a third of the headline machine count.

## Root cause

`doInit()` in `packages/opencode/src/altimate/telemetry/index.ts` gates only on:

```ts
if (process.env.ALTIMATE_TELEMETRY_DISABLED === "true") { ... }
```

There is no test/CI guard, so the baked-in `DEFAULT_CONNECTION_STRING` is used by the test suite. `bun test` sets `NODE_ENV=test`, so this affects both CI and developer machines running tests locally.

## Impact

Any dashboard counting machines, installs, or new users is inflated by ~33%. It also means adoption trends can move purely because CI volume changed.

## Constraint on the fix

The gate must key on **test runners, not CI**. `altimate-code-actions` wraps this CLI, so every run of that shipped product sets `CI`/`GITHUB_ACTIONS` — gating on those would blind a real product surface.

Telemetry suites that point at their own sink via an explicit `APPLICATIONINSIGHTS_CONNECTION_STRING` must keep working.

anandgupta42:
### Issue for this PR

Closes #1072
Closes #1073

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Two defects found while reading Azure telemetry for 2026-07-22 → 2026-08-05.

**1. `grep` was broken for 99 Windows machines (#1072).**

ripgrep's Windows release is a zip, and we extracted it by shelling out to
`powershell.exe -Command Expand-Archive`, with a literal `"powershell.exe"` string as the
fallback when neither `powershell.exe` nor `pwsh.exe` resolved. cross-spawn's
`parseNonShell()` sets `needsShell = true` whenever `resolveCommand()` returns undefined and
re-spawns through `cmd.exe /d /s /c`, so cmd.exe replied `'…' is not recognized as an internal
or external command`. We threw that string verbatim, and since `RipgrepBinary.filepath` is
`Effect.cached`, one failed extraction killed `grep` for the whole session.

This extracts the zip in-process with `@zip.js/zip.js`, so there is no external shell to
resolve. That is the same approach `packages/opencode/src/file/ripgrep.ts` already uses — which
is why `glob`/`ls`/`skill` never showed this failure while `grep` did.

Three things came out of review and are included:

- **`checkSignature: true`.** zip.js defaults CRC verification *off*, so a corrupt download
  decodes "successfully", gets written to the cache, and is trusted by every later session.
- **Atomic install.** `filepath` trusts the cached binary on existence alone, so an interrupted
  write left a truncated `rg.exe` that was reused forever. Both paths now stage to `rg.exe.tmp`
  and rename. CRC cannot catch this — it is verified before the write.
- **Attributed failures.** Child stderr was reported verbatim, so a shell-level failure looked
  like a tool bug. A resolve failure is also memoized, so it is re-reported on every later grep.
  Typed FS/HTTP failures are now wrapped as `ripgrep binary resolve failed: …`.

Upstream has the same fragility open as anomalyco/opencode#24291. Their #23457 fix only changed
how paths were passed to PowerShell (the inlined-and-escaped form we already carry); it did not
remove the dependency. `upstream/dev` still has the original code.

**2. Test runs polluted production telemetry (#1073).**

1,020 of 3,135 machine ids were test processes (`provider_id="test"`, `cli_version="local"`),
which regenerate their machine id every run — inflating install and active-machine counts by
about a third. `doInit()` now refuses the baked-in connection string under `NODE_ENV=test`,
`BUN_TEST`, `VITEST` or `JEST_WORKER_ID`.

Keyed on test runners and deliberately **not** on CI: `altimate-code-actions` wraps this CLI, so
every run of that shipped product sets `CI`/`GITHUB_ACTIONS`, and gating on those would blind a
real product surface. `bun test` sets `NODE_ENV=test`, which covers CI and developer machines
with one condition. An explicit `APPLICATIONINSIGHTS_CONNECTION_STRING` is always honoured, so
suites with their own sink are unaffected; `ALTIMATE_TELEMETRY_FORCE=true` overrides the
default-sink refusal, and the existing opt-outs still win over both.

- **8 ripgrep tests**, including a layer-level test that drives `filepath` through the Windows
  zip path with stubbed HTTP/FS and **a spawner that throws if invoked** — pinning that the
  Windows path spawns nothing and that the install is staged-then-renamed. The CRC test was
  confirmed to fail without `checkSignature: true`.
- **15 telemetry tests**, including that CI on its own must still report, and one that stubs no
  env at all so it relies on the real runner's `NODE_ENV` — the assumption the gate rests on.
- `bun turbo typecheck` green; marker guard green.
- Full suites run against a `main` baseline worktree: `packages/core` 1,038 pass,
  `packages/opencode` 11,014 pass, with **zero failures unique to this branch** (the remaining
  MCP/plugin failures reproduce on `main`).

Not verified on a real Windows host — I have no Windows runner. The zip decoding, install and
no-spawn behaviour are covered by tests; the end-to-end "grep works on a machine without
PowerShell on PATH" claim is inferred from the call graph, not executed.

Known follow-ups, deliberately not in this PR: bundling `rg` so there is no runtime download
(upstream wants this too — anomalyco/opencode#31734), unifying the two ripgrep resolvers, and
making `filepath` cache only successes so a transient failure stops poisoning the session.

### Screenshots / recordings

_Not a UI change._

### Checklist

---
## Summary by cubic
Fixes Windows grep outages by extracting ripgrep in-process (no PowerShell) and installing it atomically, and stops automated tests from sending production telemetry. Adds a Windows CI E2E that resolves and runs `rg.exe` with PowerShell removed from PATH (fixes #1072, #1073).

- **Bug Fixes**
  - Windows grep (#1072): in-process zip decode via `@zip.js/zip.js` with CRC; atomic install (stage → rename); clearer, named HTTP/extract errors; normalized failure message `ripgrep failed with code X: …` with a “no output” fallback; Windows-only E2E verifies download/extract/install/exec with PowerShell off PATH.
  - Telemetry (#1073): suppress the default App Insights under test runners (`NODE_ENV=test`, `BUN_TEST`, `VITEST`, `JEST_WORKER_ID`); CI alone still reports; explicit `APPLICATIONINSIGHTS_CONNECTION_STRING` honored; `ALTIMATE_TELEMETRY_FORCE=true` opt-in; docs updated.

- **Dependencies**
  - Added `@zip.js/zip.js@2.7.62` to `packages/core`.

<sup>Written for commit 83961acb83abec1011506e418477e62bf9f9695d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1074?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ripgrep installation on Windows with in-process extraction, CRC validation, atomic staging/rename, and better contextual errors.
  - Enhanced ripgrep process failure reporting to always include normalized stderr and exit codes (or a clear “no output” fallback).

- **Telemetry**
  - Automated test runs now avoid the baked-in production App Insights sink; CI remains enabled.
  - Explicit connection strings and `ALTIMATE_TELEMETRY_FORCE=true` still enable telemetry, with disable overrides taking precedence.

- **Documentation**
  - Added documentation for telemetry test-run exclusions and override behavior.

- **Tests / CI**
  - Added Windows ripgrep end-to-end coverage and automated telemetry tests.
  - Added a Windows-only CI job for ripgrep resolution E2E.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -379,6 +379,39 @@ jobs:
           $config.Output.Verbosity = "Detailed"
           Invoke-Pester -Configuration $config
 
+  # altimate_change start — Windows ripgrep E2E (issue #1072)
+  # ---------------------------------------------------------------------------
+  # Real Windows check for ripgrep binary resolution. Downloads and extracts the
+  # actual archive with PowerShell stripped from PATH, then executes the binary.
+  # This is the condition that broke grep for 99 Windows machines; no amount of
+  # unit testing on Linux/macOS covers it.
+  # ---------------------------------------------------------------------------
+  windows-ripgrep-e2e:
+    name: Windows ripgrep E2E
+    needs: changes
+    if: needs.changes.outputs.typescript == 'true' || github.event_name == 'push'
+    runs-on: windows-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - uses: oven-sh/setup-bun@ecf28ddc73e819eb6fa29df6b34ef8921c743461 # v2
+        with:
+          bun-version: "1.3.14"
+
+      # --ignore-scripts: `tree-sitter-powershell` has no Windows prebuild and its fallback
+      # compile needs Visual Studio Build Tools, which this runner does not have (see
+      # anomalyco/opencode#25563). This check only needs the pure-JS dependency graph
+      # (effect, @zip.js/zip.js, which, xdg-basedir), so skipping lifecycle scripts is enough.
+      - name: Install dependencies
+        run: bun install --ignore-scripts
+
+      # Run from packages/core so `effect` and the other deps resolve — they are not root deps.
+      - name: Resolve ripgrep with PowerShell unavailable
+        working-directory: packages/core
+        run: bun run script/windows-ripgrep-e2e.ts
+  # altimate_change end
+
   # ---------------------------------------------------------------------------
   # dbt-tools E2E — slow (~3 min), only on push to main.
   # Tests dbt CLI fallbacks against real dbt versions (1.8, 1.10, 1.11) and
--- a/bun.lock
+++ b/bun.lock
@@ -95,6 +95,7 @@
         "@opentelemetry/sdk-trace-base": "2.6.1",
         "@parcel/watcher": "2.5.1",
         "@silvia-odwyer/photon-node": "0.3.4",
+        "@zip.js/zip.js": "2.7.62",
         "ai-gateway-provider": "3.1.2",
         "bun-pty": "0.4.8",
         "cross-spawn": "catalog:",
--- a/docs/docs/reference/telemetry.md
+++ b/docs/docs/reference/telemetry.md
@@ -126,6 +126,24 @@ export ALTIMATE_TELEMETRY_DISABLED=true
 
 When telemetry is disabled, no events are sent and no network requests are made to the telemetry endpoint.
 
+### Test runs are excluded
+
+Test runners never reach the default telemetry endpoint. Telemetry is suppressed when `NODE_ENV=test`,
+`BUN_TEST`, `VITEST`, or `JEST_WORKER_ID` is present. This exists because test processes regenerate
+their machine ID on every run, so without the exclusion they dominate install and active-machine counts.
+
+Running in CI is **not** excluded — that is ordinary product usage (for example
+[altimate-code-actions](https://github.com/AltimateAI/altimate-code-actions) wraps this CLI), so
+`CI` and `GITHUB_ACTIONS` on their own do not suppress anything.
+
+Two escape hatches exist for reporting from a test run deliberately:
+
+- Set `APPLICATIONINSIGHTS_CONNECTION_STRING` to your own endpoint — an explicitly-configured sink
+  is always honoured, which is how the project's own telemetry tests work.
+- Set `ALTIMATE_TELEMETRY_FORCE=true` to use the default endpoint anyway.
+
+`ALTIMATE_TELEMETRY_DISABLED` and the config opt-out take precedence over both.
+
 ## Privacy
 
 We take your privacy seriously. Altimate Code telemetry **never** collects:
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -97,6 +97,7 @@
     "@opentelemetry/sdk-trace-base": "2.6.1",
     "@parcel/watcher": "2.5.1",
     "@silvia-odwyer/photon-node": "0.3.4",
+    "@zip.js/zip.js": "2.7.62",
     "@openrouter/ai-sdk-provider": "2.9.0",
     "ai-gateway-provider": "3.1.2",
     "bun-pty": "0.4.8",
--- a/packages/core/script/windows-ripgrep-e2e.ts
+++ b/packages/core/script/windows-ripgrep-e2e.ts
@@ -1,0 +1,150 @@
+/**
+ * Windows end-to-end check for ripgrep binary resolution.
+ *
+ * Reproduces as much as a GitHub runner allows of the condition behind the outage in
+ * https://github.com/AltimateAI/altimate-code/issues/1072: a cold cache on a Windows machine
+ * where PowerShell cannot be resolved from PATH. The old implementation extracted ripgrep's zip
+ * by shelling out to `powershell.exe -Command Expand-Archive`, so cross-spawn fell back to
+ * `cmd.exe /d /s /c` and the extraction died with "is not recognized as an internal or external
+ * command" — which `Effect.cached` then replayed for the rest of the session.
+ *
+ * This performs a real download and a real extraction, then executes the resulting binary. It is
+ * deliberately not a unit test: the point is to exercise the actual filesystem, the actual archive
+ * and the actual process launch on a real Windows host.
+ *
+ * Scope note: stripping PATH does NOT make PowerShell unspawnable on Windows (see the control
+ * probe at the end), so this does not reproduce the affected machines. The guarantee that
+ * extraction spawns nothing at all is established by test/ripgrep-windows.test.ts.
+ *
+ * Run: bun run script/windows-ripgrep-e2e.ts   (from packages/core — `effect` resolves there)
+ */
+import { execFileSync } from "node:child_process"
+import { mkdtempSync, existsSync, readdirSync, statSync, rmSync } from "node:fs"
+import { tmpdir } from "node:os"
+import path from "node:path"
+
+function fail(message: string): never {
+  console.error(`FAIL: ${message}`)
+  process.exit(1)
+}
+
+function ok(message: string) {
+  console.log(`ok   ${message}`)
+}
+
+if (process.platform !== "win32") fail(`this check only means anything on Windows (got ${process.platform})`)
+
+// Cold cache: point the XDG cache root at a throwaway directory *before* importing anything that
+// reads it, since Global computes its paths at module load.
+const cacheRoot = mkdtempSync(path.join(tmpdir(), "rg-e2e-"))
+process.env.XDG_CACHE_HOME = cacheRoot
+process.env.LOCALAPPDATA = cacheRoot
+
+// Remove every PATH entry that could provide PowerShell. System32 itself is kept so cmd.exe and
+// the rest of Windows still work — this simulates the locked-down/sanitized-PATH machines in the
+// telemetry, not a broken OS.
+const originalPath = process.env.PATH ?? process.env.Path ?? ""
+const stripped = originalPath
+  .split(path.delimiter)
+  .filter((entry) => entry && !/powershell/i.test(entry))
+  .join(path.delimiter)
+process.env.PATH = stripped
+process.env.Path = stripped
+
+const { which } = await import("../src/util/which")
+
+// Guard against a vacuous run: if PowerShell is still resolvable, this proves nothing.
+const ps = which("powershell.exe")
+const pwsh = which("pwsh.exe")
+if (ps || pwsh) fail(`PowerShell is still resolvable (${ps ?? pwsh}) — the scenario was not reproduced`)
+ok("PowerShell is not resolvable on PATH (failure condition reproduced)")
+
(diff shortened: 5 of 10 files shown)


anandgupta42:
Thanks — addressed in the latest push. Triage below, including two findings I'm rejecting with evidence.

## Fixed

**Concurrent installs shared `${target}.tmp`** — flagged independently by CodeRabbit, cubic (×2) and Codex. Valid and the most serious of the batch. Staging is now unique per attempt (`rg.exe.<pid>.<rand>.tmp`), so two cold-cache processes can no longer publish each other's partial download.

Also fixed the related Windows hazard cubic raised: `remove(target)` ran *before* `rename`, with its failure ignored, so a failing rename could leave no binary at all. It now attempts the rename first and only removes-and-retries if that fails — a rename failing for any other reason leaves the existing binary untouched. Staged files are cleaned up on error via `Effect.onError`.

**HTTP download failures lost ripgrep attribution** (@kilo-code-bot) — correct, and a real regression from moving attribution into `install`. A bare `HttpClientError` is now wrapped as `ripgrep download failed from <url>: …`. This matters because `filepath` is `Effect.cached`, so a proxy 403 gets re-reported on every later grep in the session. The PR description overstated the previous state; it's accurate now.

**`isAutomatedRun` inside the namespace** (@chatgpt-codex-connector) — correct per `packages/opencode/AGENTS.md:42-44`; moved to a non-exported top-level declaration.

**Tests wrote to the real home directory** (cubic) — correct. Non-suppressed cases reach `doInit`'s machine-id block, which uses `os.homedir()` directly (`telemetry/index.ts:1730`), so the suite minted a real `~/.altimate/machine-id` and assertions depended on pre-existing state. `HOME`/`USERPROFILE` now point at a temp dir per case, removed in `finally`.

## Rejecting — verified against the code

**`failure()` produces `[object Object]`** (CodeRabbit, `ripgrep.ts:147`) — not correct. `Error` in that module is not the global `Error`; `ripgrep.ts:43` declares:

```ts
export class Error extends Schema.TaggedErrorClass<Error>()("Ripgrep.Error", {
  message: Schema.String,
  cause: Schema.optional(Schema.Defect),
}) {}
```

It takes a struct, and `message` is a real string field. The verification script in the review ran `new Error({message, cause})` in a bare Node process, which resolves the *global* `Error` — hence `[object Object]`. Constructing the actual class gives `message === "ripgrep failed with code 3: boom"`. Applying the proposed diff would break the typed error.

**`Effect.result` exposes the error on `exit.cause`, not `exit.failure`** (Codex, `ripgrep-windows.test.ts:83`) — not correct for this Effect version (`4.0.0-beta.74`):

```
Effect.runPromise(Effect.result(Effect.fail(new Error("boom"))))
→ keys: failure | _tag: Failure | failure?.message: "boom" | cause: undefined
```

The tests pass; had `.failure` been undefined they'd have thrown on the property access rather than reporting the intended diagnostic.

**`Effect.orDie` before the outer `mapError`** (cubic, `binary.ts:186`) — moot now. That outer wrapper no longer exists; attribution happens at the download and install sites instead. The two `.orDie` calls on `fs.isFile` are pre-existing upstream behaviour and unchanged by this PR.

## Not doing here

A cross-process **lock** (CodeRabbit's stronger suggestion) is deliberately out of scope. Unique staging plus atomic rename removes the corruption window; the remaining race is two processes each installing a valid, identical binary and one rename winning — which is harmless. A real lock needs a cross-platform advisory-lock primitive we don't have, and this PR is a hotfix for a live Windows outage.

Verification: `turbo typecheck` green, marker guard green, `packages/core` 1,038 pass, `packages/opencode` suite unchanged — no failures beyond the pre-existing set on `main`.

anandgupta42:
One more addressed.

**Serialize the shared-state test suites** (CodeRabbit) — valid, applied. Both new suites mutate process-wide state: the telemetry one touches `process.env`, the global `fetch` and the module-level `Telemetry` singleton (which `afterEach` shuts down); the ripgrep layer test redefines `process.platform`/`process.arch`. Both `describe` blocks are now `describe.serial` (confirmed available in the pinned Bun 1.3.14).

To be precise about the risk: Bun runs a file's tests sequentially by default and this repo does not pass `--concurrent`, so nothing was actually racing today. The value is that the constraint is now encoded rather than implicit — adding `--concurrent` later would otherwise have turned these into intermittent failures with a confusing signature.

Verification after this change: `turbo typecheck` green, marker guard green, `packages/core` 1,038 pass, `packages/opencode` 11,014 pass — no failures beyond the pre-existing set on `main`.

anandgupta42:
## Now tested on real Windows — with an important caveat

Answering the direct question: until now, **no** — nothing here had run on Windows, which was the obvious gap given this fixes a Windows-only failure. There is now a `Windows ripgrep E2E` job (`windows-latest`) running `packages/core/script/windows-ripgrep-e2e.ts`. It is not a unit test: it does a real download of the real archive, extracts it, installs it and executes the binary.

Output from the run on this PR:

```
ok   PowerShell is not resolvable on PATH
ok   cold cache at C:\Users\RUNNER~1\AppData\Local\Temp\rg-e2e-…\altimate-code\bin
     downloading ripgrep … ripgrep-15.1.0-x86_64-pc-windows-msvc.zip
ok   resolved …\bin\rg.exe
ok   binary present, 4266496 bytes
ok   no staging files left behind
ok   executes: ripgrep 15.1.0 (rev af60c2de9d)
ok   search returns matches
ok   second resolve hits the cache
```

So on real Windows the new path downloads, extracts in-process, installs atomically and produces a working binary. The staging assertion also covers the atomic-install change reviewers asked for.

### What this job does *not* establish

I first wrote it to assert that the **old** PowerShell approach fails under the same conditions, so the run would be a true counterfactual. That assertion failed — and the failure is the useful part:

```
FAIL: the old PowerShell approach still succeeds here — this environment does not
      reproduce the bug, so the run above proves nothing
```

Emptying `PATH` does not make PowerShell unspawnable on Windows. `cross-spawn` falls back to `cmd.exe /d /s /c`, and Windows process creation searches beyond `PATH` (the caller's directory, the system directories, the App Paths registry key), so `powershell.exe` still starts on a stock runner even though `which()` cannot see it.

I have therefore **not** reproduced the affected machines, and I am not claiming this job proves the fix resolves #1072 for them. The probe is now informational and prints that caveat rather than failing.

What carries that argument instead is structural, and it is platform-independent: `test/ripgrep-windows.test.ts` drives `filepath` through the Windows zip branch with a `ChildProcessSpawner` that **fails the test if anything is spawned at all**. Since extraction launches no process, PowerShell's availability cannot affect it. The Windows job proves the path genuinely works on Windows; the unit test proves it needs no shell.

### Other e2e coverage added this round

The subprocess/e2e pass CI runs separately (`test/cli/acp`, `smokes`, `serve`, `run-process`, `mcp-add`, `help-snapshots`) — run locally at the CI concurrency setting: **37 pass, 0 fail**.

One incidental find: `bun install` cannot complete on `windows-latest` because `tree-sitter-powershell` has no Windows prebuild and its fallback compile needs Visual Studio Build Tools (same as anomalyco/opencode#25563). The job uses `--ignore-scripts`, which is sufficient here. Worth knowing if anyone else adds Windows CI.

PR #1085, opened 2026-08-10
fix: [#1052] bundled cleanup — 5 deferred items + 3 review-fixes

### LINKED ISSUE #1052: Post-release cleanup: deferred items from v0.9.4 multi-persona review
Consolidated tracking issue for items surfaced by the v0.9.4 pre-release multi-persona review (CTO, PM, Data Engineer, Tech Lead, Support Engineer) and deliberately deferred out of the release because each requires more than a 30-min patch (new design work, performance experiment, cross-platform testing, or org-level infra).

All items were re-verified against the release/v0.9.4 tree; the review's original list of 9 was reduced to 8 after empirical checks (`packages/tui` typecheck runs green — that item was a false positive).

---

### D1 — Interactive activation-menu picker (arrow-key selectable rows in the chat)

**What:** The activation menu that ends every `/onboard-connect` branch is currently plain LLM-emitted text — the user types the number instead of arrow-selecting a row.

**Why deferred:** Design decision, not a 30-min fix. The plain-text design was chosen in commit `e89955fe3e` because (a) an earlier `DialogActivation` modal never rendered for users with pre-loaded creds (auth-arrival timing meant the `false → true` transition never fired), and (b) a fixed option-set can't personalize per scan verdict, whereas the LLM branches on warehouse/dbt discovery. Reverting to a picker needs to solve both problems — either intercept LLM-emitted numbered messages in the TUI (fragile heuristic) or introduce a structured tool-call schema and route selection back through the LLM.

**Attribution:** Design decision — @haider (`e89955fe3e`).

---

### D2 — Retire the `globalThis` bandaid on the Question pending-registry

**What:** `packages/opencode/src/question/index.ts:165` anchors the `pendingByDir` registry on `globalThis["__altimateQuestionPending"]` so every duplicated module copy shares one Map. Root cause is dual module-graph loading in the runtime (compiled bundle vs. dev-mode source imports), not the registry itself.

**Why deferred:** Correct fix is fixing the module-graph duplication so there is only one `question` module. That means auditing every import path in `packages/opencode`, `packages/tui`, and the plugin surface for both dev and compiled builds — bigger than a release patch.

**Attribution:** @saravanan-altimate (`8c9a9a893b7`, 2026-07-24).

---

### D3 — Server-side shell-arg validation for user-pasted dbt paths (move defense out of the template)

**What:** `packages/opencode/src/command/template/onboard-connect.txt:197-213` instructs the LLM to refuse paths containing `; & | > < $ \` ( ) { } newline` or single-quotes, verify executability with `test -x`, and always wrap in single quotes. That defense-in-depth is real but template-level — a model that ignores the instructions could still shell-inject.

**Why deferred:** Requires designing a validation layer at the tool boundary (probably a new `dbt_binary_validate` server tool or a bash-wrapper allowlist) and rewriting the template to route through it. Not mechanical.

**Attribution:** @saravanan-altimate (`4f12e7d08ef`, PR #1001 landing).

---

### D4 — Slim the bundled `target/manifest.json` (gzip + inflate on materialize)

**What:** `packages/opencode/sample-projects/jaffle-shop-duckdb/target/manifest.json` ships at 684KB uncompressed. Every `npm install -g altimate-code` downloads it; every `sample_setup` copies it byte-for-byte (with `{{SAMPLE_ROOT}}` sentinel rehydration).

**Why deferred:** Needs (a) build-step change to gzip the shipped copy, (b) inflate step in `copySampleTree`, (c) size measurement + package-size regression guard. Cheap on wall-clock, but touches packaging.

**Attribution:** @saravanan-altimate (`4f12e7d08e`, PR #1001).

---

### D5 — Windows long-path handling in `copySampleTree`

**What:** `packages/opencode/src/altimate/onboarding/materialize.ts:471` uses stock `fs.cpSync` / `fs.copyFileSync` / `fs.mkdirSync`. On Windows, paths longer than 260 chars need the `\\?\` prefix unless the process opts in to long-path support. A user materializing into a deeply-nested directory would hit `ENAMETOOLONG`.

**Why deferred:** Fix is a Windows CI matrix + long-path guard. Bun's `fs` may or may not paper over this; needs empirical test on a Windows runner before choosing a solution.

**Attribution:** @saravanan-altimate (`4f12e7d08ef`, PR #1001).

---

### D6 — Dedup question events at the `/event` boundary (or formalize the dual-emit contract)

**What:** `packages/opencode/src/question/index.ts:100-130` deliberately mirrors `question.asked` / `question.replied` / `question.rejected` via `Bus.publish` in addition to `EventV2` — the webview subscribes to the wildcard PubSub feed and otherwise never sees the events. The `EventV2Bridge.listen` forwarder ALSO republishes to `GlobalBus`, so `/global/event` consumers see each event twice (different top-level ids). Verified in-repo consumers (TUI `sync.tsx`, `notifications.ts`, trace consumer) all dedupe or ignore — external subscribers to `/event` that don't dedupe would double-count.

**Why deferred:** Two viable fixes: (a) dedupe in the `/event` handler itself (server-side, safe for external subscribers), or (b) drop one of the two publish paths and migrate every internal consumer. Both require SDK contract review — this is public API.

**Attribution:** @saravanan-altimate (`8c9a9a893b7`, 2026-07-24).

---

### D8 — Sitewide internal-tracker-ref scrubber + `pre-push` git hook

**What:** `altimate-code` is a public repo; the org rule is that internal-tracker references (project prefix + numeric id, and the internal Atlassian host) must never land in branch names, commits, code, or docs on this repo. Enforcement today is manual (author discipline + release-time grep). No `.githooks/pre-push` or CI step blocks a leak automatically.

**Why deferred:** Needs a script that scans staged content + branch name + commit message trailer, a `.githooks` entry that runs it, and a repo-level `git config core.hooksPath .githooks` bootstrap (documented in `CONTRIBUTING.md` for new clones). Also needs a CI mirror (contributors who skip local hooks) — probably a lightweight GitHub Action on push. Not urgent, but the class of bug is real (already almost leaked twice this quarter).

**Attribution:** Org infra gap — no single author.

---

### D9 — `useConnected()` regression detector: check against user's prior configured state, not just current

**What:** `packages/tui/src/component/use-connected.tsx` returns whether *any* registered provider counts as connected — pure point-in-time check. The v0.9.4 grandfather fix (in-flight on `release/v0.9.4`) trusts non-OpenCode registrations without cost metadata to avoid force-feeding the welcome picker to legitimate BYOK / self-hosted / custom-registration users. That's the right call for upgrades from v0.9.3, but it doesn't defend against future regressions the same way — the check has no memory of what the user had connected yesterday.

**Why deferred:** Proper detector needs persisted "last-known-connected" state (probably stashed in the auth manifest) and a comparison at bootstrap. Small feature, but crosses TUI, auth, and manifest storage — bigger than the release-window fix.

**Attribution:** Original file @anandgupta42 (`c985718c4df`, 2026-06-23); current grandfather fix @haider (release/v0.9.4, uncommitted at time of writing).

---

## Summary

| ID | Item                                                    | Attribution                                   |
|----|---------------------------------------------------------|-----------------------------------------------|
| D1 | Interactive activation-menu picker                      | Haider (`e89955fe3e` design decision)        |
| D2 | Retire globalThis Question bandaid                      | Saravanan (`8c9a9a893b7`)                    |
| D3 | Server-side shell-arg validation for dbt paths          | Saravanan (`4f12e7d08ef`, PR #1001)          |
| D4 | Slim bundled `target/manifest.json`                     | Saravanan (`4f12e7d08e`, PR #1001)           |
| D5 | Windows long-path in `copySampleTree`                   | Saravanan (`4f12e7d08ef`, PR #1001)          |
| D6 | Dedup question events at `/event` boundary              | Saravanan (`8c9a9a893b7`)                    |
| D8 | Sitewide internal-tracker-ref scrubber + `pre-push` hook | Org infra gap                                |
| D9 | `useConnected()` regression detector vs. prior state    | Original @anandgupta42; current fix @haider  |

Milestone: v0.9.4
Label: post-release-cleanup

sahrizvi:
### Issue for this PR

Closes #1052 (partial — see "Not in this PR" below)

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Fixes 5 of 13 deferred items from #1052 (the v0.9.4 post-release cleanup tracking issue), plus 3 review-fixes flagged by a 6-model consensus code review on the initial 5 commits, plus one follow-up scrub.

**Scope:** the deferred items trace back to my own commits, so the split here is by size rather than by ownership. This batch takes the five that are self-contained enough to review together — D8, D10, D11, D12 and D14. The rest each want a change of their own: D9 crosses TUI + auth + manifest storage, D13 is a gateway-protocol change rather than a CLI patch, and D2–D6 span server-side validation, bundle slimming, Windows path handling and the `/event` contract. Bundling those in would bury the work that is here.

### Commits (bottom-up, oldest first)

Original D-item fixes:
1. `chore(hygiene): [#1052 D8] pre-push scan for internal-tracker refs` — scanner + hook + docs. CI mirror deferred to a follow-up PR (needs `workflow`-scoped token).
2. `test(build): [#1052 D10] stamp-based staleness guard for the smoke test` — build.ts emits `dist/<target>/bin/build-inputs.json` with sha256s; smoke test compares against it.
3. `test(harness): [#1052 D11] idempotent retry in cli-process.run()` — scrub `opencode*.db*` before the SQLite-lock retry so the second attempt starts clean.
4. `test(tui): [#1052 D12] deterministic regression test for phaseLabel()` — util-level unit test replacing the flaky PTY e2e.
5. `fix(models): [#1052 D14] drop eager import-time ModelsDev.refresh()` — the fetch was holding the event loop under `unshare --net` on CI. Snapshot covers cold-start.

Review-fixes (from a 6-model consensus code review of the above):
6. `fix(hygiene): [#1052 D8 review-fix M1] catch suffix-adjacent tracker leaks + self-test` — regex missed `<prefix>-<n><suffix>` (5/6 reviewers flagged, some as CRITICAL). Also adds a 28-case self-test file for the RULES.
7. `test(build): [#1052 D10 review-fix M2] widen stamp to workspace packages + lockfile` — the original stamp missed `packages/{tui,core,util,plugin,...}/src`, the workspace-root `package.json`, and `bun.lock`. Now walks each workspace `src/` and hashes those files.
8. `fix(models): [#1052 D14 review-fix M3] don't crash on non-JSON error body + fire refresh at boot` — pre-existing `JSON.parse(<HTML>)` crash on 5xx from models.dev + a fire-and-forget `Promise.resolve().then(refresh)` so short-lived commands still warm the cache without holding the loop.
9. `fix(hygiene): [#1052 D8 review-fix follow-up] scrub example strings from scanner source` — the M1 commit accidentally embedded concrete tracker-key literals in its own test file + doc comments. Fixtures now build the strings at runtime; scanner source drops the verbatim examples.

### Not in this PR (still open on #1052)

- **D1** — Interactive activation-menu picker. Design decision, not mechanical.
- **D2, D3, D4, D5, D6** — deferred on size, not ownership. Each sits in its own area (Question pending-registry, server-side shell-arg validation, manifest slimming, Windows long paths, `/event` dedup) and is better reviewed on its own.
- **D7** — pruned (false positive) at v0.9.4 review time.
- **D9** — `useConnected()` regression detector. Crosses TUI + auth + manifest storage; deserves its own PR.
- **D13** — SSH port-forwarding / device-code flow. Gateway-protocol change, not a CLI patch.
- **CI mirror for D8** — the `.github/workflows/tracker-leak-check.yml` file was in the local branch but this session's token lacks the `workflow` scope; add it in a follow-up PR.

### Test plan

- [x] `bun turbo typecheck` — 13/13 clean
- [x] `bun test packages/opencode/test/skill/tracker-leak-check.test.ts` — 28/28 (scanner RULES)
- [x] `bun test packages/tui/test/util/phase-label.test.ts` — 4/4 (phaseLabel util)
- [x] `bun script/check-tracker-leaks.ts` on the branch itself — silent (no leaks)
- [x] `bun run --cwd packages/opencode --conditions=browser ./src/index.ts --version` — exits in ~1s, no hang
- [ ] Full pre-release sanity (Verdaccio Phase 3 [10/10] no-internet check) — will run in CI. Note: D14's fire-and-forget refresh may reintroduce the sanity hang; if it does, the fix is dropping one line (documented in the D14 review-fix commit body).

### D14 caveat worth explicit reviewer attention

Commit 8 adds `Promise.resolve().then(() => ModelsDev.refresh().catch(() => {}))` alongside the hourly interval. A microtask doesn't itself keep Bun alive, and the fetch it schedules will just be abandoned at process-exit if unresolved — but if this reintroduces the original v0.9.4 blocker (sanity Phase 3 [10/10] under Linux `unshare --net`), the fix is to delete that one line and take the "snapshot-only cold-start" trade-off documented in commit 5.

https://claude.ai/code/session_01Q8FGy89Qpr39k8nCSpCcK2

---
## Summary by cubic
Closes five deferred items from #1052 and hardens release safety: blocks tracker leaks on push, makes binary staleness detection accurate across all embedded inputs, and prevents models.dev cache poisoning or boot-time hangs.

- Pre-push tracker check: `script/check-tracker-leaks.ts` now scans the refs actually being pushed (reads pre-push stdin) including source and destination ref names, commit messages ahead of base, and added diff lines only. Skips deletion-only pushes, tolerates unborn HEAD, warns when base is missing, validates `--base`, and uses shell-free `git` calls with a precise diff filter. Hook runs from `.husky/pre-push` after typecheck (bypass with `SKIP_TRACKER_CHECK=1`).
- Build staleness guard: `packages/opencode/script/build.ts` writes `dist/<target>/bin/build-inputs.json` with sha256 for all embedded inputs (CHANGELOG, migrations, bundled skills, models snapshot, parser worker, per-target native prebuild, `packages/*/src` including `.mp3`, each workspace `package.json`, workspace-root `package.json`, `bun.lock`, and `packages/opencode/tsconfig.json`). Paths are repo-root-relative and the stamp records walked roots and `packages/*/src` globs. The smoke test rehashes these and re-enumerates the roots to catch newly added files; falls back to the old mtime walk if no stamp is present. Shared walker lives in `packages/opencode/script/stamp-inputs.ts`.
- Models cache: drop the eager import-time refresh; keep the hourly `.unref()` refresh. `ModelsDev.Data()` parses and validates before caching, returns empty on non-2xx or invalid JSON, and rejects non-catalog objects; the disk cache is validated on read to avoid poisoning.
- Tests/docs: the CLI retry scrubs `opencode*.db{,-wal,-shm}` before a single retry and rethrows non-ENOENT errors; a deterministic unit test replaces the flaky PTY-based phase label test; end-to-end tests cover scanner behavior; `CONTRIBUTING.md` notes the pre-push scan and bypass.

<sup>Written for commit cafdcf4500d90db567336c864aa7d9fe9662bcc3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1085?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added automated checks to detect sensitive tracker references before code is pushed.
  * Added build integrity checks to identify stale or changed build inputs.

* **Bug Fixes**
  * Improved handling of unavailable, invalid, or incomplete model catalogs.
  * Improved retry behavior when local database files are locked.
  * Reduced unnecessary model-catalog refresh activity during startup.

* **Documentation**
  * Documented pre-push validation checks and the optional bypass.

* **Tests**
  * Added coverage for tracker detection, build freshness, database retries, and startup phase labels.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -19,6 +19,6 @@ if (process.versions.bun !== expectedBunVersion) {
 '
 bun typecheck
 # altimate_change — #1052 D8: scan pushed content for internal-tracker refs
-# (Jira keys, altimateai.atlassian.net). Silent on clean; exits 1 on hit.
-# Bypass with SKIP_TRACKER_CHECK=1 for genuine emergencies.
+# (see RULES in the script for the specific patterns). Silent on clean;
+# exits 1 on hit. Bypass with SKIP_TRACKER_CHECK=1 for genuine emergencies.
 bun script/check-tracker-leaks.ts
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -18,3 +18,7 @@ if (process.versions.bun !== expectedBunVersion) {
 }
 '
 bun typecheck
+# altimate_change — #1052 D8: scan pushed content for internal-tracker refs
+# (Jira keys, altimateai.atlassian.net). Silent on clean; exits 1 on hit.
+# Bypass with SKIP_TRACKER_CHECK=1 for genuine emergencies.
+bun script/check-tracker-leaks.ts
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -39,6 +39,8 @@ https://github.com/anomalyco/models.dev
   bun dev
   ```
 
+`bun install` sets up the pre-push hook via `husky` — subsequent `git push` runs a bun-version check, `bun typecheck`, and a scan for internal-tracker references. If the tracker scan blocks you legitimately (extremely unlikely on this public repo), bypass with `SKIP_TRACKER_CHECK=1 git push`.
+
 ### Running against a different directory
 
 By default, `bun dev` runs Altimate Code in the `packages/opencode` directory. To run it against a different directory or repository:
--- a/packages/opencode/script/build.ts
+++ b/packages/opencode/script/build.ts
@@ -541,10 +541,16 @@ for (const item of targets) {
   // are relative to the workspace root (dir = packages/opencode) so the test
   // can resolve them from its own cwd without env plumbing.
   const stampInputs: Array<{ path: string; sha256: string }> = []
+  // altimate_change — #1052 D10 review-fix (M2): resolve inputs relative to
+  // REPO_ROOT (not `dir` = packages/opencode). Workspace-package files live
+  // OUTSIDE packages/opencode; a `dir`-relative path for those would render as
+  // `../tui/src/...` which the smoke-test reader would then have to un-prefix.
+  // REPO_ROOT-relative keeps paths portable and the reader trivial.
+  const _stampRoot = path.resolve(dir, "../..") // repo root
   const addFile = (absPath: string) => {
     try {
       const buf = fs.readFileSync(absPath)
-      const rel = path.relative(dir, absPath)
+      const rel = path.relative(_stampRoot, absPath)
       const hash = createHash("sha256").update(buf).digest("hex")
       stampInputs.push({ path: rel, sha256: hash })
     } catch {
@@ -564,6 +570,12 @@ for (const item of targets) {
   addFile(parserWorker)
   // Per-target altimate-core NAPI prebuild
   addFile(platformNodeSrc)
+  // altimate_change — #1052 D10 review-fix (M2): package.json + bun.lock cover
+  // dependency-version bumps that change what Bun.build embeds. Without these,
+  // `bun install` bumping a bundled dep would leave the stamp reporting fresh.
+  const REPO_ROOT = path.resolve(dir, "../..")
+  addFile(path.join(REPO_ROOT, "package.json"))
+  addFile(path.join(REPO_ROOT, "bun.lock"))
   // src/ + script/ TypeScript tree — hash every file the compiler actually saw
   // (same extension filter build.ts globs for embedding).
   const IGNORED = new Set(["node_modules", ".turbo", ".cache", "dist", "target"])
@@ -588,6 +600,30 @@ for (const item of targets) {
   }
   walk(path.join(dir, "src"))
   walk(path.join(dir, "script"))
+  // altimate_change — #1052 D10 review-fix (M2): also hash every workspace
+  // package's src/ tree. `packages/opencode/src` imports from
+  // `@opencode-ai/{core,tui,util,plugin,sdk,server,cli,...}` and
+  // `@altimateai/{dbt-tools,drivers}` — Bun.build follows these imports and
+  // bundles them into the binary transitively. The original stamp walked only
+  // packages/opencode, so edits under any sibling workspace package would leave
+  // the binary silently stale. Enumerate `packages/*/src` at build time (rather
+  // than hard-coding names) so new packages get covered automatically.
+  const packagesRoot = path.resolve(REPO_ROOT, "packages")
+  try {
+    for (const pkg of fs.readdirSync(packagesRoot, { withFileTypes: true })) {
+      if (!pkg.isDirectory() || pkg.name.startsWith(".")) continue
+      // Skip packages/opencode — already covered by the walks above.
+      if (pkg.name === "opencode") continue
+      const pkgSrc = path.join(packagesRoot, pkg.name, "src")
+      if (fs.existsSync(pkgSrc)) walk(pkgSrc)
+      // Each workspace package.json influences its resolution/exports and could
+      // change what ends up in the binary even when its src/ files are unchanged.
+      const pkgJson = path.join(packagesRoot, pkg.name, "package.json")
+      if (fs.existsSync(pkgJson)) addFile(pkgJson)
+    }
+  } catch {
+    // packages/ missing (unlikely at build time) — skip; addFile() ignores non-existent paths anyway.
+  }
   // Deterministic order so the aggregate hash is stable across build runs.
   stampInputs.sort((a, b) => a.path.localeCompare(b.path))
   const aggregate = createHash("sha256")
--- a/packages/opencode/script/build.ts
+++ b/packages/opencode/script/build.ts
@@ -649,6 +649,9 @@ for (const item of targets) {
         // Roots the walk covered, so the read side can detect files added after
         // the build rather than only rehashing what was present at build time.
         roots: [...new Set(walkedRoots)].sort(),
+        // Glob form so the read side notices a package added AFTER this build;
+        // concrete roots only describe what existed while it ran.
+        rootGlobs: ["packages/*/src"],
         inputs: stampInputs,
       },
       null,
--- a/packages/opencode/script/build.ts
+++ b/packages/opencode/script/build.ts
@@ -17,6 +17,7 @@ process.chdir(dir)
 
 import { Script } from "@opencode-ai/script"
 import pkg from "../package.json"
+import { walkInputs } from "./stamp-inputs"
 
 // Python engine has been eliminated — all methods run natively in TypeScript.
 // ALTIMATE_ENGINE_VERSION is no longer needed at runtime.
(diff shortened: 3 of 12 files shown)


sahrizvi:
## Bot-review round — response

Went through the CodeRabbit, cubic-ai, and kilo-code bot findings. Applied fixes in commit `5b30a30561`:

### Addressed

**Scanner (`script/check-tracker-leaks.ts`)**
- **Cubic P1 — shell injection via `--base`**: replaced `sh -c "${cmd}"` with a shell-free `git(args[])` helper using Bun.$ tagged templates (each arg becomes one argv element). Added a defense-in-depth regex validator for `--base` — rejects anything that isn't git-ref-shaped. Verified: `--base='$(rm -rf /tmp/xxx)'` now exits 2 with a clear message.
- **Cubic P1 — silent-fail on git error**: old `shOK` swallowed all git failures. New `git()` fails loud (exit 2) except for `merge-base` (which legitimately returns empty on diverged history).
- **CodeRabbit Major — `!startsWith('+++')` drops content lines starting with `++`**: filter now matches file headers exactly (`+++ ` or `+++\t`) so content lines whose prefix happens to look like `+++<text>` still get scanned.

**Build stamp (`packages/opencode/script/build.ts`)**
- **Kilo — `_stampRoot` / `REPO_ROOT` duplicated**: consolidated to one `REPO_ROOT`.
- **Kilo — stale comment about paths being `packages/opencode`-relative**: rewritten (they're REPO_ROOT-relative post-M2).
- **Cubic P2 — tsconfig.json affects compiled output**: added `packages/opencode/tsconfig.json` to the stamp.

**models.dev cache (`packages/opencode/src/provider/models.ts`)**
- **CodeRabbit Minor — 2xx with HTML / truncated JSON poisons cache**: parse first, cache + return only on success. On parse failure log with a body preview and return an empty catalog (same graceful path as non-2xx).

### Not addressed (with reasoning)

- **Cubic P1 — cold-cache CLI starts `fetchApi` before exit, recreates D14 blocker** *(and kilo warning on same line)*: this is the trade-off the D14 review-fix commit body explicitly documents. If CI Verdaccio sanity Phase 3 shows the hang, the fix is the one-line delete already noted. No change here.
- **Cubic P1 — hook doesn't wire the scanner**: false positive. `.husky/pre-push` line 24 explicitly invokes `bun script/check-tracker-leaks.ts`.
- **Cubic P1 — hook doesn't pass pushed refs**: real, but low-priority — git pre-push runs when the CURRENT branch is being pushed 99% of the time; my scanner already compares HEAD to origin/main. Cross-branch push (`git push origin some-other-branch`) is an edge case; deferred.
- **CodeRabbit + cubic — new files added post-build not detected in stamp**: real limitation, orthogonal fix. Adding a walk at read-time would double the cost of every smoke-test run for a rare regression class. Deferred.
- **CodeRabbit — `as BuildStamp` doesn't runtime-validate**: test-only surface; adding zod for a stamp file with fixed shape is over-engineering. Deferred.

CodeRabbit's original "CRITICAL: duplicate `stampInputs`" was a false read — there's only one declaration. The real duplicate (`_stampRoot` + `REPO_ROOT`) that kilo caught IS fixed above.

Verified: typecheck 13/13, scanner self-test 28/28, scanner runs clean on this branch, security bypass rejected with exit 2.

saravmajestic (approved):
LG

PR #1096, opened 2026-08-13
feat: [AI-8448] count installs from the shell installers, not just npm

saravmajestic:
Fixes [AI-8448](https://altimateai.atlassian.net/browse/AI-8448).

## The dip was measurement, not installs

`first_launch` is the only install metric. It fires off a marker file rather than any network call from the installer — and that marker was written in exactly one place, `packages/opencode/script/postinstall.mjs`. Neither `install` nor `install.ps1` wrote it, so once the advertised path moved from npm to `altimate.sh/install`, those installs stopped being counted. Nothing changed about how many people were installing.

## What this does

Both shell installers now write the same marker `postinstall.mjs` writes, and `first_launch` carries a new `install_method` (`curl` | `powershell` | `npm` | `unknown`) so the recovered volume is separable from npm instead of folded into one number. `altimate upgrade` on the curl path re-runs `install`, so curl upgrades become visible too.

Brand-new installs stay `is_upgrade: false` — that field probes whether `~/.altimate/machine-id` existed before this launch:

```
count(distinct machine_id) where type = "first_launch" and is_upgrade = false
```

Expect `install_method: "unknown"` for the first upgrade after this ships — those markers predate the field.

## Details that fail silently rather than loudly

- **Marker path.** `$XDG_DATA_HOME`, default `~/.local/share/altimate-code`, on *every* platform including Windows. `welcome.ts` resolves the data dir through Node's `os.homedir()` and never consults `%LOCALAPPDATA%`; a marker written there would be ignored at read time.
- **`-Encoding ascii` in `install.ps1`.** The documented entrypoint is `powershell -c "irm ... | iex"` — Windows PowerShell 5.1, where `-Encoding utf8` prepends a BOM. `.trim()` does strip a leading BOM (U+FEFF is JS whitespace), so this was latent rather than broken, but `install_method` is matched against a fixed allowlist and shouldn't depend on that.
- **`unknown` version fallback.** An empty marker is deleted unread, so an unresolved version would lose the install outright. That's the state `check_version` leaves whenever the GitHub API is unreachable.
- **Write happens after the install dispatch.** A version already present (`check_version` exits 0 early) reports no install, and neither does a failed download.
- **Allowlisted `install_method`.** A hand-edited or truncated source file reads `unknown` rather than minting a new dimension.
- **Source file is consumed on read**, including on the empty-marker path, so an orphan can't be attributed to a later install.
- **Non-fatal in both installers.** A read-only `$HOME` costs the event, never the install.

## Privacy

No new network call and no new identifier. The installers record a version and their own name to a local file; the CLI's existing opt-out gates (`ALTIMATE_TELEMETRY_DISABLED`, `OPENCODE_DISABLE_TELEMETRY`, `telemetry.disabled`) still decide whether anything is transmitted. `docs/docs/reference/security-faq.md` and `docs/docs/reference/telemetry.md` are updated to say so.

## Tests

Followed the touchpoint set from #1064 (event union → docs → emitter → unit tests → install-script assertions).

- `test/cli/welcome.test.ts` — `is_upgrade` both ways, `install_method` attribution, allowlist rejection, source-file consumption, empty-marker path.
- `test/install/install-telemetry.test.ts` — marker path/fallback/ordering/non-fatality for both installers, no-BOM, plus the ordering invariant below.
- `test/install/postinstall.test.ts` — npm writes `.install-source`.

The load-bearing test is the ordering invariant. `is_upgrade` is only correct because `src/index.ts` fires `Telemetry.init()` **unawaited** and `doInit()` yields at `await Config.get()` before minting the machine-id, so the synchronous banner call on the next line still sees pre-launch state. An `await` added ahead of that mint would make every install report `is_upgrade: true` and silently empty the brand-new-install metric without a single existing test failing. The test asserts the machine-id is absent at that instant *and* present once the promise resolves, so it can't pass vacuously.

598 pass / 0 fail across `test/cli/welcome.test.ts test/install/ test/telemetry/telemetry.test.ts test/branding/`; typecheck clean.

## Verification

`install`'s marker writer was executed directly: XDG override honored, `v` prefix stripped, `unknown` fallback, exit 0 on a read-only `$HOME`. `install.ps1` is asserted at source level only — no `pwsh` on the dev machine, so its runtime behavior rides on CI.

## Not in scope

Installs that never launch the CLI remain uncounted, so download→launch conversion is still unmeasurable. That needs a beacon from the install script itself — a new event plus opt-out handling in bash — and is deliberately deferred.

Two pre-existing things noticed but left alone:

1. `welcome.ts:70` returns before printing the welcome box when `isUpgrade` is false, so the box only ever shows on *upgrades*. Plausibly intentional (the TUI has its own first-run flow), but it reads backwards for a "welcome" banner.
2. `test/altimate/review/telemetry.test.ts` redirects `$HOME` to keep the suite from minting a machine-id in the developer's real home — but Bun resolves `os.homedir()` at startup and ignores later `process.env.HOME` mutation, so that protection doesn't currently work. This PR's tests use `spyOn(os, "homedir")`, the convention already used in `test/mcp/discover.test.ts`.

[AI-8448]: https://altimateai.atlassian.net/browse/AI-8448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---
## Summary by cubic
Counts installs from the shell installers and the VS Code extension installer, attributing each install's source in telemetry. Previously only npm `postinstall.mjs` wrote the install marker, so installs via `altimate.sh/install`, `install.ps1`, and the VS Code extension went uncounted; now all installers write the same marker and `first_launch` carries `install_method` (`curl`/`powershell`/`npm`/`vscode-extension`/`local`, or `unknown`). Closes [AI-8448].

- Both shell installers and `postinstall.mjs` write `.install-source` before `.installed-version` and publish the version atomically (temp+rename), so a CLI reading mid-write can't misattribute or drop the install; `install --binary` reports `"local"` and both scripts fall back to `"unknown"` for an unresolvable version.
- `install.ps1` now computes its data dir inside the try: under `$ErrorActionPreference = "Stop"` a null `USERPROFILE` or bad PSDrive previously aborted the installer after the binary was placed but before the PATH write. The marker block became a `Write-InstallMarker` function so Pester can AST-extract and execute it, with the suite pinned to Stop semantics so the try/catch tests can't silently pass.
- `welcome.ts` reads and consumes `.install-source` (cleared in `finally`, and recursive so a directory-shaped file can't pin attribution to `unknown`), allowlists the five methods, and adds `install_method` to the `first_launch` event.
- `src/index.ts` runs the welcome banner before `Telemetry.init()` so `is_upgrade` is structurally correct instead of relying on init's async timing.
- Docs note the installers send no telemetry themselves and that the `telemetry.disabled` config key can be bypassed when telemetry startup runs before config resolves — env vars remain the guaranteed opt-out.
- Rollout: first upgrade after this ships may report `install_method: "unknown"` for pre-existing markers. Release the CLI before the VS Code extension starts writing markers so the reader exists first.

<sup>Written for commit b69530bf0ffe6a93a52adf89d3cd55ed355e69db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * First-launch telemetry now records installation methods including curl, PowerShell, npm, VS Code extension, local binary, or unknown.
  * Installers record version and source information locally for later telemetry transmission.
  * Telemetry respects opt-out settings, and installers do not transmit telemetry directly.

* **Documentation**
  * Clarified upgrade detection, delivery limits, installer sources, and fallback behavior.

* **Bug Fixes**
  * Improved handling of missing, invalid, or unwritable installation markers.
  * Marker updates are published safely to prevent incomplete telemetry records.

---

## Review fixes (human + 3 bots)

**One real bug.** `install.ps1` computed `$dataRoot`/`$dataDir` *above* the `try`. With `$ErrorActionPreference = "Stop"` and `Join-Path`'s provider-qualified path resolution, a null `$env:USERPROFILE` or an `XDG_DATA_HOME` naming a bad PSDrive raised a **terminating** error there — aborting the installer after the binary was placed but before the PATH registry write. Result: installed binary, not on PATH, directly contradicting the block's own "non-fatal" comment. Now everything is inside the `try`, using `[IO.Path]::Combine` to keep PSDrive resolution out of it. The old test passed because it only checked that `} catch {` appeared somewhere; it now pins that the assignments come *after* `try {`.

**One false claim of mine, corrected.** The test header said `install.ps1`'s runtime behaviour "is exercised by the Windows Installer (Pester) CI job". It is not — that job's subprocess tests deliberately stop the installer via `-Help`/unknown `-Version` so nothing downloads, and never reach the marker block. It also runs under `pwsh`, never `powershell.exe`, so **Windows PowerShell 5.1 — the documented entrypoint and the entire reason for `-Encoding ascii` — is still unexercised anywhere.** I wrote that claim from the job name passing, without reading the Pester suite.

To fix the underlying gap, the block is now a `Write-InstallMarker` function so Pester can AST-extract and execute it the way it already does `Test-Checksum`. Five new cases run it against a temp profile: byte-exact contents with no BOM, `v`-strip and `unknown` fallback, `USERPROFILE` fallback, no throw on empty `USERPROFILE`, no throw when the data dir can't be created. That last pair is the coverage that would have caught the bug above.

**Other fixes:**

- **Companion before trigger, all three writers.** `.installed-version` is the reader's trigger; writing it first let a CLI starting in between report `unknown`, and since writes truncate first, a reader could observe an *empty* version file and delete it unread — losing the install, not just its attribution.
- **`--binary` is attributed `local`, not `curl`** (that branch sets `specific_version="local"`, so it misreported both source and version). `write_install_marker` now takes the method as `$1` — which also removed an unbound `$marker_source` I'd introduced mid-edit, a `set -u` abort waiting to happen.
- **Test isolation:** the ordering test now snapshots/clears/restores `OPENCODE_DISABLE_TELEMETRY` too — `doInit()` returns before minting if either gate is set.
- **Docs:** the security FAQ no longer asserts unconditionally that the opt-out decides transmission. Env vars are always honoured; the `telemetry.disabled` **config key** can be bypassed when telemetry startup runs before config is resolvable, so that caveat is now stated with a pointer to use an env var for a guarantee.

## Still not fixed: the config-key opt-out itself

Flagged by CodeRabbit, cubic, and the human reviewer, and it stays open deliberately. The gate is shared by **every** event emitted from CLI middleware — this PR raises how often it's hit (~30x volume), it doesn't introduce it. Both available in-PR routes are wrong: failing closed emits nothing at all on the middleware path (killing the feature), and reading config here means duplicating the merge + JSONC semantics of `config/config.ts`. The fix belongs in telemetry init — make `Config` resolvable there, or adopt a module-wide fail-closed policy. Tracked in the expanded `FIXME`; needs its own ticket.

## Release order

`vscode-extension` has no producer in this repo — it ships in AltimateAI/vscode-altimate-mcp-server#453. **Release this CLI first**, so the reader exists before the extension starts writing `.install-source`; otherwise that file is orphaned by pre-#1096 readers and can be misattributed later. A zero `vscode-extension` share after release means the extension hasn't rolled out yet, not that there are no extension installs — now noted in `telemetry.md`.

Verification: typecheck clean, 1383 pass / 0 fail across `test/cli test/install test/telemetry test/branding`. The `install.ps1` Pester additions are unverified locally (no `pwsh` on this machine) and rely on the CI job.

---

## Round 3 (b69530bf0f) — both blocking items, plus four found pre-submit

**Blocking, from review:**
- `postinstall.mjs` write order — fixed. The previous commit message claimed "all three writers"; only two had been changed.
- Pester test that could not fail — fixed. `$ErrorActionPreference = "Stop"` in `BeforeAll`, a no-marker assertion proving the failure path was reached, and a third case asserting the preference itself so drift back to `Continue` fails visibly.

**Found before submitting:**
- **Atomic trigger publish in all four writers.** Companion-first ordering closes the "attribution lost" window but not truncation: a plain write truncates before filling, so a reader mid-write can observe an *empty* `.installed-version`, which `welcome.ts` deletes unread — losing the install itself. Only the extension published atomically. Now temp+rename in `install`, `install.ps1` and `postinstall.mjs` too.
- **`clearInstallSource` could not do what its comment claimed.** The comment said a directory in place of `.install-source` is still cleared; `fs.unlinkSync` throws `EPERM`/`EISDIR` on a directory, so it would have survived every launch and pinned `install_method` to `unknown` permanently. Now `rmSync` with `force`+`recursive`, with a test that creates a directory-shaped file and asserts removal.
- **Pester header no longer described the file** — read as subprocess-only after AST-extracted execution was added. Rewritten to document both layers.
- **Empty-`USERPROFILE` Pester case wrote into the git checkout.** `[IO.Path]::Combine("", ".local", "share")` returns the *relative* path `.local\share`, so the marker landed under the Pester process's cwd (checkout root in CI) while `AfterEach` only cleaned the sandbox. Now `Push-Location $script:Sandbox`.
- Two test-quality fixes: the bash "unwritable dir" case asserted only `exit 0` (passed with the writer deleted — bash twin of the Pester finding), and executed tests leaked temp dirs on assertion failure.
- Docs: `unknown` added to the FAQ's `install_method` list; `telemetry.md` now states both the boolean schema and the string KQL form of `is_upgrade` instead of one ambiguously.

**Verified:** typecheck clean; 1388 pass / 0 fail across `test/cli test/install test/telemetry test/branding`, two consecutive runs; bash marker writer executed directly for both install methods with no temp residue.

**Not verified locally:** the `install.ps1` and Pester changes. No `pwsh` on the dev machine (cask needs interactive sudo), so the Windows Installer (Pester) job is their only check — and a passing job confirms the tests pass, not that they can fail. The mutation test (delete the `try`/`catch`, confirm the case fails) has not been performed.


install.ps1
```diff
@@ -304,6 +304,40 @@ if (-not $needsBaseline) {
...
+#
+# The directory MUST match welcome.ts's resolution - $XDG_DATA_HOME, else
+# <home>\.local\share - because the CLI reads it through Node's os.homedir() and
+# never looks at %LOCALAPPDATA%. Writing to LOCALAPPDATA here would be silently
+# ignored at read time.
+#
+# No network call and no identifier is written; only the installed version is
+# recorded. The CLI's existing telemetry opt-out gates still decide whether
+# anything is ever sent.
+$dataRoot = if ($env:XDG_DATA_HOME) { $env:XDG_DATA_HOME } else { Join-Path $env:USERPROFILE ".local\share" }
```
sahrizvi: **MAJOR — the path computation that can abort the installer sits outside the `try`**

`$ErrorActionPreference = "Stop"` is set at `install.ps1:28`. `$dataRoot` and `$dataDir` are computed here at lines 321-322, *before* `try {` at line 323. `Join-Path` resolves provider-qualified paths, so a null or empty `$env:USERPROFILE` (pwsh on non-Windows, stripped service profiles) or an `XDG_DATA_HOME` naming a non-existent PSDrive raises a **terminating** error at these two lines.

The marker block spans 307-339; the PATH section begins at 341. A throw here aborts the installer *after* the binary is placed but *before*:

- the user-PATH registry write and `WM_SETTINGCHANGE` broadcast (353-372),
- the `$GITHUB_PATH` export (376-379),
- the "Get started" output (381-390).

The user ends up with an installed binary that is not on PATH, plus a red terminating error — the opposite of this block's own comment ("Non-fatal - a missing marker only costs us the install event") and of the PR description's "Non-fatal in both installers."

The test that names this invariant (`install-telemetry.test.ts:155-157`, "cannot abort the install") asserts only that `} catch {` appears somewhere in the block, so it passes with these assignments outside the `try`.

```powershell
try {
  $dataRoot = if ($env:XDG_DATA_HOME) { $env:XDG_DATA_HOME } else { [IO.Path]::Combine($env:USERPROFILE, ".local", "share") }
  $dataDir = [IO.Path]::Combine($dataRoot, "altimate-code")
  New-Item -ItemType Directory -Force -Path $dataDir | Out-Null
  ...
```

`[IO.Path]::Combine` additionally takes PSDrive resolution out of the equation.

Trigger probability is low — every real Windows user account has `USERPROFILE`. It is MAJOR because the block explicitly claims non-fatality, the blast radius is "installed but not on PATH", and the fix is two lines moved.

packages/opencode/test/install/install-telemetry.test.ts
```diff
@@ -0,0 +1,231 @@
...
+ * moving the advertised install path to altimate.sh/install took installs out of instrumentation
+ * and showed up as a dip in the dashboard.
+ *
+ * Coverage is in two layers. Source-level assertions (matching windows-install.test.ts) pin the
+ * details that fail silently — the data-dir expression, the "unknown" fallback, ordering relative
+ * to the install dispatch. Those cannot catch a syntax error or a bad variable scope, so the
+ * "executed" describe below runs the real bash installer through its `--binary` path (no network,
+ * no GitHub) against a throwaway HOME and asserts the files the CLI actually reads.
+ *
+ * install.ps1 has source-level coverage only — no pwsh on macOS/Linux runners. Its runtime
```
sahrizvi: **MAJOR — this coverage claim is not accurate, and it conceals a real defect**

> install.ps1 has source-level coverage only — no pwsh on macOS/Linux runners. Its runtime behaviour is exercised by the Windows Installer (Pester) CI job.

`test/windows/install.Tests.ps1` never reaches the marker block. Its own header states it deliberately stops the script early "via `-Help` or an unknown `-Version` so no 268 MB binary" is downloaded — both exit well before `install.ps1:307`. `Test-Checksum` is reached by AST extraction, not by running the script through to the marker.

The suite also invokes `pwsh`, never `powershell.exe`. Windows PowerShell 5.1 — the documented entrypoint (`powershell -c "irm ... | iex"`) and the entire justification for `-Encoding ascii` — is not exercised at all.

So the riskiest of the three writers has zero runtime coverage of the new code, and the `$dataRoot`/`$dataDir`-outside-the-`try` defect flagged separately on `install.ps1:321` is exactly the class of bug that source-level regex cannot see. The effect of this comment is to discourage anyone from adding the coverage that would have caught it.

**Suggested fix**: correct the comment to say source-level + AST-syntax only, with no runtime verification. Then add a Pester case that reaches the marker block the way `Test-Checksum` already is reached — AST-extract the block, execute it against a temp `$USERPROFILE` / `$XDG_DATA_HOME`, and assert byte-exact `.installed-version` / `.install-source` with no BOM, under both `pwsh` and `powershell.exe`.

packages/opencode/src/cli/welcome.ts
```diff
@@ -9,6 +9,52 @@ import { Telemetry } from "../altimate/telemetry"
...
 const MARKER_FILE = ".installed-version"
+// altimate_change start — written alongside MARKER_FILE by whichever installer ran
+// (postinstall.mjs, install, install.ps1) so first_launch can attribute the install.
+const SOURCE_FILE = ".install-source"
+// "vscode-extension" is the dominant installer by volume: the VS Code extension's
+// native installer pulls the binary straight from GitHub releases, bypassing npm and
+// both shell scripts (it stopped spawning `curl | bash` because EDR tooling flagged
+// it — vscode-dbt-power-user#2049). It writes the marker so those installs land here
+// rather than going uncounted.
+const INSTALL_METHODS = ["curl", "powershell", "npm", "vscode-extension"] as const
```
sahrizvi: **MAJOR — `vscode-extension` is allowlisted with no producer in this repo and no rollout note**

The value is allowlisted here, added to the event union (`telemetry/index.ts:479`), documented as a supported `install_method` (`security-faq.md:143`, `telemetry.md:41`), and described in the comment above as "the dominant installer by volume".

Nothing in this repository writes it. `grep -rn "vscode-extension"` returns only the allowlist, the union, the docs, and one test — no producer.

Delivery therefore depends on an out-of-repo change to the VS Code extension writing `$XDG_DATA_HOME/altimate-code/.install-source`. If that has not shipped, the dominant install source emits no `first_launch` at all — not even `unknown`, since it writes no `.installed-version` either — and the first dashboard read after this merges under-reports by whatever share the extension holds. That under-report is indistinguishable from a real dip, which is precisely the failure mode this PR exists to fix.

**Ask** (not a code change): link the extension-side change in the PR description, and add a line to `telemetry.md` naming the extension version from which `vscode-extension` starts appearing — the same way the PR already calls out the `unknown` transition for pre-field markers.

packages/opencode/src/cli/welcome.ts
```diff
@@ -27,32 +73,62 @@ function getDataDir(): string {
...
     // FIXME(telemetry-init-config-opt-out): doInit() may run before Instance.provide() has made
     // Config.get() resolvable (see the try/catch around Config.get in telemetry/index.ts::doInit
     // — the catch branch proceeds with telemetry enabled). A user who opted out via the
-    // `telemetry.disabled` config key — with no env var set — can therefore still get a
-    // machine-id minted on first launch. The env-var opt-out (ALTIMATE_TELEMETRY_DISABLED /
-    // OPENCODE_DISABLE_TELEMETRY) is unaffected — that check does not need Instance context.
-    // Pre-existing (not introduced by this release); calling it out explicitly here rather than
-    // leaving the earlier "(tracked separately)" wording, which claimed a tracking issue that
-    // does not currently exist.
+    // `telemetry.disabled` config key — with no env var set — can therefore still have early
```
sahrizvi, line 129: **MAJOR — the config-key opt-out is not honored for `first_launch`, and the FAQ says it is**

This FIXME is accurate and honest, and the fix genuinely belongs in telemetry init rather than here. Raising it anyway because of what the PR adds to the docs.

`doInit()` checks the env-var opt-outs *before* `await Config.get()`, but the `telemetry.disabled` config key is read inside `try { await Config.get() } catch { /* proceed with telemetry enabled */ }` (`telemetry/index.ts:1698-1709`). In the CLI middleware `Instance.provide()` has not run, so `Config.get()` throws, the catch proceeds with telemetry enabled, and a user who opted out via the config key alone still has `first_launch` transmitted.

The problem is the sentence added at `docs/docs/reference/security-faq.md:150`:

> ...so the opt-out above still decides whether anything is ever transmitted.

That states the guarantee unconditionally, for an event whose volume this PR grows by roughly the factor the comment above describes. The payload is low-sensitivity (version, boolean, coarse enum, random UUID), which is why this is MAJOR rather than CRITICAL — but the doc line should not assert a guarantee the code does not currently make.

**Suggested fix**: resolve the global opt-out through an API that needs no Instance context and fail closed when consent cannot be determined. PR-scoped minimum: qualify the FAQ sentence to name the config-key caveat.

sahrizvi:
## Consensus review — summary, minor findings, and rejected claims

**Verdict: request changes** — 0 critical · 4 major · 7 minor · 5 nit.

The design is sound and unusually well-reasoned. Data-dir parity across all three writers and the reader was verified independently several times over, the `is_upgrade` ordering fix is structural rather than timing-dependent, and the ordering test is genuinely non-vacuous. Tests pass (36/36 on the three touched files, 135/135 on telemetry) and typecheck is clean.

What holds it back is a cluster of claims — in comments, tests, and docs — stated more strongly than the code supports, and one of them conceals a real defect.

The four **major** findings are posted as inline comments:

| | Location | Finding |
|---|---|---|
| M1 | `install.ps1:321` | `$dataRoot`/`$dataDir` computed outside the `try`; under `$ErrorActionPreference = "Stop"` a throw there aborts the installer after the binary is placed but before PATH setup |
| M2 | `install-telemetry.test.ts:15` | The claim that the Pester CI job exercises `install.ps1`'s marker block at runtime is not accurate — it conceals M1 |
| M3 | `welcome.ts:20` | `vscode-extension` allowlisted with no producer in this repo and no rollout note |
| M4 | `welcome.ts:122` | The config-key opt-out is not honored for `first_launch`, and the FAQ line added by this PR asserts that it is |

---

## Minor

### m1 — `install:509-510`: `2>/dev/null` does not suppress a *redirection* failure

Bash sets up redirections left to right. If `> "$file"` fails, the error goes to the still-unredirected stderr before `2>/dev/null` takes effect:

```
$ bash -c 'printf "%s" hi > /tmp/ro/f 2>/dev/null || echo CAUGHT'
bash: /tmp/ro/f: Permission denied
CAUGHT
$ bash -c '{ printf "%s" hi > /tmp/ro/f; } 2>/dev/null || echo CAUGHT'
CAUGHT
```

`|| return 0` keeps it non-fatal, so this is noise on the advertised `curl | bash` path rather than breakage — but it contradicts "Details that fail silently rather than loudly". Line 508's `mkdir -p ... 2>/dev/null` is fine (external command, fd 2 redirected before exec); only the two `printf` redirects are affected.

```bash
{ printf '%s' "${marker_version#v}" > "$data_dir/.installed-version"; } 2>/dev/null || return 0
```

The executed test `"an unwritable data dir does not fail the install"` (`install-telemetry.test.ts:118-127`) blocks the *parent*, so `mkdir -p` fails at 508 and returns before either `printf` runs — the failing-write path is never executed. Worth a case where `$data_dir` exists at mode `0555`, asserting stderr is clean.

### m2 — `install:510`: `--binary` installs are attributed `install_method: "curl"`

`install:513` routes `--binary` to `install_from_binary`, which sets `specific_version="local"` (`install:77`); line 510 then writes `"curl"` unconditionally. `telemetry.md` documents the version side ("Local `--binary` installs report `version: "local"`") but not the attribution side.

It also weakens the executed tests: `describe("install — marker writer, executed")` runs *only* the `--binary` path and asserts `.install-source === "curl"`, locking the mis-attribution in as expected behaviour, while the real curl download path keeps source-level coverage only.

Documenting it alongside the existing `version: "local"` note is probably enough. A distinct `local` value would need a schema decision, not just a code change.

### m3 — `welcome.ts:78`: the missing-marker early return leaves an orphaned `.install-source`

Three exit shapes, two of which clear the companion file: marker absent → plain `return` at `:78`; marker empty → clears both (`:82`, `:87`); happy path → `readInstallMethod`'s `finally` clears it (`:54`). The asymmetry defeats the lockstep invariant claimed at `:42-44`. A crash between `:99` and `:143` orphans the source file, the next launch takes path (a) and leaves it there, and a later install whose installer writes only `.installed-version` reads the stale value.

```ts
if (!fs.existsSync(markerPath)) {
  clearInstallSource(dataDir)
  return
}
```

### m4 — `install:503`: Git Bash / MSYS `$HOME` can differ from `USERPROFILE`

`install:84` maps `MINGW*|MSYS*|CYGWIN*` to `os="windows"`, so the bash installer is a supported path there, and it writes `$HOME/.local/share/...`. The CLI resolves the data dir through Node's `os.homedir()`, which follows `USERPROFILE`. Under an MSYS2 profile with `db_home` configured these differ, so the marker lands where the CLI never reads it and the install goes uncounted — despite the comment at `install:495-497` claiming the path matches "on every platform, including Windows".

Either derive the fallback from `USERPROFILE` (via `cygpath` where available) on Windows bash targets, or drop the "every platform" claim.

### m5 — `install-telemetry.test.ts:66-72`: the non-fatality test counts substrings

```ts
expect(fn.match(/\|\| return 0/g)?.length).toBeGreaterThanOrEqual(3)
```

This asserts a token appears three times, not that any failure is survivable — it would pass with all three guards attached to the wrong commands. With the executed `describe` covering only the `mkdir` failure (see m1), the real guarantee rests on a single path.

### m6 — `install:522`: a network-less re-run reinstalls and emits `version: "unknown"`

`check_version:276` only exits early when `specific_version` is non-empty; when the GitHub API is unreachable it is empty (`:220-227`), so the version check is skipped, the install proceeds, and the marker is written as `unknown`. Combined with `altimate upgrade` re-running `install` on the curl path, a machine with flaky egress emits a `first_launch` with `version: "unknown"` on every attempt — indistinguishable from pre-field markers in version breakdowns.

### m7 — `security-faq.md:143`: omits `unknown` from the `install_method` enumeration

The FAQ lists `curl`, `powershell`, `npm`, `vscode-extension`; the schema (`telemetry/index.ts:479`) and `telemetry.md` both include `unknown`, which is a frequent real value — every upgrade from a pre-field version reports it.

---

## Nit

- `machine-id.ts:1-3` still lists `cli/welcome.ts` as a `getOrCreateMachineId` call site; after this PR `welcome.ts:132-133` only probes with `existsSync`.
- `welcome.ts:145` — `if (!isUpgrade) return` means the welcome box only ever prints on upgrades. Pre-existing and already acknowledged in the description; noted so it does not get lost.
- Marker writes are not atomic (no temp file + rename) in any of the three writers, and the two-file protocol has no commit record, so a crash between the two writes pairs a new version with a stale or absent source. Self-correcting and rare; one JSON record written via rename would close it.
- All three writers redundantly strip a leading `v` from a value already stripped upstream (`install:222`/`:231`, `install.ps1:167`/`:185`, npm `package.json` versions). Harmless, but it implies a `v` that never arrives.
- `welcome.ts:60-62` resolves the data dir by hand while `global/index.ts` uses the `xdg-basedir` package. They agree today; two independent resolutions of one path is a maintenance hazard.

---

## Raised and rejected

Recording these so they don't get re-raised.

**"Fresh PowerShell installs on AVX2 machines report `is_upgrade: true`."** The theory: `install.ps1:299`'s `& $InstalledBinary --version` probe runs the yargs middleware, minting `~/.altimate/machine-id` before the marker is written at `:321`. Refuted with a positive control against a real binary — isolated `HOME` and `XDG_DATA_HOME`, marker pre-seeded:

| invocation | `.installed-version` | `~/.altimate/machine-id` |
|---|---|---|
| `altimate --version` | present | absent |
| `altimate auth list` | **consumed** | **created** |

The control shows the middleware *does* run for a normal command and *does* consume the marker and mint the id — and that `--version` does neither, because yargs short-circuits it. A synthetic reproduction using `.exitProcess(false)` shows middleware running, but that changes the short-circuit path; the CLI uses the default. The ordering is also safe regardless: the probe at `:299` precedes the marker write at `:321`, so there is no marker to consume.

**"Concurrent CLI launches double-count `first_launch`."** Refuted. Both processes read the marker, then both call `fs.unlinkSync(markerPath)` at `welcome.ts:99`; the loser throws `ENOENT` into the function-level `catch` and returns before reaching `Telemetry.track` at `:137`. Exactly one event fires. The at-most-once comment at `:91-98` is accurate.

**"`getOrCreateMachineId` is synchronous, so the ordering rationale is wrong."** Refuted. The function is synchronous but is called at `telemetry/index.ts:1742`, after `await Config.get()` at `:1702`, so it does not run in `init()`'s synchronous prefix. The test's assertion is meaningful.

**"The allowlist check is case-sensitive."** Intentional. A case-varied value is a corrupt marker and `unknown` is the correct reading; lowercasing would accept malformed input.

**"The PowerShell marker fires on a skipped install."** Not reachable — the only skip path is `exit 0` at `install.ps1:213`.

---

## What's done well

- Data-dir parity is real across all three writers and the reader, including the empty-string-falsy behaviour of `${XDG_DATA_HOME:-...}` / `if ($env:XDG_DATA_HOME)` / `process.env.XDG_DATA_HOME ||`. The `%LOCALAPPDATA%` trap is explicitly warned against.
- The `is_upgrade` fix at `index.ts:132` converts a timing coincidence into a structural guarantee, and `Telemetry.track()` (`telemetry/index.ts:1774-1783`) genuinely buffers pre-init and clears the buffer if init resolves to disabled — nothing is lost, and the env-var opt-out stays intact.
- The ordering test is not vacuous: it asserts the machine-id is absent *and* then present after the await, and it clears `ALTIMATE_TELEMETRY_DISABLED` and sets a connection string so `doInit` cannot early-return into a tautological pass.
- The allowlist prevents a hand-edited marker in user-writable space from minting a free-form telemetry dimension.
- `-Encoding ascii` and its justification (PS 5.1 BOM) are correct, and the comment explaining why not `utf8` is the kind that survives a refactor.
- `telemetry.md` is unusually honest about the `is_upgrade` semantic limit, the `altimate uninstall` interaction, and at-most-once loss.

## Missing tests

- `install.ps1` marker block: any runtime execution, under both `pwsh` and `powershell.exe` (M1, M2).
- `install`: the failing-`printf` path with a writable parent and an unwritable target (m1).
- `install`: marker attribution on the real download path — currently source-level only (m2).
- `welcome.ts:78`: no test asserts `.install-source` is cleared on the missing-marker return (m3).
- `first_launch` with the `telemetry.disabled` config key set and a fake sink: assert no request and no machine-id (M4).
- App Insights serialization asserting `install_method` reaches the envelope; the telemetry test only asserts `track()` does not throw.

packages/opencode/script/postinstall.mjs
```diff
@@ -239,6 +239,9 @@ function writeUpgradeMarker(version) {
     const dataDir = path.join(xdgData, "altimate-code")
     fs.mkdirSync(dataDir, { recursive: true })
     fs.writeFileSync(path.join(dataDir, ".installed-version"), version.replace(/^v/, ""))
+    // Record the installer so first_launch can distinguish npm from the curl /
+    // PowerShell install scripts, which write the same marker.
+    fs.writeFileSync(path.join(dataDir, ".install-source"), "npm")
```
sahrizvi, line 247: 🔴 **BLOCKING**

**MAJOR — the write-order fix was not applied to npm, though the commit says it was**

The response commit states: *"Companion written before trigger, in all three writers."* Two were changed. npm still writes the trigger first:

```js
fs.writeFileSync(path.join(dataDir, ".installed-version"), version.replace(/^v/, ""))   // trigger
fs.writeFileSync(path.join(dataDir, ".install-source"), "npm")                          // companion
```

Both failure modes closed for `install` and `install.ps1` are still live here, and by the rationale given for the flip they matter:

- a CLI starting between the two writes sees `.installed-version` present and `.install-source` absent, so the npm install reports `install_method: "unknown"`;
- `fs.writeFileSync` truncates before writing, so a reader can observe an **empty** `.installed-version`, which `welcome.ts:83-91` deletes unread — losing the npm install outright.

npm is the only channel that was ever counted before this PR, so this is not a leftover on a dead path.

Both new order tests are per-writer source assertions (`install-telemetry.test.ts:76-89` for bash, `:210-218` for PowerShell). Nothing covers `postinstall.mjs`, and `postinstall.test.ts:107-114` asserts only that both files exist.

**Fix**: swap the two lines, and add the matching assertion to `postinstall.test.ts`.

test/windows/install.Tests.ps1
```diff
@@ -184,3 +184,94 @@ Describe "install.ps1 Test-Checksum" {
...
+    # The regression guard for path computation inside the try. With
+    # $ErrorActionPreference = "Stop", computing this outside the try raised a
+    # terminating error that aborted the installer AFTER the binary was placed but
+    # BEFORE the PATH write — leaving an installed binary that is not on PATH.
+    $env:XDG_DATA_HOME = ""
+    $env:USERPROFILE = ""
+    { Write-InstallMarker -Version "1.0.0" } | Should -Not -Throw
+  }
+
+  It "does not throw when the data dir cannot be created" {
```
sahrizvi: 🔴 **BLOCKING**

**MAJOR — this test cannot fail: the extracted function runs under different error semantics than the installer**

`install.ps1:28` sets `$ErrorActionPreference = "Stop"` at script scope, and the entire justification for `Write-InstallMarker`'s `try`/`catch` is that under `Stop` a cmdlet failure is terminating. The `BeforeAll` at `:196-205` dot-sources the function body out of the AST into a session where `$ErrorActionPreference` is pwsh's default `Continue`; nothing in `BeforeAll` or `BeforeEach` sets it.

Under `Continue`, a `New-Item` failure is non-terminating: it writes to the error stream, execution continues, the function returns, and `Should -Not -Throw` passes — **with or without the `try`/`catch`**. This case would pass with the `try`/`catch` deleted, so it cannot distinguish "protected by the guard" from "the error was never terminating anyway".

`"does not throw when USERPROFILE is empty"` (`:261-268`) survives, because `[IO.Path]::Combine($null, ...)` raises a .NET `ArgumentNullException`, which terminates under any preference. That one is a real regression guard for the path-outside-the-`try` bug.

**Fix**: one line in `BeforeAll` —

```powershell
$ErrorActionPreference = "Stop"
```

Then confirm both "does not throw" cases still pass, and that this one fails with the `try`/`catch` removed. Asserting no marker was created would also prove the fixture actually hit the intended failure path.

Extracting the function to make it AST-reachable was the right answer to the round-1 coverage finding. This is that same gap one level down: the suite reproduces the shape of the production path but not its semantics.

install.ps1
sahrizvi: **Resolved.** Verified at `df22213aac`: path computation is inside the `try`, `[IO.Path]::Combine` replaces `Join-Path` so PSDrive resolution can't throw at all, and the block is now a `Write-InstallMarker` function. The replacement test (`install-telemetry.test.ts:197-208`) pins `$dataRoot`/`$dataDir` after `try {` rather than grepping for `} catch {`, so it would catch a regression.

packages/opencode/test/install/install-telemetry.test.ts
sahrizvi: **Resolved, with one follow-up.** The comment now states the gap accurately instead of implying coverage that didn't exist, and the five AST-extracted Pester cases are real runtime coverage in a suite CI runs (`ci.yml:364-380`).

One case in that new suite can't fail, though — `install.Tests.ps1:270-276` runs under pwsh's default `$ErrorActionPreference = "Continue"` rather than the installer's `Stop`, so it passes with the `try`/`catch` deleted. Raised separately on that line.

packages/opencode/src/cli/welcome.ts
sahrizvi: **Resolved as scoped.** The `telemetry.md:41` note — that a zero `vscode-extension` share means the extension hasn't rolled out rather than no extension installs — is the part that mattered: the metric is no longer silently misreadable. Linking the extension-side change in the description would still help whoever reads the dashboard first, but that's an ask, not a blocker.

packages/opencode/src/cli/welcome.ts
sahrizvi: **Resolved as scoped.** The `security-faq.md:148-151` caveat states the config-key gap plainly and points at the env vars for a guarantee, and the unconditional wording is gone. Deferring the code fix to telemetry init is the right call for this PR.

One leftover from the same round: `security-faq.md:143` still omits `unknown` from the `install_method` list. `local` was added, but `unknown` is in both the schema (`telemetry/index.ts:481`) and `telemetry.md:41`, and it's what every upgrade from a pre-field version reports.

sahrizvi:
## Consensus re-review (round 2) — `df22213aac`

**Verdict: request changes** — 🔴 **2 blocking** (both major) · 🟡 **14 non-blocking** (8 minor, 6 nit).

Everything below is labelled. The two blocking items are posted inline; nothing else needs to hold the merge, though r3 and m7 are one-liners worth taking while you're in here.

All four round-1 majors are resolved: M1 and M2 in code, M3 and M4 by documentation with the underlying behaviour explicitly scoped out. Replies are on each of those threads.

The PowerShell work is substantive rather than cosmetic — `Write-InstallMarker` is extracted, AST-loaded and executed by five new Pester cases in a suite CI actually runs (`ci.yml:364-380`). M1's fix is the right shape: inside the `try`, and `[IO.Path]::Combine` removes PSDrive resolution entirely, with a test that pins position rather than grepping for `} catch {`.

### 🔴 Blocking — 2

Both posted inline:

| | Location | Finding |
|---|---|---|
| R1 | `postinstall.mjs:244` | The write-order flip was not applied to npm, though the commit says it was applied "in all three writers" |
| R2 | `install.Tests.ps1:270` | The AST-extracted function runs under `Continue`, not the installer's `Stop`, so the "cannot abort the install" case passes with the `try`/`catch` deleted |

### Round-1 ledger

| ID | Status at `df22213aac` |
|---|---|
| M1 `install.ps1` paths outside `try` | **Fixed** — `install.ps1:314-367`, order pinned at `install-telemetry.test.ts:197-208` |
| M2 false Pester coverage claim | **Fixed** — comment corrected, 5 executing cases added; see R2 |
| M3 `vscode-extension` rollout | **Fixed as scoped** — `telemetry.md:41` note |
| M4 config-key opt-out | **Fixed as scoped** — `security-faq.md:148-151` caveat; code path deferred |
| m2 `--binary` → `local` | **Fixed** — writer, allowlist, event type, tests, both docs |
| m7 FAQ `install_method` list | **Partially fixed** — `local` added, `unknown` still missing |
| m1, m3, m4, m5, m6, n1–n5 | **Still open**, unaddressed |

---

## 🟡 Non-blocking — minor (8)

None of these hold the merge. r3, m7 and m3 are one- or two-line changes; the rest are judgement calls or follow-ups.

### 🟡 r3 (new, non-blocking) — `install:475-483` + `:522`: a no-op `--binary` install still emits an install event

`install_from_binary` short-circuits when source and destination are the same file — it prints "nothing to do" and `return 0`. The caller at `:522` then writes the marker unconditionally, so re-running `install --binary ~/.altimate/bin/altimate` reports a `local` install that never happened.

The `else` branch is protected by `check_version`'s `exit 0`; this branch has no equivalent — which is also why the "Only reached when an install actually happened" comment now sits only on the `else`.

Have `install_from_binary` signal the no-op and skip the marker, or move `write_install_marker "local"` onto the successful-copy path. A same-file test asserting no marker would pin it.

### 🟡 m1 (still open, non-blocking) — `install:516-517`: `2>/dev/null` still does not suppress a redirection failure

Reproduced against the current function body — a marker directory that exists but is not writable prints to the user's terminal mid-`curl | bash`:

```
bash: line 5: /tmp/…/altimate-code/.install-source: Permission denied
RC=0
```

Still non-fatal, still contradicts "Details that fail silently rather than loudly". The order flip only changed which file names it.

```bash
{ printf '%s' "$marker_source" > "$data_dir/.install-source"; } 2>/dev/null || return 0
```

The one executed non-fatality test (`install-telemetry.test.ts:150-159`) blocks the *parent*, so `mkdir -p` returns early and neither `printf` runs — this path is still unexecuted.

### 🟡 m3 (still open, non-blocking) — `welcome.ts:81`: the missing-marker return leaves an orphan

`if (!fs.existsSync(markerPath)) return` still does not call `clearInstallSource(dataDir)`, while the empty-marker path at `:83-91` does.

The order flip makes the orphan a *designed* outcome rather than a crash-only one: with companion-first, any failure of the second write — `|| return 0` in bash, the `catch` in PowerShell — leaves `.install-source` with no trigger beside it, and `welcome.ts:81` returns past it forever.

```ts
if (!fs.existsSync(markerPath)) {
  clearInstallSource(dataDir)
  return
}
```

### 🟡 r4 (new, non-blocking design note) — the flip trades one window for another when an old trigger is unconsumed

Companion-first is strictly better only on a clean machine where both writes succeed. When a previous marker was never consumed — installed, never launched, then upgraded — the new writer overwrites `.install-source` first, so a CLI starting in that gap pairs the **old** version with the **new** installer's attribution. Trigger-first had the mirror-image flaw; neither ordering makes a two-file protocol atomic.

Not a reason to revert the flip, which is the better of the two orderings. Noting it because the real fix is one record written to a temp file and renamed:

```json
{"version":"1.2.3","install_method":"curl"}
```

Design note, not a change request for this PR.

### 🟡 m4 (still open, non-blocking) — `install:509`: Git Bash / MSYS `$HOME` can differ from `USERPROFILE`

`install:84` still maps `MINGW*|MSYS*|CYGWIN*` to `os="windows"` and `:509` still resolves `$HOME/.local/share`, while the CLI reads through `os.homedir()` (which follows `USERPROFILE`). The comment at `:501-503` still claims the path matches "on every platform, including Windows".

### 🟡 m5 (still open, non-blocking) — `install-telemetry.test.ts:150-159`: still counts `|| return 0` substrings

A token count, not a survivability proof. Notable because the fix elsewhere in this commit was the opposite move — the `} catch {` substring assertion was replaced with an order-pinning one. This is the same shape, left in place.

### 🟡 m6 (still open, non-blocking) — `install:530`: a network-less re-run reinstalls and emits `version: "unknown"`

`check_version` only short-circuits when `specific_version` is non-empty, so an unreachable GitHub API means reinstall-and-record-`unknown` on every attempt.

### 🟡 m7 (partially fixed, non-blocking) — `security-faq.md:143`: `unknown` still missing from the list

`local` was added; the list now reads `curl`, `powershell`, `npm`, `vscode-extension`, `local`. The schema (`telemetry/index.ts:481`) and `telemetry.md:41` both include `unknown`, and it is what every upgrade from a pre-field version reports. One word.

## 🟡 Non-blocking — nit (6)

- **n1–n5** from round 1 all still open and all still trivial (`machine-id.ts` call-site list; welcome box only on upgrades; non-atomic marker writes; redundant `v`-strip; two data-dir resolutions).
- **n6 (new)** — `install:512-522`: the "Only reached when an install actually happened" comment now sits on the `else` branch only, and per r3 is not actually true of the `--binary` branch.

---

## What improved

- M1's fix is the right shape, not a minimal patch: `[IO.Path]::Combine` removes PSDrive resolution from the path entirely, and the replacement test pins position rather than presence.
- Extracting `Write-InstallMarker` to make it AST-reachable is the correct answer to the round-1 coverage finding rather than a comment-only fix, and the byte-level BOM assertion (`install.Tests.ps1:229-236`) checks actual bytes rather than source text.
- The corrected header comment (`install-telemetry.test.ts:15-21`) now names the pwsh-vs-5.1 gap explicitly instead of implying coverage that does not exist.
- `write_install_marker` taking the method as `$1` removes the unbound-variable hazard, and the new test asserts the binding rather than just the literal.
- Snapshotting `OPENCODE_DISABLE_TELEMETRY` alongside `ALTIMATE_TELEMETRY_DISABLED` in the ordering test (`:255-295`) closes a real vacuous-pass route on a runner that exports the second gate.
- The `security-faq.md` caveat is honest about the config-key gap and points at the guarantee that does hold.

## Remaining test gaps

- 🔴 `postinstall.mjs` write order — no assertion anywhere (R1).
- 🔴 `$ErrorActionPreference = "Stop"` fidelity in the Pester suite (R2).
- 🟡 Same-file `--binary` asserting no marker (r3).
- 🟡 The failing-`printf` path in bash: writable parent, unwritable target (m1).
- 🟡 `.install-source` cleared on the missing-marker return (m3).
- 🟡 Windows PowerShell 5.1 anywhere — acknowledged in the header comment, still true.

---

Round-1 rejected claims (`--version` minting machine-id, concurrent-launch double-count, the synchronous `getOrCreateMachineId` argument, case-sensitive allowlist, marker on a skipped install) were re-checked at this head and remain rejected. They are not re-raised here.

packages/opencode/script/postinstall.mjs
saravmajestic: Fixed in b69530bf0f — and you were right that the commit message overclaimed.

`postinstall.mjs` now writes the companion first, and `postinstall.test.ts` gained the assertion you asked for. It checks real output (mtime ordering plus a `readdirSync` of the marker dir), with a source-index guard because mtimes can tie on fast filesystems.

While fixing it I found the reorder was only half the fix. Companion-first closes the "attribution lost" window; it does nothing about the truncation window you also named. A plain write truncates before filling, so a reader mid-write can still observe an **empty** `.installed-version`, which `welcome.ts` deletes unread — losing the install itself. Only the extension published atomically, so **all three CLI writers still had that window open**. That's the same partial-application pattern as the write-order fix, one level down.

All four writers now publish the trigger via temp+rename: `mv -f` in `install`, `Move-Item -Force` in `install.ps1`, `renameSync` in `postinstall.mjs`, matching the extension. Covered by a no-temp-residue assertion in the executed bash test and source assertions per writer.

test/windows/install.Tests.ps1
saravmajestic: Fixed in b69530bf0f. Correct on both counts — the dot-sourced function ran under pwsh's default `Continue`, so `Should -Not -Throw` passed with the `try`/`catch` deleted.

- `$ErrorActionPreference = "Stop"` set in `BeforeAll`
- the unwritable-dir case now also asserts no marker exists under the blocked root, proving the fixture reached the intended failure path
- added a third case asserting `$ErrorActionPreference` **is** `Stop` inside an `It`, so drift back to `Continue` fails visibly rather than silently disarming the suite

While in that file I also fixed something you'd have caught next: the empty-`USERPROFILE` case was writing into the git checkout. `[IO.Path]::Combine("", ".local", "share")` doesn't throw — it returns the **relative** path `.local\share`, so the marker landed under the Pester process's cwd (the checkout root in CI) while `AfterEach` only cleaned the sandbox. Now wrapped in `Push-Location $script:Sandbox`, with an assertion documenting that an empty profile resolves relative to cwd rather than `$HOME`.

And the file header, which still described the suite as subprocess-only after I added AST-extracted execution — the same stale-comment problem you flagged on the other test file. Rewritten to document both layers and the `Stop` requirement.

**One thing I could not do:** verify any of this locally. There's no `pwsh` on this machine and the Homebrew cask needs an interactive sudo password, so the Pester job is the only check — it passes, but that confirms the tests pass, not that they can fail. The mutation test you asked for (delete the `try`/`catch`, confirm this case fails) has **not** been performed. Flagging that rather than implying otherwise.

packages/opencode/test/install/install-telemetry.test.ts
saravmajestic: Fixed in b69530bf0f — `security-faq.md:143` now lists `unknown`, described as what every upgrade from a pre-field version reports. Schema, `telemetry.md` and the FAQ now agree on all six values (`curl`, `powershell`, `npm`, `vscode-extension`, `local`, `unknown`); cross-checked rather than eyeballed.

Also tightened `telemetry.md` while there: it wrote `is_upgrade != "true"` without saying which layer it meant. The schema types it boolean, but App Insights serializes `customDimensions` to strings, so the KQL form is `tostring(customDimensions.is_upgrade) != "true"`. Both are now stated explicitly.

sahrizvi:
## Consensus re-review (round 3) — `b69530bf0f`

**Verdict: 🟢 approve** — 🔴 **0 blocking** · 🟡 3 new minor/nit · 🟡 8 carried non-blocking.

Both round-2 blockers are closed, and B2 is verified by CI rather than by claim. All 13 checks are green.

### 🔴 → 🟢 Blockers, both closed

**B1 — npm write order.** `postinstall.mjs:247-256` writes the companion first, then publishes the trigger via `writeFileSync` to a PID-suffixed temp + `renameSync`, all still inside `writeUpgradeMarker`'s `try`/`catch`. The test asserts real output rather than source text — mtime ordering, an exact directory listing that also proves no temp survives, plus a source-order tiebreak for equal mtimes. That guard is sound: `".installed-version"` occurs exactly once in the file, in code, so `indexOf` can't match a comment.

**B2 — Pester error semantics.** `BeforeAll` now sets both `$ErrorActionPreference` and `$script:ErrorActionPreference` to `"Stop"`, the unwritable-dir case asserts no marker was created, and a third case asserts the preference itself so drift back to `Continue` fails visibly.

You flagged this as unverified locally. CI did the part that matters — job `98494401242`, windows-latest:

```
Describing install.ps1 Write-InstallMarker
  [+] does not throw when the data dir cannot be created, and writes no marker   16ms
  [+] runs under the installer's Stop semantics                                  10ms
Tests Passed: 15, Failed: 0
```

That third assertion is the load-bearing one — it proves `$script:ErrorActionPreference` set in `BeforeAll` actually reaches the `It` blocks under Pester 5's scoping, which was the one real doubt about this fix. The mutation test is still worth running when you have pwsh, but the preference assertion covers the failure mode it was meant to catch.

### Audit of the unrequested changes

The atomic publish is correct in all three in-repo writers: temp and target share a directory in every case, so the rename is same-filesystem by construction, and every rename sits inside the existing failure guard — `mv -f … || { rm -f "$tmp"; return 0; }` in bash, the function `try`/`catch` in PowerShell (`install.ps1:338`/`:365`), `writeUpgradeMarker`'s in npm (`postinstall.mjs:237`/`:255`). Non-fatality holds on every branch.

`clearInstallSource` → `rmSync({force, recursive})` is a strict improvement, and the new test creates a directory-shaped `.install-source` and asserts both `install_method === "unknown"` and removal — it would fail against `unlinkSync`.

The `Push-Location` fix is a real bug you caught unprompted: `[IO.Path]::Combine("", ".local", "share")` returns a *relative* path, so that case had been writing markers into the CI checkout root while `AfterEach` cleaned only the sandbox.

---

## 🟡 New, non-blocking

### N1 — `install.ps1:362`: fixed temp name, and no temp cleanup in the `catch` · MINOR

`.installed-version.tmp` carries no PID or GUID, where bash uses `.$$` and npm `.${process.pid}.tmp`. Two concurrent `install.ps1` runs share one temp path: A writes it, B truncates it, A renames a partially-written file into place — the exact failure the atomic publish exists to prevent, moved one file over. Concurrent `irm | iex` by one user is the only way to reach it, and the `try`/`catch` keeps it non-fatal.

There's also no cleanup of the temp on a `Move-Item` failure — the `catch` swallows it and `.installed-version.tmp` stays behind, where bash explicitly does `rm -f "$tmp"`. And the "no temp left behind" assertion exists for bash (`install-telemetry.test.ts:191-201`) and npm (`postinstall.test.ts:145`) but not for PowerShell, where only a source-level `-match` is checked.

```powershell
$tmpMarker = [IO.Path]::Combine($dataDir, ".installed-version.$PID.tmp")
# ... and in the catch:
Remove-Item -Force -ErrorAction SilentlyContinue $tmpMarker
```

### N2 — `install:521`: the temp is not cleaned when the temp write itself fails · NIT

`printf '%s' … > "$tmp" 2>/dev/null || return 0` returns before the `rm -f`, so a failure that still creates the file (disk full, quota) leaves a stale `.installed-version.$$`. One dotfile, and the very next line already has the cleanup idiom. Same gap in npm and PowerShell, where nothing is cleaned on either write or rename failure.

### N3 — `security-faq.md:143`: the new `unknown` gloss overstates when it occurs · NIT

"what every upgrade from a version predating this field reports" isn't right — an upgrade performed by a current curl/PowerShell/npm writer records its current method. `unknown` applies when the companion is missing, unreadable, or unrecognized, including an *unconsumed* marker written before the field existed. m7 itself is closed; only the explanation needs narrowing.

## 🟡 Carried, non-blocking

**m7 is now closed.** Still open and unchanged: **r3** (no-op `--binary` still writes a marker), **m1** (`2>/dev/null` doesn't suppress redirection-setup failure; the temp write adds a second occurrence), **m3** (missing-marker early return leaves an orphaned companion — marginally more reachable now, and better fixed together with r4 than alone, since clearing on the early return races companion-first writers), **m4** (Git Bash / MSYS `$HOME`), **m5** (substring-counting non-fatality test), **m6** (network-less re-run records `version: "unknown"`), **r4** (the pair still isn't atomic; only the trigger is), **n1–n6**.

Several comments now drift from the code: `welcome.ts:49` still says "unlink", both installers still say they record "only" the version, and the CI Windows-job comment (`ci.yml:355`) still describes subprocess coverage only.

---

## Raised and rejected this round

**"`rmSync({force, recursive})` is an unbounded recursive delete."** The path is always `$XDG_DATA_HOME/altimate-code/.install-source` — the app's own data directory, never derived from user input. `rmSync` with `recursive` does **not** follow a symlink; verified directly:

```
ln -s $T/precious $T/data/.install-source
fs.rmSync(t+"/data/.install-source",{force:true,recursive:true})
→ symlink entry exists: false · target dir survived: true · target file survived: true
```

No escape from the data dir. The `rmdirSync` alternative that was suggested is strictly worse: it throws `ENOTEMPTY` on a non-empty directory-shaped marker, the `catch` swallows it, and the marker survives forever — precisely the bug that moving off `unlinkSync` fixed.

**"PID collision in the bash temp name is critical."** A collision needs two live processes sharing a PID, which can't happen. Sequential reuse is harmless — the earlier temp is gone, and the later `printf` overwrites it regardless.

**"`install.ps1` and `postinstall.mjs` don't catch the rename exception."** Both do: `install.ps1:338-365` and `postinstall.mjs:237-255` wrap the rename in the same `try`/`catch` that makes each writer non-fatal.

**"Cross-filesystem rename."** Temp and target share a directory in all three writers, so they're on the same filesystem by construction.

Earlier rounds' rejected claims were re-checked at this head and remain rejected.

---

## CI

All 13 checks green at `b69530bf0f` — TypeScript, Windows Installer (Pester), CodeRabbit, cubic, Kilo Code Review, GitGuardian, and the rest.

Nothing here blocks. N1 is the one worth taking before merge if you're touching the file anyway; N2, N3 and the carried items are fine as follow-ups.

install.ps1
```diff
@@ -304,6 +304,71 @@ if (-not $needsBaseline) {
...
+    # Companion first, trigger last. The CLI returns early unless .installed-version
+    # exists, then consumes .install-source - so writing the trigger first would let a
+    # CLI starting in between report install_method "unknown". Set-Content also
+    # truncates before writing, and an empty .installed-version is deleted unread,
+    # which would lose the install outright.
+    Set-Content -Path ([IO.Path]::Combine($dataDir, ".install-source")) -Value "powershell" -NoNewline -Encoding ascii
+    # Trigger published atomically: Set-Content truncates before writing, so a CLI
+    # starting mid-write could observe an EMPTY .installed-version and delete it
+    # unread, losing the install. Move-Item within one directory is atomic.
+    $tmpMarker = [IO.Path]::Combine($dataDir, ".installed-version.tmp")
```
sahrizvi, line 362: 🟡 **Non-blocking (MINOR)** — fixed temp name, and no temp cleanup in the `catch`

`.installed-version.tmp` carries no PID or GUID, where bash uses `.$$` (`install:521`) and npm uses `.${process.pid}.tmp` (`postinstall.mjs:252`). Two concurrent `install.ps1` runs share one temp path: A writes it, B truncates it, A renames a partially-written file into place — the exact failure the atomic publish exists to prevent, moved one file over.

Concurrent `irm | iex` by the same user is the only way to reach it and the `try`/`catch` keeps it non-fatal, so this is not a blocker. It is the one place where this writer is weaker than its two siblings.

There is also no cleanup of the temp when `Move-Item` fails — the `catch` swallows it and `.installed-version.tmp` stays behind, where bash explicitly does `rm -f "$tmp"`. And the executed "no temp left behind" assertion exists for bash (`install-telemetry.test.ts:191-201`) and npm (`postinstall.test.ts:145`) but not here, where only a source-level `-match` is checked.

```powershell
$tmpMarker = [IO.Path]::Combine($dataDir, ".installed-version.$PID.tmp")
```

and in the `catch`:

```powershell
Remove-Item -Force -ErrorAction SilentlyContinue $tmpMarker
```

A `Get-ChildItem` assertion in the Pester "writes both marker files" case would close the test-coverage half.

install
```diff
@@ -487,11 +487,55 @@ install_from_binary() {
...
+    # exists, then consumes .install-source. Trigger-first would let a CLI starting in
+    # between report install_method "unknown", and a truncated .installed-version is
+    # deleted unread — losing the install rather than just its attribution.
+    printf '%s' "$marker_source" > "$data_dir/.install-source" 2>/dev/null || return 0
+    # The trigger is published atomically. Companion-first alone only closes the
+    # "attribution lost" window; a plain redirect truncates before filling, so a CLI
+    # starting mid-write can still observe an EMPTY .installed-version, which it
+    # deletes unread — losing the install itself. mv within one directory is atomic.
+    local tmp="$data_dir/.installed-version.$$"
+    printf '%s' "${marker_version#v}" > "$tmp" 2>/dev/null || return 0
```
sahrizvi, line 522: 🟡 **Non-blocking (NIT)** — the temp is not cleaned when the temp write itself fails

`|| return 0` returns before the `rm -f` on the next line, so a failure that still creates the file (disk full, quota) leaves a stale `.installed-version.$$` behind. One inert dotfile in the data dir that nothing reads — but line 523 already has the cleanup idiom, so it is a small asymmetry rather than a design question.

The same gap exists in the other two writers, where nothing is cleaned on either write or rename failure. A single best-effort cleanup path covering both stages would cover all three.

docs/docs/reference/security-faq.md
```diff
@@ -140,10 +140,16 @@ A single `first_launch` event is sent containing only:
 
 - The installed version (e.g., "0.5.9")
 - Whether this is a fresh install or upgrade (boolean)
+- Which installer was used (`curl`, `powershell`, `npm`, `vscode-extension`, `local`, or `unknown` — what every upgrade from a version predating this field reports)
```
sahrizvi, line 143: 🟡 **Non-blocking (NIT)** — the new `unknown` gloss overstates when it occurs

"what every upgrade from a version predating this field reports" is not quite right. An upgrade performed by a current curl / PowerShell / npm writer records its current method — the writer is what decides, not the version being upgraded from.

`unknown` is what the reader falls back to when the companion is missing, unreadable, or unrecognised. The predating-the-field case reaches it only through an *unconsumed* marker written before the field existed.

Listing `unknown` closes m7; only the explanation needs narrowing. Something like: "reported when the installer wrote no source marker, or wrote one this version does not recognise."

sahrizvi (approved):
## ✅ Approved

Both blocking items from the last round are closed, and the second one is verified by CI rather than by claim. All 13 checks are green at `b69530bf0f`. Nothing outstanding blocks the merge.

**B1 — npm write order.** `postinstall.mjs:247-256` writes the companion first, then publishes the trigger through a PID-suffixed temp and `renameSync`, still inside `writeUpgradeMarker`'s `try`/`catch`. The test asserts real output rather than source text: mtime ordering, an exact directory listing that also proves no temp survives, and a source-order tiebreak for equal mtimes. That tiebreak is sound — `".installed-version"` occurs exactly once in the file, in code, so `indexOf` cannot match a comment.

**B2 — Pester error semantics.** You flagged this as unverified locally. The Windows job did the part that matters — `install.ps1 Write-InstallMarker`, 15/15, including `runs under the installer's Stop semantics`. That assertion is the load-bearing one: it proves `$script:ErrorActionPreference` set in `BeforeAll` really does reach the `It` blocks under Pester 5's scoping, which was the only genuine doubt about the fix. The mutation test is still worth doing when you have pwsh, but the preference assertion covers the failure mode it was meant to catch.

**On the unrequested changes.** The atomic publish is correct in all three in-repo writers — temp and target share a directory in every case, so the rename is same-filesystem by construction, and every rename sits inside the existing failure guard. Non-fatality holds on every branch. The `clearInstallSource` switch to `rmSync({force, recursive})` is a strict improvement, and the directory-shaped-marker test would fail against `unlinkSync`. The `Push-Location` fix is a real bug you found unprompted: `[IO.Path]::Combine("", ".local", "share")` returns a relative path, so that case had been writing markers into the CI checkout root.

Three review comments below, all non-blocking. **N1** (the PowerShell fixed temp name and the missing cleanup in its `catch`) is the one worth taking if you touch the file again — it is the only place where the atomic publish is weaker than its bash and npm siblings. N2 and N3 are nits.

Eight carried non-blocking items from the previous round remain open and unchanged (r3, m1, m3, m4, m5, m6, r4, n1–n6); m7 is now closed. They are itemised in the round-3 comment above. Several code comments have also drifted — `welcome.ts:49` still says "unlink", both installers still say they record "only" the version, and `ci.yml:355` still describes subprocess coverage only.

One argument raised against this round and dropped after verification, recorded so it does not resurface: `rmSync({force, recursive})` was called an unbounded recursive delete. The path is always inside the app's own data directory and never derived from user input, and `rmSync` does not follow a symlinked marker — verified directly, the link is removed and the target survives untouched. The `rmdirSync` alternative suggested in its place would be strictly worse, throwing `ENOTEMPTY` on a non-empty directory-shaped marker and leaving it in place forever, which is exactly the bug moving off `unlinkSync` fixed.

Nice work on the round — particularly correcting your own earlier claim rather than letting it stand, and finding the checkout-littering bug without being asked.

PR #1156, opened 2026-08-26
feat(workspace): route warehouse tools through the bound workspace's engine

### LINKED ISSUE #1155: Warehouse tools and workspace engine tools both serve the same capability, with no arbitration
When a bound workspace's integration engine attaches, the model is offered two implementations of the same capability with different credential sources: the native warehouse tools (`sql_execute`, `sql_explain`, `schema_inspect`, `warehouse_list`) over local keychain connections, and the engine's MCP tools over the workspace's SaaS connection.

Nothing arbitrates between them. The model picks by description, and that choice silently decides which credentials run the query and whether the call is audited — engine calls are audited server-side, native calls are not.

Two properties make this hard to get right:

- **The engine's warehouse integrations are not symmetric.** Snowflake serves execute/explain/inspect; BigQuery and PostgreSQL serve execute only; Databricks serves execute only. Anything that arbitrates per *warehouse type* rather than per *capability* would redirect `sql_explain` on a BigQuery connection to a tool that does not exist.
- **A running engine is not necessarily serving the bound workspace.** An IDE writes its engine entry unpinned, and such an engine serves whichever teammate is active in that IDE, changing at runtime.

Expected: the engine wins for the capabilities it actually serves on a given type, local connections keep serving everything else, and anything undetermined runs locally with an explicit notice rather than silently.

ralphstodomingo:
### Issue for this PR

Closes #1155

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation

### What does this PR do?

Stacked on #1154 — review that first; this PR is the commits on top.

When a bound workspace's engine attaches, the model gets two ways to do the same thing: the native warehouse tools over local keychain connections, and the engine's MCP tools over the workspace's SaaS connection. Nothing chose between them, so the model picked whichever description read better — and that pick decided which credentials ran the query and whether it was audited. Engine calls are audited server-side; native ones aren't.

This adds a per-session decision: shadow only what materialised *and* is attributable to the bound workspace; anything undetermined runs locally and says why; nothing is silent. Why it works:

- **Materialised, not declared** — it reads the engine tool keys actually present, so a declared-but-absent integration shadows nothing.
- **Attributed** — an IDE writes its entry unpinned, and such an engine serves whichever teammate is active there, changing at runtime. #1154 makes attach guarantee the pin; this re-checks it and refuses to shadow otherwise, reusing that PR's parser rather than writing a second one.
- **Per capability, not per warehouse type** — the integrations aren't symmetric. Snowflake has execute, explain and table stats; BigQuery and Postgres have execute only. Keying on the type would send an explain on a BigQuery connection to a tool that doesn't exist, and there is deliberately no fallback.
- **Calls naming no warehouse** are judged against the target they'd really reach, mirroring each handler's own resolution — and the registry connection behind dbt is weighed too, since dbt falls back there whenever it yields nothing.
- **The redirect** names the exact engine tool, executes nothing, and is marked in metadata so telemetry can tell it from a real execution.
- **It runs after every native safety check** — the hard deny on destructive statements and the write confirmation both guard things the engine side has no equivalent for.
- `--integrations=local` turns it off for a session.

Two deliberate deviations: the guard needs a companion call to attach the fail-open notice, which a pre-execution check can't do; and I left an adjacent warehouse-type reporting bug alone, since fixing it changes a shipped telemetry field.

`bun run typecheck` and the workspace test suites pass; lint adds no new errors. 250 unit tests, including each decision rather than just the happy path.

End-to-end against a live workspace, engine serving 12 tools:

| Scenario | Outcome |
|---|---|
| Shadow marking | The served connection is marked per capability; local DuckDB rows unmarked |
| Redirect, nothing executed | Redirect returned, and **no matching row in query history** — proven with a positive control: the marker from a run that followed the redirect appears, the one that stopped at it does not |
| Control untouched | DuckDB connection runs normally |
| Default target | No warehouse named → resolved the default and redirected, naming it |
| Model behaviour | Unprompted, the model followed the redirect and ran the query through the engine tool |
| Write confirmation | An INSERT against the served connection stops at the permission gate, with no redirect |
| Escape hatch | `--integrations=local` → plain listing, local execution |

Not exercised: the asymmetry case needs a workspace serving only some capabilities (unit-tested instead); the dbt default path needs a bound dbt project; de-materialisation is held until other testing here finishes.

Automated review: four rounds plus one routed in, six findings, all real, all fixed — details in the [Codex review log](https://github.com/AltimateAI/altimate-code/pull/1156#issuecomment-5429804252).

GitGuardian flags a masked placeholder in a help-text snapshot that only moved columns; it is present unchanged on the base commit and is not a credential.

### Screenshots / recordings

n/a — CLI change, no UI.

### Residuals

Shadowing turns itself off and the call runs locally. All fail open, all in the attach
machinery rather than in this change:

- **Attach re-validation window** — no attribution while a session's memo re-validates.
  Being addressed in the attach layer.
- **Session-cache eviction** — beyond 256 tracked sessions the oldest is dropped and
  re-derived on the next call. Bounded and deliberate.
- **Cross-session teardown** — a redirect can name a client another session has removed.
  A tool-layer failure, not a query sent to the wrong warehouse.

### Checklist

### DIFF
--- a/packages/core/src/flag/flag.ts
+++ b/packages/core/src/flag/flag.ts
@@ -76,6 +76,14 @@ export const Flag = {
   get ALTIMATE_WORKSPACE() {
     return truthy("ALTIMATE_WORKSPACE")
   },
+  /**
+   * Workspace precedence escape hatch, set by `--integrations=local`. When on, the
+   * native warehouse tools serve every local connection themselves and nothing is
+   * redirected to the bound workspace's integration engine, for the whole session.
+   */
+  get ALTIMATE_INTEGRATIONS_LOCAL() {
+    return process.env["ALTIMATE_INTEGRATIONS"]?.toLowerCase() === "local"
+  },
   // altimate_change end
 
   // Evaluated at access time (not module load) because tests, the CLI, and
--- a/packages/opencode/src/altimate/native/connections/register.ts
+++ b/packages/opencode/src/altimate/native/connections/register.ts
@@ -98,7 +98,15 @@ async function ensureDbtAdapter(): Promise<any | null> {
 
 /** Where a `warehouse`-less call would actually go. */
 export type DefaultTarget =
-  | { source: "dbt"; type?: string }
+  | {
+      source: "dbt"
+      type?: string
+      /** Where execution actually lands if the dbt attempt yields nothing. `sql.execute`
+       * falls back to the registry not only when dbt is absent, but whenever
+       * `tryExecuteViaDbt` returns null — an unrecognised result shape, or any throw.
+       * A caller deciding anything about this call has to consider both targets. */
+      fallback?: { type: string; name: string }
+    }
   | { source: "registry"; type: string; name: string }
   | { source: "none" }
 
@@ -131,7 +139,9 @@ export async function resolveDefaultTarget(
         // Adapter not initialised far enough to answer; leave the type undetermined.
       }
       if (!type) type = await adapterTypeFromManifest()
-      return { source: "dbt", type }
+      const warehouses = Registry.list().warehouses
+      const fallback = warehouses.length > 0 ? { type: warehouses[0].type, name: warehouses[0].name } : undefined
+      return { source: "dbt", type, fallback }
     }
   }
 
--- a/packages/opencode/src/altimate/native/connections/register.ts
+++ b/packages/opencode/src/altimate/native/connections/register.ts
@@ -43,22 +43,23 @@ import { Telemetry } from "../../../telemetry"
 /** Cached dbt adapter (lazily created on first use). */
 let dbtAdapter: any | null | undefined = undefined
 
-/**
- * Try to execute SQL via dbt's adapter (which uses profiles.yml for connection).
- * Returns null if dbt is not available or not configured — caller should fall back
- * to native driver.
- *
- * This is the preferred path when working in a dbt project: dbt already knows
- * how to connect, so users don't need to configure a separate connection.
- */
-async function tryExecuteViaDbt(
-  sql: string,
-  limit?: number,
-): Promise<SqlExecuteResult | null> {
-  // Only attempt dbt once — if it's not configured, don't retry on every query
-  if (dbtAdapter === null) return null
+// altimate_change start — single-flight adapter creation.
+// Two concurrent `warehouse`-less calls used to construct the adapter twice, and
+// construction is expensive: it spawns a detached Python bridge, rebuilds the
+// manifest and starts file watchers. Share one in-flight promise instead. The
+// permanent negative cache (`dbtAdapter === null`) is unchanged — a project that
+// becomes valid mid-session is still not retried.
+let dbtAdapterInflight: Promise<any | null> | undefined
 
-  if (dbtAdapter === undefined) {
+/**
+ * Resolve the dbt adapter for this project, or null when there is no usable dbt
+ * project. Idempotent, single-flight, and permanently negative once it has failed.
+ */
+async function ensureDbtAdapter(): Promise<any | null> {
+  if (dbtAdapter !== undefined) return dbtAdapter
+  if (dbtAdapterInflight) return dbtAdapterInflight
+
+  dbtAdapterInflight = (async () => {
     try {
       // Check if dbt config exists
       const { read: readDbtConfig } = await import(
@@ -83,12 +84,96 @@ async function tryExecuteViaDbt(
       // Create the adapter
       const { create } = await import("../../../../../dbt-tools/src/adapter")
       dbtAdapter = await create(dbtConfig)
+      return dbtAdapter
     } catch {
       // dbt-tools not available or config invalid — fall back to native
       dbtAdapter = null
       return null
+    } finally {
+      dbtAdapterInflight = undefined
+    }
+  })()
+  return dbtAdapterInflight
+}
+
+/** Where a `warehouse`-less call would actually go. */
+export type DefaultTarget =
+  | { source: "dbt"; type?: string }
+  | { source: "registry"; type: string; name: string }
+  | { source: "none" }
+
+/**
+ * Resolve the target a call with no `warehouse` would reach, mirroring the resolution
+ * the handler for `op` performs itself — so a caller inspecting the target ahead of
+ * time cannot disagree with where execution actually lands.
+ *
+ * Only `sql.execute` consults dbt. `sql.explain` and `schema.inspect` are
+ * registry-only, and must stay that way: resolving them through dbt would drag
+ * adapter construction (Python bridge, manifest rebuild, file watchers) onto paths
+ * that never touch dbt today.
+ *
+ * For the dbt path the reported `type` is the project's adapter type, which is what
+ * decides *which* warehouse the profile reaches. It is left undefined when it cannot
+ * be established — the adapter coalesces an unknown type to the string "unknown", and
+ * the call can throw before initialisation completes.
+ */
+export async function resolveDefaultTarget(
+  op: "sql.execute" | "sql.explain" | "schema.inspect",
+): Promise<DefaultTarget> {
+  if (op === "sql.execute") {
+    const adapter = await ensureDbtAdapter()
+    if (adapter) {
+      let type: string | undefined
+      try {
+        const reported = adapter.getAdapterType?.()
+        if (typeof reported === "string" && reported && reported.toLowerCase() !== "unknown") type = reported
+      } catch {
+        // Adapter not initialised far enough to answer; leave the type undetermined.
+      }
+      if (!type) type = await adapterTypeFromManifest()
+      return { source: "dbt", type }
     }
   }
+
+  const warehouses = Registry.list().warehouses
+  if (warehouses.length === 0) return { source: "none" }
+  return { source: "registry", type: warehouses[0].type, name: warehouses[0].name }
+}
+
+/** Fallback adapter type: the dbt manifest records it as `metadata.adapter_type`. */
+async function adapterTypeFromManifest(): Promise<string | undefined> {
+  try {
+    const { read: readDbtConfig } = await import("../../../../../dbt-tools/src/config")
+    const dbtConfig = await readDbtConfig()
+    if (!dbtConfig) return undefined
+    const fs = await import("fs")
+    const path = await import("path")
+    const manifestPath = path.join(dbtConfig.projectRoot, "target", "manifest.json")
+    if (!fs.existsSync(manifestPath)) return undefined
+    const raw = JSON.parse(fs.readFileSync(manifestPath, "utf8"))
+    const adapter = String(raw?.metadata?.adapter_type ?? "").toLowerCase()
+    return adapter || undefined
+  } catch {
+    return undefined
+  }
+}
+// altimate_change end
+
+/**
+ * Try to execute SQL via dbt's adapter (which uses profiles.yml for connection).
+ * Returns null if dbt is not available or not configured — caller should fall back
+ * to native driver.
+ *
+ * This is the preferred path when working in a dbt project: dbt already knows
+ * how to connect, so users don't need to configure a separate connection.
+ */
+async function tryExecuteViaDbt(
+  sql: string,
+  limit?: number,
+): Promise<SqlExecuteResult | null> {
+  // altimate_change start — share the single-flight creation path with resolveDefaultTarget
+  if (!(await ensureDbtAdapter())) return null
+  // altimate_change end
 
   try {
     const raw = limit
@@ -146,6 +231,9 @@ async function tryExecuteViaDbt(
 /** Reset dbt adapter (for testing). */
 export function resetDbtAdapter(): void {
   dbtAdapter = undefined
+  // altimate_change — drop any in-flight creation too, or a test that resets mid-flight
+  // would still receive the previous adapter.
+  dbtAdapterInflight = undefined
 }
 
 // ---------------------------------------------------------------------------
--- a/packages/opencode/src/altimate/native/connections/register.ts
+++ b/packages/opencode/src/altimate/native/connections/register.ts
@@ -474,6 +474,11 @@ register("sql.execute", async (params: SqlExecuteParams): Promise<SqlExecuteResu
     // Reading once here makes the decided connection and the executed connection the
     // same by construction. The dbt-first ordering below is unchanged.
     const fallbackName = params.warehouse || Registry.list().warehouses[0]?.name
+    // Pinning the name is not enough on its own: the same name can be re-added against a
+    // different warehouse while this call is suspended, and the routing decision made for
+    // it is a function of that connection's canonical type. Pin the type too, so a
+    // replacement is caught rather than executed under a decision that never covered it.
+    const fallbackType = fallbackName ? Registry.canonicalType(Registry.getConfig(fallbackName)?.type) : undefined
     // altimate_change end
 
     // Strategy: try dbt adapter first (if in a dbt project), then fall back to native driver.
@@ -488,6 +493,13 @@ register("sql.execute", async (params: SqlExecuteParams): Promise<SqlExecuteResu
         "No warehouse configured. Use warehouse.add, set ALTIMATE_CODE_CONN_* env vars, or configure a dbt profile.",
       )
     }
+    // altimate_change start — refuse rather than execute under a stale decision.
+    if (Registry.canonicalType(Registry.getConfig(fallbackName)?.type) !== fallbackType) {
+      throw new Error(
+        `Connection "${fallbackName}" changed while this query was being prepared, so the routing decided for it no longer applies. Re-run the query.`,
+      )
+    }
+    // altimate_change end
     const connector = await Registry.get(fallbackName)
     const result: SqlExecuteResult = await connector.execute(params.sql, params.limit)
     try {
--- a/packages/opencode/src/altimate/native/connections/register.ts
+++ b/packages/opencode/src/altimate/native/connections/register.ts
@@ -467,6 +467,15 @@ register("sql.execute", async (params: SqlExecuteParams): Promise<SqlExecuteResu
   const startTime = Date.now()
   const warehouseType = getWarehouseType(params.warehouse)
   try {
+    // altimate_change start — resolve the fallback connection before the dbt attempt.
+    // `tryExecuteViaDbt` awaits, and the registry is mutable: re-reading it afterwards
+    // could pick a different connection than the one the caller's routing decision was
+    // computed against (a concurrent `warehouse.add` can change which name sorts first).
+    // Reading once here makes the decided connection and the executed connection the
+    // same by construction. The dbt-first ordering below is unchanged.
+    const fallbackName = params.warehouse || Registry.list().warehouses[0]?.name
+    // altimate_change end
+
     // Strategy: try dbt adapter first (if in a dbt project), then fall back to native driver.
     // dbt knows how to connect using profiles.yml — no separate connection config needed.
     if (!params.warehouse) {
@@ -474,22 +483,13 @@ register("sql.execute", async (params: SqlExecuteParams): Promise<SqlExecuteResu
       if (dbtResult) return dbtResult
     }
 
-    const warehouseName = params.warehouse
-    let result: SqlExecuteResult
-    if (!warehouseName) {
-      const warehouses = Registry.list().warehouses
-      if (warehouses.length === 0) {
-        throw new Error(
-          "No warehouse configured. Use warehouse.add, set ALTIMATE_CODE_CONN_* env vars, or configure a dbt profile.",
-        )
-      }
-      // Use the first warehouse as default
-      const connector = await Registry.get(warehouses[0].name)
-      result = await connector.execute(params.sql, params.limit)
-    } else {
-      const connector = await Registry.get(warehouseName)
-      result = await connector.execute(params.sql, params.limit)
+    if (!fallbackName) {
+      throw new Error(
+        "No warehouse configured. Use warehouse.add, set ALTIMATE_CODE_CONN_* env vars, or configure a dbt profile.",
+      )
     }
+    const connector = await Registry.get(fallbackName)
+    const result: SqlExecuteResult = await connector.execute(params.sql, params.limit)
     try {
       Telemetry.track({
         type: "warehouse_query",
--- a/packages/opencode/src/altimate/native/connections/registry.ts
+++ b/packages/opencode/src/altimate/native/connections/registry.ts
@@ -133,6 +133,29 @@ const DRIVER_MAP: Record<string, string> = {
   trino: "@altimateai/drivers/trino",
 }
 
+// altimate_change start — canonical driver identity for workspace precedence.
+/**
+ * Collapse a `config.type` onto the canonical name of the driver that serves it, so
+ * callers reasoning about "which database is this really" cannot be fooled by an
+ * alias: `postgresql` and `postgres` are one driver, as are `mariadb`/`mysql`,
+ * `mssql`/`fabric`/`sqlserver`, and `mongo`/`mongodb`.
+ *
+ * Derived by inverting `DRIVER_MAP` rather than restating it, so a type added there
+ * cannot silently desync from everything keyed on driver identity. Returns null for a
+ * type no driver serves.
+ *
+ * `redshift` maps to its own driver and therefore stays distinct from `postgres`: a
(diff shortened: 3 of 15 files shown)


ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
## Codex review log

Kept in one comment and edited in place, so the PR description stays short. Every finding below was confirmed against the code or the running system before anything was changed; each fix's test was checked against the pre-fix version and observed to fail.

### Round 1 — 2 findings, both real

| Finding | Verdict | Fix |
|---|---|---|
| Precedence wasn't gated on the workspace pilot flag | Confirmed. The binding and the pinned entry both persist in config and the MCP client connects that entry regardless of the flag, so engine tools materialise for someone who opted out. Checked live: with the flag unset, an engine still spawned and served its tools | Derivation returns disabled unless the pilot is on |
| The default-target resolver ignored the native fallback after dbt | Confirmed. The dbt helper returns null on an unrecognised result shape and inside a bare catch, not only when dbt is absent, and the handler then falls through to the first registry connection | The fallback is reported alongside the dbt target, and a served fallback redirects |

### Round 2 — 2 findings, both real

| Finding | Verdict | Fix |
|---|---|---|
| The redirect jumped over the hard deny | Confirmed, and a safety regression. DROP DATABASE, DROP SCHEMA and TRUNCATE are blocked with a message saying it cannot be overridden, and the engine tools apply no equivalent list — verified, neither resolver references the classifier. A blocked statement came back as an instruction to call the engine tool | Deny runs first |
| The redirect jumped over explain's input validators | Confirmed. A redirect reads as success, so malformed input was sent onward instead of reported | Validators run first. Schema inspection has no pre-flight checks and is unaffected |

### Round 3 — 1 finding, real

| Finding | Verdict | Fix |
|---|---|---|
| The redirect jumped over the write confirmation | Confirmed. The write permission is "ask", but an engine tool key is matched by the builder agent's allow-all rule, so a redirected INSERT, UPDATE or DELETE reached the warehouse without the confirmation the same statement needed a moment earlier | The guard runs last, after every native safety check |

This corrected my own reasoning. I had put precedence ahead of the write prompt because approving a write that then redirects seemed to authorise something that never runs. The write does run — through the engine — and what the user authorises is the write, not which connection carries it.

### Routed in from the stacked PR — 1 finding, real

| Finding | Verdict | Fix |
|---|---|---|
| The fail-open notice was attached only to the success path | Confirmed, and wider than reported. Review named three catch blocks; auditing every exit after the guard found three more. An undetermined target usually means a misconfigured setup, so those calls fail more often than average — the telemetry marker went missing in exactly the population it exists to measure | Every failure exit carries it; schema inspection routes its failures through one helper so one cannot be overlooked |

### Round 4 — 1 finding, real

| Finding | Verdict | Fix |
|---|---|---|
| The undetermined verdict returned before the fallback was weighed | Confirmed. A dbt target whose type cannot be identified failed open even when the connection behind it is served — and an undetermined type makes reaching that fallback more likely, not less | The fallback is weighed before giving up on the type |

That was the third defect in the order of the same branches. The pattern was the real finding: each fix was correct and each reopened a sibling path, because reaching a dbt-sourced target through the live path needs a real dbt project, so the order was never directly testable. The decision is now a pure function with the order documented and its own suite.

### Round 5 — 1 finding, real

| Finding | Verdict | Fix |
|---|---|---|
| A redirect could send a caller to a tool it is not permitted to call | Confirmed, in the code and against the running system. The analyst agent denies everything it does not name; it names the native warehouse tools but never the engine keys. So on a served connection its permitted reads were redirected to a tool it is forbidden to call — taking away the one thing that agent exists to do | Shadowing now requires the destination to be reachable as well as materialised and attributed. All three redirect sites are gated; an absent ruleset means unknown and stays reachable |

This is the counterpart to round 3. There, an agent that allows everything let a redirected write through without the confirmation the native path required. Here, an agent that denies everything unnamed made the redirect impossible to follow. The same redirect crossing the same permission boundary fails in opposite directions depending on the agent, which is the general lesson: a redirect is only sound if the destination is reachable under the caller's own rules.

### Round 6 — 1 finding, real

| Finding | Verdict | Fix |
|---|---|---|
| The listing claimed routing that would not happen | Confirmed against the running system. After round 5, a caller denied the engine key had its calls run locally — but the warehouse listing still reported the connection as served by the workspace and counted it as shadowed. An analyst was being told to expect routing its own permissions rule out, in the very output the model reads before choosing a tool | Every user-facing report goes through one helper returning the capabilities that materialised *and* whose destination this caller may call. Partial reachability is reported per capability |

Round 5 fixed the routing and left the reporting behind it. Worth naming as its own class: when a decision gains a new condition, everything that *describes* that decision inherits the condition too, and the describing code is easy to forget because nothing it does is wrong on its own.

### Round 7 — 2 findings, both real

| Finding | Verdict | Fix |
|---|---|---|
| Attribution read the saved config rather than the running engine | Confirmed. An MCP entry can be rewritten from unpinned to pinned while the already-connected process keeps serving, so the config names this workspace while the running engine serves another — precedence would then route calls into someone else's engine | Attribution now requires the attach outcome as well as the pin: the outcome is the runtime signal, the pin catches a later rewrite naming somewhere else |
| The per-session caches were unbounded | Confirmed. A long-running server sees an unbounded number of session ids and each entry holds a merged permission ruleset | Same cap and insertion-ordered eviction as the attach module |

The first is the same class as the pilot-flag and reachability findings: something outside this module can change under it, and reading the convenient signal rather than the authoritative one leaves a window. Verified against the running system before and after, because a wrong reading would have disabled routing everywhere silently rather than failing loudly.

### Round 8 — 1 finding, real

| Finding | Verdict | Fix |
|---|---|---|
| Attribution awaited the attach task | Confirmed. Round 7's fix read the attach outcome by awaiting the attach itself, which is deliberately uncapped — the prompt loop bounds its own wait and lets a turn proceed without engine tools past the cap. Awaiting it hung every affected turn for the full connection timeout, reintroducing from this side the delay the attach work had just removed | Read through the attach module's new read-only accessor instead. Not known yet and never attached are indistinguishable, and both fail open with a notice |

Triage of this one took two false starts, both environmental rather than code: a shared fixture whose entry another session had disabled, and the shared engine binary crashing before it could print a version — which attach reports as "engine too old", a plausible product outcome rather than a broken toolchain. Two smokes were read as evidence against the fix before the environment was checked. The corrective is now a standing rule: before attributing a behaviour change to code, verify the fixture's entry and that the engine actually runs.

### Round 9 — 2 findings, both real, both P2

| Finding | Verdict | Fix |
|---|---|---|
| A tool's description claimed redirection for capabilities it does not have | Confirmed. The claim was made whenever the workspace served *any* capability for that warehouse type, so an execute-only integration had explain and inspect described as redirecting while their bodies correctly stayed local — steering the model away from the local tool that works | Each tool claims redirection only for its own capability; the listing keeps its whole-listing note |
| A stale inventory was never corrected | Confirmed. The line was announced once per session. The first turn can legitimately say routing is off — an attach outrunning its bounded wait looks the same as no engine — and precedence is re-derived every turn, so the truth changes under a session already told otherwise | The last announced line is remembered and compared: corrections are delivered, unchanged inventories stay quiet |

Both are the same class as round 6 — when a decision gains a condition, the code describing it inherits the condition — this time at capability granularity rather than warehouse type. Three findings across the run have now been that class, which is why every user-facing report resolves through the same predicate the routing uses.

### Round 10 — clean

No findings on `a06841c5c`. Nothing inline, no review, and the verdict arrived as a plain comment: *"Didn't find any major issues."* That closes the cycle: nine rounds produced thirteen confirmed findings, and the tenth — run against the commit the ninth had just modified — produced none.

### Round 11 — 2 findings, both real

| Finding | Verdict | Fix |
|---|---|---|
| **P1** — a routing decision could outlive its binding | Confirmed. Re-linking mid-session is supported, so the turn's snapshot can name a workspace the project has already left, and a redirect naming it would send the query to that workspace's engine with its credentials | The binding is re-checked before any redirect — a local cache read, paid only on the path about to redirect |
| P2 — an evicted snapshot ran silently | Confirmed. A session with no snapshot returned a bare "run", indistinguishable from a considered "not served", so an entry evicted between tool resolution and the call would let a shadowed connection execute locally with no indication | Unknown now runs locally *and says so*, like every other undetermined case |

The P1 is the within-turn drift recorded as a residual early in the design — it had been named but not closed, and review found the concrete consequence. Worth noting that a documented residual is not a handled one.

### Round 12 — 1 finding, real

| Finding | Verdict | Fix |
|---|---|---|
| **P1** — routing ending was never announced | Confirmed. The announcement compared inventories, which catches a capability appearing or disappearing but not routing stopping altogether: with nothing left to enumerate the line is empty, and an empty line was read as nothing to say. A session told its calls were routed kept believing it after the engine stopped serving anything | The transition is stated explicitly, and only to a session that had previously been told otherwise |

Third finding in the same class — the code that describes a decision keeps asserting what used to be true. Twice that was a stale claim; here it was silence, which is the same failure wearing different clothes.

### Round 13 — 2 findings, 1 real, 1 refuted as stated

| Finding | Verdict | Fix |
|---|---|---|
| P1 — recheck connections after concurrent registry updates | **Refuted as stated; reachable variant found and fixed.** The claim was that a parallel `warehouse.add` could land between the routing decision and `Registry.get(name)`. It cannot: for a named warehouse there is no suspension point anywhere on that path — the tool body does not await between the verdict and `Dispatcher.call`, the dispatcher's lazy-registration await is cleared after the first call in the process, the handler falls through to the registry with no await, and `Registry.get` reads its config synchronously before its own first await. Single-threaded JS cannot interleave. The *warehouse-less* path is a different story: it awaits the dbt attempt and only afterwards re-reads the registry, so the connection executed need not be the one the decision covered | The handler read the registry twice, either side of an await. It now reads once, before it. Decided and executed connections are the same by construction, and a connection that disappears mid-flight is reported by name instead of quietly running somewhere else |
| P2 — an announcement that was not routing counted as routing | Confirmed. Round 12's fix remembered the last announcement as a bare string, so any non-empty line counted as "was routing". But "shadowing off, the engine could not be attributed" is a statement about a failed attribution, not about served capabilities — a session that got that first and later refreshed to an attributed engine serving nothing was told routing had stopped when it had never started | The announcement is remembered alongside whether it described actual routing, and only a session that was routing can be told routing stopped |

First finding in this run that is wrong as written. Worth recording as such: the fix it prompted is real, but it is not the fix that was asked for, and taking the suggestion at face value would have added a recheck to a path that cannot race while leaving the path that can untouched. The round-12 P2 is also the first regression introduced by a review fix rather than found in the original design.

### Round 14 — 1 finding, real

| Finding | Verdict | Fix |
|---|---|---|
| P2 — an announcement that never arrived was remembered as said | Confirmed, and the suggested remedy could not be applied as written. The line was recorded before publishing, and `announce` swallowed its own failure, so the call site could not tell a delivered line from a dropped one — "record only after a successful publish" was not expressible until `announce` reported delivery at all. A toast lost to a briefly unavailable event bridge was remembered as delivered, and every later turn with the same inventory skipped it as unchanged | `announce` reports whether the line reached the session; a failed delivery restores the previous record so the next turn retries. The record is still written before publishing, so a second refresh in the same window cannot send it twice — and restoring the *previous* value rather than clearing it keeps whether the session had been routing, which a later "routing stopped" depends on |

Fifth finding in the class that keeps recurring — the code that describes a decision keeps asserting what used to be true. Here the false assertion was internal: the map claimed the session had been told something it had not.

Both tests were checked against a wrong implementation as well as the real one: the first fails under record-before-publish-and-never-retry, the second under clear-on-failure, which loses the routing state and silently suppresses a later stop. A fix this small has two plausible wrong shapes, and only one of them is the reported defect.

### Round 15 — 1 finding, real, and it is a regression in round 14's fix

| Finding | Verdict | Fix |
|---|---|---|
| P2 — a failed rollback reinstated a line that was never delivered | Confirmed. Round 14 rolled a failed publication back to the record it replaced, but that record was itself optimistic. Two refreshes can publish different lines before either settles; when both failed, the second's rollback reinstated the first's undelivered line *as delivered*, and a later turn returning to that inventory skipped it as already said | Confirmed deliveries and in-flight attempts are separate maps. Nothing reaches the delivered map until the line arrives, so a failure leaves the session's state untouched and the next turn retries. Each attempt is its own object, so identity settles which publication is current — one finishing after a newer one started finds its attempt gone and records nothing |

Two rounds in a row on the same eight lines, and the second was caused by the first. Worth being plain about why: round 14 fixed the symptom it was handed — an undelivered line remembered as delivered — by restoring the previous value, without asking whether that value was itself trustworthy. It was not. The bug was never the rollback; it was that one map was being asked to hold two different facts, what the session has been told and what is currently being sent. Splitting them is what actually closes it, and it makes the out-of-order case fall out for free rather than needing its own guard.

Both tests were checked against a wrong implementation as well as the real one: the overlapping-failure test fails under round 14's rollback, and the out-of-order test fails when the identity guard is dropped.

### Round 16 — 1 finding, real, and the half of round 13's P1 that was left open

| Finding | Verdict | Fix |
|---|---|---|
| **P1** — the fallback was pinned by name, not by what the name meant | Confirmed, and correctly scoped: round 13's fix closed the case where the *identity* of the default connection changes mid-flight, and this is the case it did not close. The same name can be re-added against a different warehouse while the call is suspended in the dbt attempt, and `Registry.get(name)` still consults the mutable registry afterwards, so the call would execute under a decision computed for the connection that name used to mean | The routing decision is a function of the connection's canonical type, so pinning the type pins the decision. A replacement whose type differs is refused, naming the connection; a rewrite that keeps the type — including an alias change like `postgres` to `postgresql` — cannot change where the call is routed and still runs |

This one was known: the scope limit was recorded when round 13's fix landed, as the part needing the verdict bound to the connector rather than the name. Review found it independently, which is the useful outcome — it means the gap was real rather than theoretical, and it arrived with the evidence needed to fix it narrowly. The fix stays inside the handler: no coupling to precedence, no session identity, no enforcement at the connector layer that Decision 9 rejected.

Both tests were checked against a wrong implementation as well as the real one, in both directions: the replacement test fails with no guard at all, and the still-runs test fails when the guard compares config identity instead of canonical type — too strict turns every harmless rewrite into a spurious failure, which is the more likely mistake of the two.

### Round 17 — 1 finding, real, and it corrects the shape of rounds 14–16

| Finding | Verdict | Fix |
|---|---|---|
| P2 — two successful publications could land out of order | Confirmed, but only after checking the premise. Two `refresh()` calls for one session cannot interleave: `resolveTools` is awaited inside the step loop, and a second prompt for a running session parks on a callback instead of entering it. What *can* overlap is the publishing — `announce` is deliberately not awaited, so a toast can never stall a turn — which means two lines can be in flight across consecutive turns. Landing out of order leaves the stale line on screen while the newer one is recorded as the session's state | Publications are chained per session, so arrival order matches decision order and the last line a session sees is the newest |

This one is worth reading as a correction to the three rounds before it rather than a fourth bug. The announcement path had accumulated a guard per race: a rollback (14), an identity check (15), a scope pin (16). Each was a correct answer to the finding it was handed. But ordering is a property worth having outright, not a race worth guarding, and once publications are chained the overtaken-publication case cannot occur at all — so the two tests written to guard that race now assert the ordering instead. Fewer moving parts than before the round.

The premise check mattered and I nearly skipped it: the previous round's overlapping-publication finding was accepted and fixed without first establishing that two refreshes could overlap. They cannot. The scenario is reachable anyway — through the un-awaited publish rather than through concurrent refreshes — so the fix stands, but it was reached by luck rather than by evidence, and the mechanism recorded in that round's entry was wrong.

### Round 18 — 1 finding, real

| Finding | Verdict | Fix |
|---|---|---|
| P2 — an evicted session's announcement could resurrect itself | Confirmed. Publishing is not awaited, so a line can still be in flight when its session falls out of the cache; writing the delivery back afterwards recreated an entry eviction had already removed. Eviction only ever walks `bySession`, so nothing could reclaim it, and the announcement cache would grow with the lifetime session count — the bound the eviction exists to hold | The write-back happens only while the session is still tracked |

This is the first finding in four rounds on this path that is an invariant violation rather than a race: the suite already asserts these caches are bounded, and this was a way through that assertion. That is the distinction now being applied — a stated invariant that can be broken gets fixed; a marginal race on a path that already has an invariant gets declined with the evidence.

The test asserts the behaviour rather than the size. A size bound alone passed for an incidental reason on the first attempt — the count it produced (283) did not correspond to the defect being described, which is the same trap as a test named for one case that exercises another. Re-derived, a resurrected record suppresses the announcement an evicted session should get when it comes back, and that assertion fails for exactly the right reason.

### Round 19 — 1 finding, real, and fixing it removed a condition

| Finding | Verdict | Fix |
|---|---|---|
| P2 — a correction back to the delivered line was suppressed | Confirmed. Inventory can return to what was already announced while a different line is still being published; the dedupe compared only against the *delivered* line, so the correction was dropped and the queue delivered the stale line last | The comparison is now against the newest line the session is committed to saying — the one still in flight if there is one, otherwise the one it has been told |

Reachability is thin: it needs a toast publication to outlive a full turn, and the state self-corrects on the next refresh, so the exposure is one turn of stale guidance rather than an indefinite one. Under the standing rule that would be a candidate to decline. It was fixed instead because the fix *removes* a condition rather than adding a guard — the two separate comparisons, one against the delivered line and one against the pending line, collapse into a single question with the same answer whenever nothing is pending. A finding that makes the code smaller is worth taking regardless of how marginal the trigger is.

### Round 20 — 1 finding, real; last round

| Finding | Verdict | Fix |
|---|---|---|
| P2 — routing that stopped before its announcement landed was never reported stopped | Confirmed, and it is the other half of round 19. That round made the dedupe compare against the newest line a session is committed to saying, but left the *stop* decision reading the delivered record alone. With the first routing line still in flight, a refresh serving nothing queued no correction — and the routing line then arrived after routing had already stopped | Both questions now consult one record: the announcement still being published if there is one, otherwise the one the session has been told |

Taken rather than declined on the same test as round 19: the fix removes the second source of truth instead of adding a guard for the case. It also finishes a change that had been applied to one of the two questions and not the other, which is its own kind of defect — a half-applied invariant reads as deliberate and is harder to spot than an absent one.

### Test consolidation

Seven tests retired into three that state the property directly, after the last round.

The attestation block asserted three refused attach outcomes individually alongside a loop over the whole union, so the samples were already subsumed. The loop now carries what they were adding — the refusal *reason*, which is what the inventory line and the tool descriptions render, so a refusal with the wrong reason is a wrong explanation shown to the user — and covers `undefined`, which it could not reach before. The two qualifying outcomes became one loop over the other half of the same allowlist. The four reporting surfaces, each asserted against the same unreachable-caller ruleset, now assert together that every surface agrees with the routing decision — which is the actual property, given that three findings in this review were a surface still asserting what had stopped being true.

Every consolidated invariant was checked by breaking what the deleted tests used to catch: making `superseded` qualify, making one reporting surface claim routing regardless of reachability, and reporting the wrong refusal reason. Each fails the surviving invariant.

Line count barely moves — 1,003 to 1,001 — because an invariant carries more scaffolding and more explanation than the sample it replaces. That is the honest number, and it is not the point: the gain is that a new outcome variant or a new reporting surface is covered by construction rather than by remembering to add a case. Source grew 695 to 748 across the nine rounds of fixes.

### Residuals, and one thing that is not one

The PR names three cases where shadowing turns itself off and the call runs locally. All
fail open, and all live in the attach machinery rather than in this change. The detail
behind the one-liners:

- **Attach re-validation window.** A session's outcome reads as unsettled while its memo
  is re-validated, so a turn resolving tools inside that window sees no attribution.
  Being addressed in the attach layer by awaiting the bounded validation. The point
  worth keeping: "unsettled" must keep meaning in-flight rather than being papered over
  with the last settled value — remembering past re-validation would reintroduce exactly
  the stale attribution the attestation exists to prevent. If a genuine spawn-class
  re-attach still exceeds the bounded wait afterwards, "nothing is served any more" is a
  true statement and the notice is correct, not a flap.
- **Session-cache eviction.** Past 256 tracked sessions the oldest entry goes while its
  session may still be live; the next call re-derives it. Bounded and deliberate — the
  alternative is a map that grows with the lifetime session count, which is the bound
  round 18 existed to protect.
- **Cross-session teardown.** Another session's refusal can remove the shared client
  while this session's attribution still reads attached, so a redirect can name a tool
  that is no longer connected. It surfaces as a tool-layer failure rather than a query
  sent to the wrong warehouse, which is the distinction that matters here.

**Not a residual, recorded because it looks like one.** The pin parser fails open on
every string-shaped miss but throws on a non-string token after the last pin, or on an
object-shaped command. That is reachable through a raw-disk read elsewhere in the
codebase — but not through this code: attribution reads via the config cache on both
passes, the cached one and the confirmation after invalidating it, and the config path
normalises the shape before the parser sees it. The read is also wrapped in a catch that
refuses. So it is guarded twice, in the safe direction, and patching it here would be
patching the wrong place.

That safety rests on an asymmetry worth stating plainly, because it is the same
reasoning behind confirming the pin against disk only when the cached answer is about to
enable: on the attribution path, staleness and failure are dangerous in one direction
only. A stale or errored "pinned elsewhere" merely refuses; only a wrong "pinned to us"
routes a query. Everything there is arranged to fail toward refusing, which is why a
parser throw on that path is a non-event rather than a vulnerability.

### Why the later rounds keep landing in the same two places

Thirteen findings reduce to a handful of shapes, and two of them account for six:

- **A guard placed where an early return skips a check that matters** (3). The redirect returns early, so anything above it stops running — that took out the hard deny on destructive statements, the explain validators, and the write confirmation in turn. Each fix was correct and each moved the problem rather than removing it, because the ordering was never directly testable: reaching a dbt-sourced target through the live path needs a real dbt project. The decision is now a pure function with its order documented and its own suite.
- **Describing code that keeps asserting what used to be true** (3). When the routing decision gained a condition — reachability, then per-capability reachability — the listing, the inventory line and the tool descriptions carried on advertising the old rule. Nothing they did was wrong on its own; they simply repeated a fact that had stopped holding. Every user-facing report now resolves through the same predicate the routing uses.

Both are structural rather than careless, which is why they recurred: the code made them easy. That is also the argument for having kept going — the later rounds were not finding new instances of the first bug, they were finding new *classes*.

### Tally

Twenty-six findings across twenty rounds plus one routed in — twenty-five real and fixed, one refuted as stated. One false positive in twenty-six, and it still surfaced a real defect next door.

Two of the twenty were regressions introduced by review fixes rather than defects in the original design, both in the announcement path and both in consecutive rounds. That is the honest cost of fixing at the point a finding lands: round 12's fix created round 13's P2, and round 14's created round 15's. Each was caught by the next round, which is the argument for running rounds until one is clean rather than stopping when the findings look small.

The yield never decayed. Rounds 7, 8 and 9 were each run to see whether the well had run dry, and each returned something real; round 10 came back clean and rounds 11–20 returned eleven more. Two patterns account for most of it — a guard placed where an early return skips a check that matters (3), and describing code that keeps asserting what used to be true (5, once by staying silent and once by lying to itself). Both are structural rather than careless, which is why they recurred: the code made them easy. That is the argument for having kept going — the later rounds were not finding new instances of the first bug, they were finding new classes.

Two findings were a pair worth naming: a redirect crosses a permission boundary, and the same crossing fails in opposite directions for an agent that allows everything versus one that denies everything unnamed. And one lesson arrives late — round 13's P1 is a reminder that a confident finding still has to be traced to a suspension point before it is believed.

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
@codex review

ralphstodomingo:
Superseded by #1168 (the precedence layer of the re-cut stack #1167 → #1168 → #1169, derived overlay, never persisted). Closing; branch left in place for reference.

PR #1172, opened 2026-08-28
feat(workspace): sync a bound workspace's custom skills into the project

### LINKED ISSUE #1173: [Feature] Sync a linked workspace's skills into the project
## Description

When a project is linked to a workspace, the skills uploaded to that workspace
should be available to the agent in that project. Today they are not: skills
have to exist on the machine already, so anything authored in the workspace is
invisible to the CLI.

## Use Case

A team keeps its shared conventions as skills on a workspace — how their dbt
project is laid out, which warehouse patterns are house style, the runbook for a
recurring migration. Someone uploads a skill; everyone whose project is linked
to that workspace should get it.

Concretely:

- A new joiner clones a linked repo and gets the team's skills without being
  told they exist or copying files by hand.
- Someone fixes a skill in the workspace and the correction reaches teammates,
  rather than each person carrying a private copy that drifts.
- A skill can ship supporting files — reference notes, examples — not just a
  single markdown file, since that is how non-trivial skills are actually
  written.

The gap is most obvious with worktrees and second clones: the same repo, the
same user, the same linked workspace, but nothing arrives.

## Proposed Solution

Sync a linked workspace's skill bundles onto disk in the project, where the
existing skill discovery already looks. Whole bundle, so bundled reference files
come with it.

Behaviour that matters:

- Refresh when a skill is added or changed in the workspace, without needing a
  restart.
- Never confuse a failed fetch with an empty workspace — a network failure must
  leave whatever is already on disk alone, because "empty" is the answer that
  deletes it.
- Never replace or delete a directory the client did not create.
- Stop serving a workspace's skills when the account is disconnected or the
  feature is turned off.
- Keep the synced tree out of the user's version control.

## Alternatives Considered

- **Copy skills by hand.** What happens today. Drifts immediately and does not
  survive a new clone.
- **Fetch on demand when the model asks for a skill.** Puts a network round-trip
  in the middle of a turn, and gives no way to list what is available up front.
- **A separate CLI command to pull skills.** Workable, but it is another thing to
  remember, and the common failure is not knowing the skills exist at all.

## Pre-submission Checklist

- [x] I have searched existing issues and this is not a duplicate
- [x] I have described a concrete use case, not just a feature in the abstract
- [x] I understand that feature requests require a linked issue before a PR will be accepted

sahrizvi:
### Issue for this PR

Closes #1173

### Type of change

- [x] New feature

### What does this PR do?

A project bound to an Altimate workspace pulls that workspace's uploaded skill
bundles into `.altimate-code/skill/_workspace/<public_id>/`, where the existing
skill discovery finds them. Whole bundle, including `references/`, because the
Skill tool hands the model the skill's directory and it reads those files
itself. Keyed on `public_id`, not name, so a rename in the SaaS doesn't orphan a
directory.

Syncs on bind and on the first turn, then re-polls every 5 minutes so a skill
added in the SaaS reaches a session that is already open. Activation is
model-dependent for v0: skills appear in `<available_skills>` and load when the
model invokes the Skill tool.

Two design rules do most of the work:

- **Error is never treated as empty.** A failed or malformed list keeps whatever
  is on disk. "Empty workspace" is the one answer that deletes the tree, so it
  has to be unambiguous. Same rule for the binding lookup and file reads.
- **Only replace a tree we own.** The managed directory is ours if it is absent
  or carries a manifest we can parse. Anything else is a user's file and the
  sync declines rather than deleting it.

Binding resolution falls back to the server, because the local cache is written
only by an explicit link — so a git worktree, a second clone or a teammate's
checkout looked unbound and got nothing. The lookup is access-controlled
server-side, so it can only surface a binding the caller could already see.
Adoption also enables the ongoing memory mirror; only the one-shot backfill
stays behind an explicit link. Adopted rows are marked so that stays
distinguishable.

Requires the backend custom-skills API, which is on `development` and not yet
in `main`.

Unit: 38 tests on the sync, plus discovery and memory coverage. Full suite
green. Every guard is mutation-checked — the code is broken deliberately and the
test must fail. Two guards are honestly not pinned and are marked as such in the
commits.

End to end against a local backend on `development` with real bundles in S3:

- bundle lands with `references/` intact and byte-exact
- **a model invokes a synced skill and reads its bundled reference** — a skill
  whose `SKILL.md` deliberately omits the answer and points at
  `references/codeword.md`, holding a random token present nowhere else. The
  agent called the Skill tool, read the reference, and returned the token.
- a skill added in the SaaS reaches an already-open session
- a project bound only on the server adopts and syncs
- backend down leaves the snapshot untouched; a rebind purges the previous
  workspace's skills
- disconnect removes the snapshot; reconnect restores it; corrupt credentials
  keep it

A 7-reviewer consensus review returned REQUEST CHANGES; all blocking findings
are fixed (symlink traversal in the sweep, ownership on any `readdir` error,
cross-process staging deletion, account-switch leakage), plus prompt-injection
escaping of skill bodies, a response-size bound, and pagination validation.
Deferred items are listed in the commits.

### Screenshots / recordings

Not a UI change.

### Checklist

https://claude.ai/code/session_012wmN54fRA4WLgdVJVunaNk

## Summary by CodeRabbit

- **New Features**
  - Workspace-specific skill bundles now sync automatically and become available without restarting.
  - Skills can be refreshed when new or updated bundles are detected.
  - Workspace bindings can be resolved and revalidated from the server.
- **Bug Fixes**
  - Added safeguards for oversized or incomplete API responses.
  - Improved protection against malformed, unsafe, or misleading skill content and file paths.
  - Existing synced skills are preserved when synchronization fails.
  - Disabled synchronization now removes previously synced workspace skills.

### DIFF
--- a/packages/opencode/src/altimate/workspace/skill-sync.ts
+++ b/packages/opencode/src/altimate/workspace/skill-sync.ts
@@ -522,7 +522,12 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
     // credentials" — the first is a decision the user made and must take
     // effect, the second is unknown, and unknown never destroys a snapshot.
     if (!(await AltimateApi.isConfigured())) {
-      if (await deactivate(canon, "no altimate credentials")) changed = true
+      // Symlink-guarded like every other deletion here. This branch runs BEFORE
+      // the `pathsAreReal` check below, so without its own guard a symlinked
+      // `.altimate-code` has `removeManaged`/`sweepStaging` resolve through the
+      // link and delete the target's tree. (bot review)
+      if ((await pathsAreReal(canon).catch(() => false)) && (await deactivate(canon, "no altimate credentials")))
+        changed = true
       return
     }
 
--- a/packages/opencode/src/altimate/workspace/skill-sync.ts
+++ b/packages/opencode/src/altimate/workspace/skill-sync.ts
@@ -87,17 +87,6 @@ function managedRoot(directory: string): string {
  * start racing on the same project do not both stage and swap. */
 const inFlight = new Map<string, Promise<void>>()
 
-/** Bumped whenever a sync actually changes what is on disk, for any project.
- * Consumers that cache a view of the skill registry compare the value they last
- * acted on against this one, so a sync triggered elsewhere — a mid-session bind,
- * say — still causes them to refresh, without them having to re-run the sync
- * themselves just to learn whether anything moved. */
-let generation = 0
-
-export function snapshotGeneration(): number {
-  return generation
-}
-
 async function readManifest(directory: string): Promise<Manifest | null> {
   try {
     const raw = await fs.readFile(path.join(managedRoot(directory), MANIFEST_NAME), "utf8")
@@ -315,7 +304,6 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
   } finally {
     inFlight.delete(canon)
   }
-  if (changed) generation++
   return { changed }
 }
 
--- a/packages/opencode/src/altimate/workspace/skill-sync.ts
+++ b/packages/opencode/src/altimate/workspace/skill-sync.ts
@@ -124,6 +124,15 @@ function managedRoot(directory: string): string {
  * stays in the name so the cross-PROCESS liveness guard still works. */
 const REALM_ID = Math.random().toString(36).slice(2, 8)
 
+/** How long a sibling realm's staging tree is presumed live.
+ *
+ * Sparing every same-pid tree that is not ours protects a concurrent thread,
+ * but a worker killed mid-sync leaves one behind that no later realm would ever
+ * collect — permanent disk garbage. A sync takes seconds, so a tree untouched
+ * for this long belongs to a thread that died rather than one in flight.
+ * (review) */
+const STAGING_LEASE_MS = 30 * 60 * 1000
+
 const STORE_KEY = Symbol.for("altimate.workspace.skill-sync.store")
 
 interface SyncStore {
@@ -467,7 +476,19 @@ async function sweepStaging(directory: string): Promise<void> {
     // this realm's own tree is swept: a sibling thread's staging is in flight,
     // and deleting it makes that thread publish a tree missing everything
     // written so far, under a manifest that claims the files.
-    if (owner && owner === String(process.pid) && !mine) continue
+    if (owner && owner === String(process.pid) && !mine) {
+      // A sibling realm's tree — but only while it still looks live. Past the
+      // lease it belongs to a thread that died mid-sync, and no other realm
+      // would ever collect it.
+      let age = 0
+      try {
+        age = Date.now() - (await fs.stat(path.join(dir, entry))).mtimeMs
+      } catch {
+        continue // vanished under us; nothing to collect
+      }
+      if (age < STAGING_LEASE_MS) continue
+      log.info("collecting a staging tree left by a terminated worker", { entry, ageMs: age })
+    }
     if (owner && owner !== String(process.pid) && processAlive(Number(owner))) continue
     const target = path.join(dir, entry)
     try {
@@ -695,6 +716,13 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
     await fs.mkdir(path.join(canon, STAGING_DIR), { recursive: true })
     await fs.writeFile(path.join(canon, STAGING_DIR, ".gitignore"), "*\n").catch(() => {})
     await fs.rm(staging, { recursive: true, force: true })
+    // Created up front rather than as a side effect of the first file write.
+    // With every skill failing nothing was written, so the later `.gitignore`
+    // write failed with ENOENT and the outer catch abandoned the snapshot — the
+    // right outcome reached by accident, which left the explicit "nothing
+    // survived" guard below unreachable and one refactor away from silently
+    // publishing an empty tree. (review)
+    await fs.mkdir(staging, { recursive: true })
     const next: Manifest = {
       version: 1,
       tenant: creds.altimateInstanceName,
@@ -711,10 +739,20 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
         // out of the whole loop, so one bad file cost the workspace EVERY skill,
         // on every client, re-attempted every poll forever. The blast radius of a
         // bad bundle is now that bundle. (review)
+        // Validated BEFORE the try, not inside it. The catch below builds
+        // filesystem paths from this id, and `path.join(staging, "..")` is the
+        // staging PARENT — a recursive delete there takes every sibling
+        // thread's in-flight and retired tree with it. Throwing into the catch
+        // for an unsafe id put a remote string back into a recursive delete,
+        // which is the exact thing `safePathComponent` exists to prevent.
+        // (review)
+        if (!safePathComponent(summary.publicId)) {
+          skipped += 1
+          failed = true
+          log.warn("skipping a workspace skill with an unusable id", { skill: summary.publicId })
+          continue
+        }
         try {
-          if (!safePathComponent(summary.publicId)) {
-            throw new WorkspaceApiError(`unusable skill id in the workspace listing: ${summary.publicId}`)
-          }
           const detail = await altimateRequest<unknown>(
             "GET",
             `/${encodeURIComponent(summary.publicId)}`,
@@ -765,12 +803,16 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
             await fs.writeFile(dest, bytes)
             recorded[file.path] = file.size
           }
-          next.skills[summary.publicId] = { updatedAt: summary.updatedAt, files: recorded }
           totalFiles += skillFiles
           totalBytes += skillBytes
           next.skills[summary.publicId] = { updatedAt: summary.updatedAt, files: recorded }
         } catch (err) {
           skipped += 1
+          // Keeps the run out of the poll window. A partial snapshot that
+          // stamped `lastSyncedAt` would leave the failed skill unretried for a
+          // full interval, which is the opposite of what a failure should
+          // cause — the same rule the whole-run failure path follows. (review)
+          failed = true
           // A skill that synced before keeps its previous copy rather than
           // disappearing over a transient failure — the same "error is never
           // emptiness" rule the rest of this file follows, applied per skill.
--- a/packages/opencode/src/altimate/workspace/skill-sync.ts
+++ b/packages/opencode/src/altimate/workspace/skill-sync.ts
@@ -305,6 +305,30 @@ async function sweepStaging(directory: string): Promise<void> {
   }
 }
 
+/** Take the snapshot out of service when this client is no longer entitled to
+ * serve it — the account was disconnected, or the feature was switched off.
+ *
+ * Leaving it is not neutral. Discovery loads whatever is on disk without
+ * consulting the manifest, so a disconnected user keeps getting the workspace's
+ * skills, and any of them carrying ``alwaysApply`` keeps being injected into
+ * every prompt. Returns whether anything was actually removed, so the caller
+ * knows to refresh the registry.
+ *
+ * Only removes a tree this client owns, for the same reason the sync does. */
+async function deactivate(directory: string, why: string): Promise<boolean> {
+  const root = managedRoot(directory)
+  try {
+    await fs.stat(root)
+  } catch {
+    return false // nothing published here
+  }
+  if (!(await ownsManagedDir(directory))) return false
+  await removeManaged(directory)
+  await sweepStaging(directory)
+  log.info("removed the workspace skill snapshot", { why, path: root })
+  return true
+}
+
 async function removeManaged(directory: string): Promise<void> {
   await fs.rm(managedRoot(directory), { recursive: true, force: true })
 }
@@ -315,8 +339,14 @@ async function removeManaged(directory: string): Promise<void> {
  * failure path leaves whatever is already on disk in place, except the
  * deliberate purge described below. */
 export async function syncSkills(directory: string): Promise<{ changed: boolean }> {
-  if (!isEnabled()) return { changed: false }
   const canon = path.resolve(directory)
+  if (!isEnabled()) {
+    // Opting out has to actually take effect: a snapshot left behind keeps
+    // loading into every session.
+    const dropped = await deactivate(canon, "the workspace feature is off").catch(() => false)
+    if (dropped) snapshotChangedAt.set(canon, Date.now())
+    return { changed: dropped }
+  }
   const existing = inFlight.get(canon)
   if (existing) {
     await existing.catch(() => {})
@@ -328,6 +358,16 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
   // project been "checked", and only then should the poll interval start.
   let sawRemote = false
   const run = (async () => {
+    // Checked BEFORE the binding: `resolveBinding` needs credentials too, so a
+    // disconnected client would otherwise return on a null binding and never
+    // reach this. Disconnected is different from "could not read the
+    // credentials" — the first is a decision the user made and must take
+    // effect, the second is unknown, and unknown never destroys a snapshot.
+    if (!(await AltimateApi.isConfigured())) {
+      if (await deactivate(canon, "no altimate credentials")) changed = true
+      return
+    }
+
     // `resolveBinding`, not `readLocalBinding`: the local cache is written only
     // by an explicit link, so a project bound server-side (fresh clone, new
     // machine, cleared state) would otherwise never get its workspace's skills.
@@ -350,7 +390,9 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
     try {
       creds = await AltimateApi.getCredentials()
     } catch (err) {
-      log.warn("no altimate credentials; skipping skill sync", { err: String(err) })
+      log.warn("could not read altimate credentials; keeping the existing snapshot", {
+        err: String(err),
+      })
       return
     }
 
--- a/packages/opencode/src/altimate/workspace/skill-sync.ts
+++ b/packages/opencode/src/altimate/workspace/skill-sync.ts
@@ -8,15 +8,26 @@
 // skills document MCP tool names rather than this CLI's tools — neither is
 // synced here.
 //
-// Activation is deliberately NOT handled: a synced skill is discovered like any
-// other, listed in ``<available_skills>`` by name + description, and loaded when
-// the model invokes the Skill tool. Whatever frontmatter an author put in the
-// bundle flows through untouched.
+// Activation is not handled HERE, but that is not the same as "cannot happen".
+// A synced skill is discovered like any other: normally it is listed in
+// ``<available_skills>`` by name + description and loaded only when the model
+// invokes the Skill tool. Whatever frontmatter an author put in the bundle
+// flows through untouched — including ``alwaysApply`` and ``applyPaths``, which
+// discovery carries into ``Info`` (skill/index.ts) and which
+// ``collectAutoLoadedSkills`` (session/system.ts) injects into every applicable
+// system prompt with no Skill-tool call and no permission prompt.
+//
+// That is a real consequence worth stating plainly: anyone who can upload a
+// skill to a workspace can put standing instructions into the prompts of every
+// member bound to it. The backend has no activation field, so this can only
+// arrive through the uploaded SKILL.md. Whether workspace skills should be
+// allowed to auto-activate is a product decision, not one this module should
+// make silently by stripping frontmatter an author wrote.
 //
 // Server contract (app/api/datamates/custom_skills.py, mounted at ``/skills``):
 //   GET ""                       -> Page[CustomSkillSummary]  (paginated)
 //   GET "/{public_id}"           -> CustomSkillDetail  (adds files[] + content)
-//   GET "/{public_id}/files/{p}" -> raw file bytes
+//   GET "/{public_id}/files/{p}" -> {path, content} JSON (NOT raw bytes)
 // The summary carries ``file_count``, not an inventory, and nothing in the API
 // exposes a checksum — ``CustomSkillFileMeta`` is ``{path, size}``. So change
 // detection is per-skill ``updated_at`` and the only integrity check available
@@ -99,6 +110,14 @@ const inFlight = new Map<string, Promise<void>>()
  * list call per project per interval, and the list is Postgres-only server-side
  * (no S3 reads), so the check is cheap when nothing changed. */
 const POLL_INTERVAL_MS = 5 * 60 * 1000
+
+/** Ceilings on one snapshot. Nothing upstream bounds a workspace's size, and
+ * every file is read fully into memory before it reaches disk, so without these
+ * a single oversized bundle is an out-of-memory crash rather than a failed
+ * sync. Exceeding either abandons the snapshot the same way any other error
+ * does — the previous one is kept. */
+const MAX_TOTAL_BYTES = 32 * 1024 * 1024
+const MAX_TOTAL_FILES = 2000
 
 /** Last SUCCESSFUL sync per canonical project, for the interval above. A failed
  * attempt must not stamp: doing so suppresses retry for a full interval on a
@@ -206,6 +225,11 @@ function parsePage(payload: unknown): { rows: RemoteSummary[]; pages: number } |
   const p = payload as { items?: unknown; pages?: unknown }
   if (!Array.isArray(p.items)) return null
   const pages = typeof p.pages === "number" && p.pages >= 0 ? p.pages : 1
+  // An empty page while the envelope claims rows exist is a proxy or backend
+  // inconsistency, not an empty workspace — and "empty workspace" is the one
+  // answer that deletes the user's snapshot. Refuse it.
+  const total = (payload as { total?: unknown }).total
+  if (p.items.length === 0 && typeof total === "number" && total > 0) return null
   const rows: RemoteSummary[] = []
   for (const row of p.items) {
     if (!row || typeof row !== "object") return null
@@ -385,6 +409,8 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
       skills: {},
     }
     try {
+      let totalFiles = 0
+      let totalBytes = 0
       for (const summary of remote) {
         if (!safePathComponent(summary.publicId)) {
           throw new WorkspaceApiError(`unusable skill id in the workspace listing: ${summary.publicId}`)
@@ -398,6 +424,13 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
         if (!files) throw new WorkspaceApiError(`unrecognised detail for ${summary.publicId}`)
         const recorded: Record<string, number> = {}
         for (const file of files) {
+          totalFiles += 1
+          totalBytes += file.size
+          if (totalFiles > MAX_TOTAL_FILES || totalBytes > MAX_TOTAL_BYTES) {
+            throw new WorkspaceApiError(
+              `workspace skill bundle exceeds the client limit (${totalFiles} files, ${totalBytes} bytes)`,
+            )
+          }
           const encoded = file.path.split("/").map(encodeURIComponent).join("/")
           // The file endpoint answers with ``{path, content}`` JSON, not the raw
           // object — the server decodes the bundle file and hands back a string.
@@ -437,9 +470,26 @@ export async function syncSkills(directory: string): Promise<{ changed: boolean 
       // Written into staging so it lands atomically with the snapshot.
       await fs.writeFile(path.join(staging, ".gitignore"), "*\n")
       await fs.writeFile(path.join(staging, MANIFEST_NAME), JSON.stringify(next, null, 2))
-      await fs.rm(root, { recursive: true, force: true })
+      // Move the live tree aside rather than deleting it first. `rm` then
+      // `rename` leaves a window with no snapshot at all — a crash or a reader
+      // inside it sees the skills vanish. The retired tree is removed only
+      // after the new one is in place.
       await fs.mkdir(path.dirname(root), { recursive: true })
-      await fs.rename(staging, root)
+      const retired = path.join(canon, STAGING_DIR, `retired-${process.pid}`)
+      await fs.rm(retired, { recursive: true, force: true }).catch(() => {})
+      let hadPrevious = true
+      try {
+        await fs.rename(root, retired)
+      } catch {
+        hadPrevious = false // nothing published yet
+      }
+      try {
+        await fs.rename(staging, root)
+      } catch (err) {
+        if (hadPrevious) await fs.rename(retired, root).catch(() => {})
+        throw err
+      }
+      await fs.rm(retired, { recursive: true, force: true }).catch(() => {})
       changed = true
       log.info("workspace skills synced", {
         datamateId: binding.datamateId,
--- a/packages/opencode/src/altimate/workspace/skill-sync.ts
+++ b/packages/opencode/src/altimate/workspace/skill-sync.ts
@@ -33,6 +33,7 @@
 // detection is per-skill ``updated_at`` and the only integrity check available
 // is byte length.
 import fs from "fs/promises"
+import { statSync } from "node:fs"
 import path from "path"
 import { Flag as CoreFlag } from "@opencode-ai/core/flag/flag"
 import { Log } from "@/altimate/util/log"
@@ -97,9 +98,37 @@ function managedRoot(directory: string): string {
   return path.join(directory, MANAGED_DIR)
 }
 
+/** Every mutable table below is anchored on a process-global rather than being
+ * plain module state. This file is reached through two different module graphs
+ * in the same process — the bind path resolves it via one specifier and the
+ * per-turn hook via another — and the runtime keeps a separate module record
+ * for each, so module-level `Map`s silently fork — `inFlight` included, which
+ * would let a bind and a turn stage and swap the same project concurrently.
+ * A symbol-keyed global gives every copy of this module the same tables.
+ *
+ * Note this only reaches copies sharing a realm. Threads do NOT share
+ * `globalThis`, so anything a bind must hand to a later turn goes through disk
+ * instead — see `snapshotFingerprint`. */
+const STORE_KEY = Symbol.for("altimate.workspace.skill-sync.store")
+
+interface SyncStore {
+  inFlight: Map<string, Promise<{ changed: boolean }>>
+  lastSyncedAt: Map<string, number>
+  registryAppliedAt: Map<string, number>
+  syncedFor: Map<string, string>
+}
+
+const globals = globalThis as unknown as Record<symbol, SyncStore | undefined>
+const store: SyncStore = (globals[STORE_KEY] ??= {
+  inFlight: new Map(),
+  lastSyncedAt: new Map(),
+  registryAppliedAt: new Map(),
+  syncedFor: new Map(),
+})
+
 /** In-flight sync per canonical project directory, so a bind and a session
  * start racing on the same project do not both stage and swap. */
-const inFlight = new Map<string, Promise<{ changed: boolean }>>()
+const inFlight = store.inFlight
 
 /** How long a snapshot is trusted before the next turn re-checks the workspace.
  *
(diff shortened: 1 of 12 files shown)


packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,867 @@
...
+          if (content === null) {
+            throw new WorkspaceApiError(`unrecognised file body for ${summary.publicId}/${file.path}`)
+          }
+          // No checksum exists in the API, so length is the only integrity check
+          // available. `size` is the stored object's byte count, so the
+          // comparison has to be on UTF-8 bytes rather than string length — the
+          // two differ for any non-ASCII skill. It still catches a truncated
+          // download, which is what would otherwise publish half a skill.
+          const bytes = Buffer.from(content, "utf8")
+          if (bytes.byteLength !== file.size) {
```
saravmajestic: **Blocking.** This check is correct, but against the real backend it makes any workspace containing a non-text bundle file permanently unsyncable — and the failure takes the whole snapshot, not just the file.

The server side (`altimate-backend`, `development`):

- `validate_bundle` (`app/service/custom_skills/bundle.py:69`) imposes **no** content-type restriction. Only `SKILL.md` must be valid UTF-8 — `decode_skill_md` raises on it, and nothing else is checked. So a bundle may legitimately hold a PNG in `references/`, a zip, a font.
- `size` in the inventory is the raw stored byte count: `written.append({"path": path, "size": len(content)})` (`app/service/custom_skills/storage.py:161`).
- But the file endpoint returns text, not bytes: `return {"path": key, "content": raw.decode("utf-8", errors="replace")}` (`app/api/datamates/custom_skills.py:860`).

For any file that is not valid UTF-8, `errors="replace"` substitutes U+FFFD and the re-encoded length can never equal the stored size. Because this `throw` is inside the per-file loop, the outer `catch` abandons the entire staged snapshot — so one binary file in one skill means that workspace never receives **any** skill, on any client, and re-downloads the full set every 5 minutes indefinitely.

Neither test layer can catch this: the fake server derives size from the string (`size: Buffer.from(c).byteLength`), so it can only ever produce content that round-trips, and the E2E used markdown bundles.

The check itself is the right instinct — the real gap is that the API has no way to transport a non-text bundle file, and the client treats that as fatal to the snapshot rather than to one file. Worth choosing between: fail one skill instead of the whole snapshot; skip-and-log a file whose length cannot be verified; or serve raw/base64 server-side for non-text paths.

packages/opencode/src/session/system.ts
```diff
@@ -149,7 +149,17 @@ export namespace SystemPrompt {
...
+        // altimate_change start — neutralise the closing tag inside the body.
+        // The name is escaped but the body was not, so content containing
+        // `</auto_loaded_skill>` closed the wrapper and continued as unwrapped
+        // system-prompt text — able to impersonate the harness's own framing,
+        // directly after the prompt has told the model to treat this as binding
+        // guidance. Skill bodies are now remote content (a bound workspace
+        // syncs them), so this is reachable by anyone who can upload a skill.
+        // Deliberately not a full XML escape: bodies legitimately contain code
+        // and angle brackets, and mangling those would break working skills.
+        parts.push(neutralizeSkillWrapper(skill.content.trim()))
```
saravmajestic, line 161: This fix is right, but it covers the *less* reachable of the two places remote skill text reaches the system prompt.

`Skill.fmt` (`packages/opencode/src/skill/index.ts:424-425`) still interpolates raw:

```ts
`    <name>${skill.name}</name>`,
`    <description>${skill.description}</description>`,
```

`description` comes straight from bundle frontmatter (`md.data.description`, `skill/index.ts:162`). It is pre-existing in `main` — but this PR is precisely what makes it remote-attacker-controlled, and the reasoning in the comment above applies verbatim: *"Skill bodies are now remote content (a bound workspace syncs them), so this is reachable by anyone who can upload a skill."*

The listing path is the more exposed one: `auto_loaded_skill` requires the author to set `alwaysApply`/`applyPaths`, whereas **every** synced skill's description lands in `<available_skills>` in **every** session. A description ending `...</description></skill></available_skills>` followed by fabricated harness framing breaks out of the listing structure the same way the body did.

Same narrow treatment would do — neutralise the wrapper tag names rather than a full XML escape, so code samples in descriptions still survive.

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,867 @@
...
+
+    // Stage a complete snapshot, then swap. A partial bundle is never
+    // published: any failure abandons the staging directory and leaves the
+    // previous snapshot untouched.
+    const root = managedRoot(canon)
+    // Staged OUTSIDE `.altimate-code/skill/`, because discovery globs
+    // `{skill,skills}/**/SKILL.md` from the config dir — a staging tree that
+    // lived beside `_workspace` would be scanned, so a half-downloaded snapshot
+    // (or one abandoned by a SIGKILL) would be loaded as real skills.
+    const staging = path.join(canon, STAGING_DIR, `pending-${process.pid}`)
```
saravmajestic: The staging name is unique per **process**, but this module's own comments say the colliding parties are **threads**.

From the `STORE_KEY` block: *"this only reaches copies sharing a realm. Threads do NOT share `globalThis`"*, and from `registryAppliedAt`: *"A bind and a turn do not share memory: the runtime loads this module once per thread, so each gets its own module record AND its own `globalThis`"* — which is why `snapshotFingerprint` has to go through disk at all.

Threads share a pid but not `globalThis`, so for a bind and a turn on different threads:

- `inFlight` does not serialise them (different `globalThis`, different table), and
- both compute the same `pending-<pid>` path, and
- `sweepStaging`'s guard only spares *another process's* tree — for its own pid it deletes, so each run sweeps the other's in-flight staging.

The TUI leaves the bind sync detached (`state.ts:518`, `awaitBackfill` unset), so bind-during-turn is the ordinary case rather than a rare race. Result is a published tree mixing two runs under one run's manifest; `upToDate`'s size check repairs it on the next poll, so it is self-healing but visible to the model in the window.

A per-realm random suffix alongside the pid closes it, and keeps the cross-process guard working since the pid is still in the name. `retired-${process.pid}` on line 752 has the same shape.

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,867 @@
...
+// Activation is not handled HERE, but that is not the same as "cannot happen".
+// A synced skill is discovered like any other: normally it is listed in
+// ``<available_skills>`` by name + description and loaded only when the model
+// invokes the Skill tool. Whatever frontmatter an author put in the bundle
+// flows through untouched — including ``alwaysApply`` and ``applyPaths``, which
+// discovery carries into ``Info`` (skill/index.ts) and which
+// ``collectAutoLoadedSkills`` (session/system.ts) injects into every applicable
+// system prompt with no Skill-tool call and no permission prompt.
+//
+// That is a real consequence worth stating plainly: anyone who can upload a
```
saravmajestic, line 20: Flagging this rather than disagreeing with it — the analysis is right and stating it plainly here was the correct call.

But this is the single largest behavioural change in the PR: anyone who can upload a skill to a workspace can put standing instructions into every bound member's system prompt, with no Skill-tool call and no permission prompt. "Whether workspace skills should be allowed to auto-activate is a product decision" is exactly right, which is why it should carry an explicit product sign-off before merge, rather than shipping with the decision recorded only in a source comment where it is easy for nobody to have made it.

Worth noting the blast radius depends on who can upload: if any workspace member can, this is member-to-member prompt injection by design. A v0 that strips `alwaysApply`/`applyPaths` from *synced* bundles (leaving local skills untouched) would ship the feature without that question needing an answer first — offered as an option, not a request.

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,867 @@
...
+ * list call per project per interval, and the list is Postgres-only server-side
+ * (no S3 reads), so the check is cheap when nothing changed. */
+const POLL_INTERVAL_MS = 5 * 60 * 1000
+
+/** Ceilings on one snapshot. Nothing upstream bounds a workspace's size, and
+ * every file is read fully into memory before it reaches disk, so without these
+ * a single oversized bundle is an out-of-memory crash rather than a failed
+ * sync. Exceeding either abandons the snapshot the same way any other error
+ * does — the previous one is kept. */
+const MAX_TOTAL_BYTES = 32 * 1024 * 1024
```
saravmajestic, line 172: Minor: this ceiling and the server's do not compose.

The backend caps a bundle at `MAX_BUNDLE_BYTES = 10 * 1024 * 1024` per skill (`app/service/custom_skills/bundle.py:29`), so four skills near that limit exceed the 32MB snapshot ceiling here. Since exceeding it throws and abandons the whole snapshot, a workspace that is perfectly legal server-side becomes wholly unsyncable client-side, with the same all-or-nothing shape as the size-mismatch case above.

Not necessarily wrong as a memory guard, but the two limits should be chosen with reference to each other, and the failure is probably better as "skip what does not fit" than "publish nothing".

packages/opencode/src/session/prompt.ts
```diff
@@ -112,12 +149,7 @@ export namespace SessionPrompt {
   // The trace span is a sibling of the root (tracing.ts:1009 assigns
   // parentSpanId to rootSpanId), not a nested child — good enough for
   // waterfall correlation via timestamps, and no schema change is required.
-  async function traceSpan<T>(
-    name: string,
-    fn: () => Promise<T>,
-    input?: unknown,
-    sessionID?: SessionID,
-  ): Promise<T> {
+  async function traceSpan<T>(name: string, fn: () => Promise<T>, input?: unknown, sessionID?: SessionID): Promise<T> {
```
saravmajestic: Minor: this reflow is unrelated to the feature, as are the `lastAssistantHasToolParts` and `taskTool` hunks further down — a formatter run rather than a change. The PR checklist ticks "I have not included unrelated changes in this PR".

Not worth another round on its own; worth splitting out if you touch the branch again, since it costs blame on lines this PR has no reason to own.

saravmajestic (changes requested):
Reviewed against both repos — the backend `custom_skills` API this depends on is the half the CI bots can't see, and that is where the blocking finding is.

The client-side engineering here is genuinely strong: the error-vs-empty discipline is applied consistently and correctly (`parsePage` rejecting a missing `pages`, the empty-page-with-nonzero-`total` guard, `null` meaning unknown at every layer), `ownsManagedDir` refuses on `ENOTDIR`/`EACCES` rather than only `ENOENT`, deletions are `lstat`-gated, and the move-aside-then-swap with rollback plus the streaming response bound are all correct and non-obvious. If the backend API is not in `main` yet, degradation is safe: the list throws, returns `null`, and nothing is written.

Requesting changes on one issue that makes the feature silently dead for a class of workspaces, plus one escaping gap that this PR's own threat model implies.

**1. Blocking — a single non-UTF-8 bundle file kills the whole workspace's sync, permanently.** Cross-repo; details inline on the size check.

**2. Escaping landed on the less reachable of the two paths.** `neutralizeSkillWrapper` is right, but `Skill.fmt` still interpolates `name`/`description` raw into `<available_skills>`, and that path needs no `alwaysApply`. Details inline.

**3. Staging path collides across threads** — same pid, different `globalThis`, which is the topology this module's own comments describe.

**4. The auto-activation exposure deserves an explicit product sign-off**, not only a code comment — it is the largest behavioural change in the PR.

Two minor notes inline (client/server size ceilings, unrelated formatting churn).

packages/opencode/src/altimate/workspace/skill-sync.ts
sahrizvi: Verified this end to end before changing anything — `validate_bundle` gates count/size/paths/`SKILL.md` and imposes no content-type restriction, `size` is `len(content)` raw, and the file endpoint returns `raw.decode("utf-8", errors="replace")`. So a legal bundle holding a PNG can never satisfy the length check, and the throw took the whole snapshot with it. Fixed in e8c3d14.

Failure is now per skill. The rest of the snapshot publishes, and a skill that had synced before keeps its previous copy rather than disappearing over a transient error. If nothing survives, the snapshot is abandoned rather than published empty — an empty publish is indistinguishable from a broken server and would delete skills the user still has.

Your point about the test layer was the sharpest part and it was right: the fake server derives `size` from a string, so it could only ever emit content that round-trips. The new case forges the mismatch `errors="replace"` actually produces, and it fails if the per-skill catch is removed.

I went with "fail one skill" over "skip one file" deliberately — a size mismatch is also what a truncated download looks like, and publishing a bundle missing a file its `SKILL.md` points at seemed worse than withholding the bundle. Happy to revisit if you'd rather have the partial.

packages/opencode/src/session/system.ts
sahrizvi: Right, and the reachability argument is the part I'd missed — a body needs `alwaysApply`, whereas every synced description lands in `<available_skills>` in every session. Fixed in e8c3d14.

`Skill.fmt` now runs `name` and `description` through a neutraliser for the listing's own tag names, same narrow treatment as the body rather than a full XML escape. Tested both directions: the injected `</description></skill></available_skills>` no longer closes the element, and `use <T> generics` still renders intact.

packages/opencode/src/altimate/workspace/skill-sync.ts
sahrizvi: Confirmed, and it follows directly from the topology documented a few lines above — which makes it a gap I should have caught when I wrote that comment. Fixed in e8c3d14.

Staging is now `<kind>-<pid>-<realm>` with a per-realm id, so two threads in one process never compute the same path. The sweep spares a sibling realm's tree while still deleting our own and still honouring the cross-process liveness check, and the regex tolerates the older `<kind>-<pid>` names. Test creates a sibling realm's staging directory and asserts a sync leaves it alone.

packages/opencode/src/altimate/workspace/skill-sync.ts
sahrizvi: Agreed it needed a decision rather than a comment, and it now has one: auto-activation is **approved for the pilot**. Synced bundles keep `alwaysApply`/`applyPaths`.

Updated the comment in e8c3d14 to record that as an accepted, bounded exposure rather than an open question, including your framing that the blast radius depends on who can upload, with a note to revisit if upload rights widen beyond the pilot's members. Kept your strip-frontmatter option in mind as the fallback if that changes.

packages/opencode/src/altimate/workspace/skill-sync.ts
sahrizvi: Correct, and it shares the all-or-nothing shape of the blocking one. Fixed in e8c3d14.

The ceiling is now checked per skill, against its declared size, before downloading — so a workspace that is legal server-side (10MB per bundle) loses only the bundles that don't fit rather than all of them. The two limits still aren't derived from each other; if you'd rather the client ceiling were expressed in terms of `MAX_BUNDLE_BYTES` so they can't drift apart, say and I'll do that.

packages/opencode/src/session/prompt.ts
sahrizvi: Fair, and the checklist tick was wrong. Reverted in e8c3d14 — seven pure-reformatting hunks in `prompt.ts` (line joins/splits and one reindent) are back to `main`, leaving only the marked changes plus one real addition (`existsSync`, used by the opt-out gate).

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,938 @@
...
+              await fs.rm(path.join(staging, summary.publicId), { recursive: true, force: true })
+              await fs.cp(priorDir, path.join(staging, summary.publicId), { recursive: true })
+              next.skills[summary.publicId] = prior
+              carried = true
+            } catch {
+              carried = false
+            }
+          }
+          if (!carried) {
+            await fs.rm(path.join(staging, summary.publicId), { recursive: true, force: true }).catch(() => {})
```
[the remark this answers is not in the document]
sahrizvi: Fixed in `dfb2054`. `safePathComponent` now runs before the `try`, so an unsafe id skips the skill without any filesystem path being built from it.

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,938 @@
...
+            }
+            const dest = path.join(staging, summary.publicId, file.path)
+            await fs.mkdir(path.dirname(dest), { recursive: true })
+            await fs.writeFile(dest, bytes)
+            recorded[file.path] = file.size
+          }
+          next.skills[summary.publicId] = { updatedAt: summary.updatedAt, files: recorded }
+          totalFiles += skillFiles
+          totalBytes += skillBytes
+          next.skills[summary.publicId] = { updatedAt: summary.updatedAt, files: recorded }
```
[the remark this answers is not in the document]
sahrizvi: Fixed in `dfb2054` — dead duplicate removed.

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,938 @@
...
+        } catch (err) {
+          skipped += 1
```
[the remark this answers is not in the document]
sahrizvi: Fixed in `dfb2054`. `failed = true` on skip, so a partial run stays out of the poll window and the failed skill retries next turn.

packages/opencode/test/altimate/workspace/skill-sync.test.ts
```diff
@@ -0,0 +1,1254 @@
...
+    const inner = globalThis.fetch
+    globalThis.fetch = (async (input: string | URL, init?: RequestInit) => {
+      if (String(input).includes("/files/")) throw new Error("all downloads fail")
+      return inner(input as never, init as never)
+    }) as unknown as typeof fetch
+    await syncSkills(project)
+
+    // Carried forward, not deleted, and not replaced by an empty tree.
+    expect(readFileSync(skillFile("pub-1", "SKILL.md"), "utf8")).toBe("one")
+  })
```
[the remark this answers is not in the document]
sahrizvi: Right — it passed with the guard deleted. Rewritten in `dfb2054` to use a skill with no prior copy, which is the only shape that reaches the guard. Staging is also created up front now: previously nothing was written, so an incidental ENOENT abandoned the snapshot and the guard was unreachable. Deleting it now fails four tests.

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,938 @@
...
+    // deleting a live owner's staging makes it publish a snapshot missing every
+    // file written before the sweep, with a manifest that claims them.
+    // `<kind>-<pid>` (older trees) or `<kind>-<pid>-<realm>` (current).
+    const owner = /-(\d+)(?:-[a-z0-9]+)?$/.exec(entry)?.[1]
+    const mine = entry.endsWith(`-${process.pid}-${REALM_ID}`)
+    // Another live process's tree is never touched. Within THIS process only
+    // this realm's own tree is swept: a sibling thread's staging is in flight,
+    // and deleting it makes that thread publish a tree missing everything
+    // written so far, under a manifest that claims the files.
+    if (owner && owner === String(process.pid) && !mine) continue
```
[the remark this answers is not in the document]
sahrizvi: Fixed in `dfb2054`. A sibling realm's tree is spared only inside a 30-minute lease; past that it belonged to a worker that died mid-sync and is collected.

packages/opencode/test/skill/skill.test.ts
```diff
@@ -568,4 +568,112 @@ description: A skill in the .opencode/skills directory.
       { git: true },
     ),
   )
+
+  // altimate_change start — a synced skill's frontmatter is remote content, and
+  // the listing is the path that needs no `alwaysApply` to be reached.
+  it.live("the available_skills listing cannot be closed from a skill description", () =>
```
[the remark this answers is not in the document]
sahrizvi: Declined — plain `it` leaves an unhandled error between tests here, so it stays on `it.live`. Noted as a harness limitation rather than intent.

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,938 @@
...
+            const content = parseFileContent(body, file.path)
+            if (content === null) {
+              throw new WorkspaceApiError(`unrecognised file body for ${summary.publicId}/${file.path}`)
+            }
+            // No checksum exists in the API, so length is the only integrity check
+            // available. `size` is the stored object's byte count, so the
+            // comparison has to be on UTF-8 bytes rather than string length — the
+            // two differ for any non-ASCII skill. It still catches a truncated
+            // download, which is what would otherwise publish half a skill.
+            const bytes = Buffer.from(content, "utf8")
```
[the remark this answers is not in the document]
sahrizvi: Partly. `e8c3d14` stops a binary reference killing the whole snapshot — that skill is skipped and the rest publish. The underlying gap stands: the file API has no way to transport non-UTF-8 bytes, so such a bundle can never sync completely. That needs a server-side change (raw or base64 for non-text paths); tracked separately rather than worked around here.

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,834 @@
...
+  // project been "checked", and only then should the poll interval start.
+  let sawRemote = false
+  const run = (async () => {
+    // Checked BEFORE the binding: `resolveBinding` needs credentials too, so a
+    // disconnected client would otherwise return on a null binding and never
+    // reach this. Disconnected is different from "could not read the
+    // credentials" — the first is a decision the user made and must take
+    // effect, the second is unknown, and unknown never destroys a snapshot.
+    if (!(await AltimateApi.isConfigured())) {
+      if (await deactivate(canon, "no altimate credentials")) changed = true
```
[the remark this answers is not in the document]
sahrizvi: Fixed in `e6d5756`. The no-credentials purge is `pathsAreReal`-gated like every other deletion in that function.

packages/opencode/src/altimate/workspace/state.ts
```diff
@@ -245,6 +253,172 @@ function sameBinding(a: CachedBinding, b: CachedBinding): boolean {
...
+    }
+    const fresh = await lookupBinding(directory, key)
+    if (fresh.status === "unknown") {
+      // Cannot reach the server: keep serving what we have rather than tearing
+      // a working setup down over a network blip.
+      return { status: "bound", binding: local }
+    }
+    lastValidatedAt.set(canonicalizeKey(directory), Date.now())
+    if (fresh.status === "unbound") {
+      forgetBinding(directory, key)
```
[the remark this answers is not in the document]
sahrizvi: Fixed in `1147dc7`. An explicit bind clears the negative memo and stamps the row as server-validated, so a link inside `MISS_TTL_MS` is no longer undone by its own revalidation.

packages/opencode/src/altimate/workspace/state.ts
```diff
@@ -245,6 +253,172 @@ function sameBinding(a: CachedBinding, b: CachedBinding): boolean {
...
+      return { status: "unbound" }
+    }
+    // Rebound elsewhere: adopt the server's answer, replacing the cached row.
+    if (fresh.binding.datamateId !== local.datamateId) return fresh
```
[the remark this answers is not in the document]
sahrizvi: Fixed in `1147dc7`. Confirming a cached binding now preserves `seededAt`, `adopted` and `linkedAt` instead of relabelling it.

packages/opencode/src/altimate/workspace/state.ts
```diff
@@ -245,6 +253,116 @@ function sameBinding(a: CachedBinding, b: CachedBinding): boolean {
...
+  }
+  if (!hit) {
+    serverLookupMissed.set(canon, Date.now())
+    return { status: "unbound" }
+  }
+  // cubic P2: a malformed 2xx would otherwise throw on the dereference below,
+  // outside the try above, aborting the whole sync. An unrecognised body is
+  // unknown, not unbound — the same rule the rest of this feature follows.
+  const row = (hit as { binding?: Partial<Binding> }).binding
+  if (!row || typeof row.datamate_id !== "number" || typeof row.datamate_name !== "string") {
```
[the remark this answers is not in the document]
sahrizvi: Fixed in `1147dc7` — identity validation widened to the optional string fields.

packages/opencode/src/session/prompt.ts
```diff
@@ -272,6 +272,63 @@ export namespace SessionPrompt {
...
+    // for every instance plus a full re-scan. This runs under the instance ALS,
+    // which `attach()` propagates into the facade's runtime, so it invalidates
+    // THIS session's registry — verified by harness test.
+    try {
+      const skillSync = await import("../altimate/workspace/skill-sync")
+      // Nothing to do at all when the feature is off — not even a stat. This
+      // sits on the latency-measured path and runs for every user, including
+      // the ones who never opted in. NOT an early `return`: that would exit
+      // `prompt` itself and skip the message this function exists to create.
+      if (skillSync.isEnabled()) {
```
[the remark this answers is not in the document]
sahrizvi: Fixed. The opt-out path purges again, now behind a synchronous `existsSync` so an opted-out user with no snapshot pays one `stat` and no module load.

packages/opencode/src/skill/index.ts
```diff
@@ -444,6 +461,13 @@ export async function get(name: string) {
 export async function available(agent?: Agent.Info) {
   return runSkill((svc) => svc.available(agent))
 }
+// altimate_change start — imperative wrapper for the same reason as the three
```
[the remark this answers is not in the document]
sahrizvi: Fixed in `6a47be5` — nested marker removed (and its orphaned `end`, which broke the pairing check first time round).

packages/opencode/test/altimate/workspace/memory-sync.test.ts
```diff
@@ -142,6 +162,7 @@ beforeEach(() => {
   listFails = false
   createResult = [{ id: "mem-new" }]
   workspaces = [{ id: 42, name: "acme", memory_enabled: true }]
+  serverBinding = null
```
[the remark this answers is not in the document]
sahrizvi: Stale — `serverBinding` is read by the tests at `memory-sync.test.ts:1125/1141`, which drop the `resolveBinding` seam. The reset is doing work; leaving it.

packages/opencode/test/altimate/plugin/workspace.test.ts
```diff
@@ -275,6 +277,61 @@ describe("workspace binding cache", () => {
...
+      expect(afterFirst).toBeGreaterThan(0)
+
+      // Same workspace, same project: memory will skip, skills must not.
+      await recordApprovedBinding(proj, { ...binding, linkedAt: 2 }, { awaitBackfill: true })
+      expect(skillListCalls).toBeGreaterThan(afterFirst)
+    } finally {
+      globalThis.fetch = originalFetch
+      if (ORIGINAL_FLAG === undefined) delete process.env.ALTIMATE_WORKSPACE
+      else process.env.ALTIMATE_WORKSPACE = ORIGINAL_FLAG
+    }
```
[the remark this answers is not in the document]
sahrizvi: Deferred, not ignored: I tried scoping these per test and it broke 7 `materialize.test.ts` cases by materialising into the real home dir. Worth doing properly in a test-infra pass rather than inside this PR.

packages/opencode/test/altimate/workspace/skill-sync.test.ts
```diff
@@ -0,0 +1,1027 @@
...
+    // The cache is written by an explicit link and otherwise never expires, so
+    // without revalidation a project detached in the SaaS keeps serving its old
+    // workspace's skills on this machine forever.
+    serve({ "pub-1": { "SKILL.md": "one" } })
+    await syncSkills(project)
+    expect(existsSync(skillFile("pub-1", "SKILL.md"))).toBe(true)
+    // The local binding is still on disk — this is NOT the unbound-cache case.
+    expect(JSON.parse(readFileSync(cachePath(), "utf8")).bindings[realpathSync(project)]).toBeDefined()
+
+    globalThis.fetch = (async (input: string | URL) => {
```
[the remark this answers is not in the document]
sahrizvi: Same as the sibling thread — deferred to a test-infra pass. Related: `run-process.test.ts` is genuinely flaky from `cliIt.concurrent` (2/10 vs 0/10 serial), which I'm splitting into its own PR.

packages/opencode/src/altimate/workspace/skill-sync.ts
```diff
@@ -0,0 +1,980 @@
...
+        // on every client, re-attempted every poll forever. The blast radius of a
+        // bad bundle is now that bundle. (review)
+        // Validated BEFORE the try, not inside it. The catch below builds
+        // filesystem paths from this id, and `path.join(staging, "..")` is the
+        // staging PARENT — a recursive delete there takes every sibling
+        // thread's in-flight and retired tree with it. Throwing into the catch
+        // for an unsafe id put a remote string back into a recursive delete,
+        // which is the exact thing `safePathComponent` exists to prevent.
+        // (review)
+        if (!safePathComponent(summary.publicId)) {
```
saravmajestic, line 749: Confirming this is the right fix, and recording the severity for the file's history — the commit message understates it.

While the check sat inside the try, an unsafe id threw into the catch, which built paths from that same id and ran `fs.rm(..., { recursive: true, force: true })` on them:

```
".."                      -> <project>/.altimate-code/skill-staging
"../.."                   -> <project>/.altimate-code
"../../skill/_workspace"  -> <project>/.altimate-code/skill/_workspace   (the live snapshot)
"../../../.."             -> /
```

So not only "every sibling thread's in-flight and retired tree" — the traversal was unbounded, and a sufficiently long id reached the filesystem root. Reachable from a compromised or buggy backend response, which is the threat `safePathComponent`'s own docstring names.

Validating before the try and `continue`ing without touching disk is correct, and setting `failed = true` here is the right call too so the poll window is not stamped by a listing that contained an unusable id. Verified the guard is covered: removing this block fails a test.

saravmajestic (approved):
## Re-review: approving

All five findings from the first round are fixed, and the two rounds since have closed more than I raised. Verified against both repos and by running the suites, not from the commit messages.

**One correction to my own previous round, up front.** After `e8c3d146` I said all five were fixed and only a test gap remained. That was wrong: moving `safePathComponent` inside the per-skill `try` had introduced a path traversal, and I missed it. The catch built filesystem paths from the same untrusted id and ran a recursive delete on them, so a remote `public_id` reached `fs.rm(..., { recursive: true, force: true })`:

```
".."                      -> /proj/.altimate-code/skill-staging
"../.."                   -> /proj/.altimate-code
"../../skill/_workspace"  -> /proj/.altimate-code/skill/_workspace
"../../../.."             -> /
```

Worse than that commit's own description of it — the last one is a recursive delete of the filesystem root. `dfb20549` fixes it correctly by validating before the try and skipping without touching disk, and removing that check now fails a test. I should have caught it when reviewing the commit that introduced it; the per-skill refactor moved a guard out of the path it was guarding, which is exactly the shape a reviewer is there to notice.

### Verified fixed

| Finding | State |
|---|---|
| Non-UTF-8 bundle file killed the whole snapshot | Per-skill failure; the rest publishes, a previously-synced skill keeps its copy, and nothing-survived abandons rather than publishing empty |
| `Skill.fmt` interpolated `name`/`description` raw | `neutralizeListingWrapper` on both, narrow so `<T>` generics survive |
| `pending-<pid>` collided across threads | `REALM_ID` in the path; sweep spares a live sibling and collects a dead one past a 30-minute lease |
| Client/server ceilings did not compose | Checked per skill from declared size, before downloading |
| Auto-activation had no decision | Recorded as approved for the pilot, bounded by upload rights |

Plus, from the rounds after mine: the traversal above, the abandon guard made reachable, a partial snapshot no longer consuming the poll window, and the duplicated manifest write dropped.

### What I checked myself

- **CI has no unit-test job**, so "full suite green" is not machine-verified on this PR. I ran the four touched suites on Bun 1.3.14: **176 pass, 0 fail**.
- **Mutation-checked the two guards that matter.** Disabling the nothing-survived guard fails **8** tests (the commit message says four — it is better than claimed). Removing the pre-try `safePathComponent` check fails 1. Both are genuinely covered now; the earlier version of the abandon test passed with the guard deleted, which is what the rewrite fixed.
- **`fs.cp`**, which the new carry-forward depends on, works under this repo's Bun — a silent gap there would have made carry-forward never fire.
- **`location` needs no escaping.** `pathToFileURL` percent-encodes `<`, `>` and newlines (`abc%3C/location%3E…`), so a hostile id cannot form a tag there even though `safePathComponent` permits those characters. Checked because the listing fix escapes its siblings and this one is left raw.

### Withdrawing my formatting nit

My `prompt.ts` comment was not worth acting on and I sent you chasing it. Three reflows remain, but `printWidth` is 120 and those lines are 101-120 characters, so prettier *joins* them — `main` simply is not prettier-clean there, and any format run reintroduces them. That is the formatter doing its job, not churn anyone chose. I should have read the config before raising it.

### One note, not a request

The staging lease reads the staging root's `mtime`, which advances when a skill directory is created in it but not while files are written inside one. A single-skill sync running past 30 minutes could therefore have its tree collected by a sibling. With an 8MB per-file and 32MB per-snapshot ceiling that is not a realistic duration, and the outcome is a failed sync that retries — so this is an observation for the record rather than something to change.

Good work through three rounds on a genuinely difficult surface. The error-versus-emptiness discipline held all the way through the per-skill rewrite, which is the part that would have been easiest to lose.

packages/opencode/src/altimate/workspace/skill-sync.ts
saravmajestic: Resolved in `e8c3d146`, and verified against the real contract rather than the commit message.

Failure is now per skill: a bundle holding a PNG loses that bundle, the rest of the snapshot publishes, and a skill that synced before keeps its previous copy. Nothing-survived abandons rather than publishing empty, which is the right call - an empty publish is indistinguishable from a broken server.

The test forges what the API actually produces (`"\ufffd\ufffd"` against a declared size of 2), which is the only way to cover this given the fake server derives `size` from a string and so can only produce content that round-trips. Mutation-checked: disabling the nothing-survived guard now fails 8 tests. It failed none in the first version of that test, because the test seeded a skill that HAD a prior copy and so went down the carry-forward path - fixed in `dfb20549`.

packages/opencode/src/session/system.ts
saravmajestic: Resolved in `e8c3d146`. `neutralizeListingWrapper` covers `name` and `description`, with the same narrow treatment as the body so `<T>` generics in a description still read correctly.

I also checked the sibling this leaves raw - `location` - and it needs no escaping: `pathToFileURL` percent-encodes `<`, `>` and newlines, so `abc</location></skill>` becomes `abc%3C/location%3E%3C/skill%3E` and no tag can form. Worth knowing, since `safePathComponent` does permit those characters in an id.

packages/opencode/src/altimate/workspace/skill-sync.ts
saravmajestic: Resolved in `e8c3d146`, and improved further in `dfb20549`.

`REALM_ID` in the path fixes the collision, and the sweep spares a same-pid sibling while keeping the cross-process liveness guard. `dfb20549` then closed the gap that created - sparing every same-pid tree leaked one from a worker killed mid-sync, which no later realm could ever collect - by only sparing a sibling inside a 30-minute lease. That is the better answer; I had been prepared to accept the leak as a fair trade.

One note for the record, not a request: the lease reads the staging root's `mtime`, which advances when a skill directory is created in it but not while files are written inside one, so a single-skill sync running past 30 minutes could have its tree collected by a sibling. Not a realistic duration under an 8MB-per-file ceiling, and the outcome is a failed sync that retries.

packages/opencode/src/altimate/workspace/skill-sync.ts
saravmajestic: Resolved in `e8c3d146` - the decision is now recorded (approved for the pilot, bounded by who can upload to a workspace, revisit if upload rights widen).

That is what I was asking for. The point was never that stripping frontmatter was the right answer; it was that "a product decision" should not live only in a source comment as an open question, where it is easy for nobody to have made it. It has an owner and a scope now.

packages/opencode/src/altimate/workspace/skill-sync.ts
saravmajestic: Resolved in `e8c3d146`. The ceiling is checked per skill from its declared size before downloading, so a workspace that is legal server-side (10MB per bundle) loses only the bundles that do not fit rather than all of them - the same all-or-nothing shape as the size-mismatch case, fixed the same way.

packages/opencode/src/session/prompt.ts
saravmajestic: Withdrawing this one - it was not worth raising and I sent you chasing something that cannot be fixed by reverting.

Three reflows remain (`traceSpan`, `taskTool`, `hoistedReminders`), but `printWidth` is 120 and those lines are 101-120 characters, so prettier *joins* them. `main` is simply not prettier-clean there, and any format run reintroduces them. That is the formatter doing its job, not churn anyone chose.

I should have read `package.json`'s prettier config before flagging it. Nothing to do here.

PR #849, opened 2026-05-27
feat(validators): dbt-tests-pass validator + dbt-schema-verify hardening

anandgupta42:
### What does this PR do?

Completes the harness-side completion-gate validator framework. The registry and `dbt-schema-verify` were shipped in earlier PRs; this adds `dbt-tests-pass` and hardens the existing validator.

**New: `dbt-tests-pass` validator**

Fires after the agent declares done (`finishReason === "stop"`). Finds every dbt model `.sql` file the session modified (by mtime since `sessionStartMs`), runs `altimate-dbt test --model <name>` against each, and injects a synthetic user message if any tests fail — forcing one more agent turn to fix the SQL.

What this catches that schema-verify doesn't: row-data correctness errors. A model whose column names all match `schema.yml` can still fail `relationships`, `unique`, `not_null`, `accepted_values`, or `AUTO_*_equality` tests because the `SELECT` logic produces wrong values or wrong row counts.

**Hardened: `dbt-schema-verify`**

- `parseSchemaVerifyOutput()` — scans the stdout blob for the last balanced `{...}` block. `altimate-dbt` emits dbt log noise (ANSI codes, parser warnings) before the JSON verdict; `JSON.parse(stdout)` was failing on it. This fix was validated on `airbnb006` where the validator correctly triggered a schema-mismatch retry.
- Debug logging (spawn errors, close events, stdout/stderr heads) gated behind `ALTIMATE_VALIDATORS_DEBUG=1`.

**`prompt.ts`: diagnostic cleanup**

- All `console.error` calls gated behind `ALTIMATE_VALIDATORS_DEBUG=1` — previously they were always on, producing noise in normal sessions.
- Added missing `validatorsEnabled &&` guard on dispatch condition.
- Added `dispatch_enter`, `dispatch_result`, `dispatch_error` debug events.

**`build-local-tarball.sh`: `altimate-dbt` on PATH**

Added `"altimate-dbt": "./dbt-tools/bin/altimate-dbt"` to bin entries in the benchmark npm package. Validators spawn `altimate-dbt` and it wasn't in PATH inside the Docker containers used by ade-bench.

**`system.ts`: Marker Guard fix**

Moved `// altimate_change end` from inside the `skills()` function body (after the unreachable `return`) to outside it. The function's closing `}` was appearing as unguarded new code in the diff, tripping Marker Guard CI.

### Type of change
- [x] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)

[...]

PR #1213, opened 2026-08-31
fix(skills): stop walking the whole tree to answer "does any file match"

sahrizvi:
### Type of change

- [x] Bug fix (performance)
- [ ] New feature
- [ ] Documentation

### What does this PR do?

Skill `applyPaths` auto-load asks one question per skill — *does at least one file in the worktree match this glob* — and answered it with `Glob.scan(...).length > 0`. `scan` resolves only after the entire walk completes, so each skill paid for the whole tree even when the first directory already answered.

Two builtin skills ship with `applyPaths` (`dbt-develop`, `dbt-schema-verify`), so **every session pays this twice, before the first token** — no configuration, no custom skills, no workspace required.

How much it costs depends on what the worktree resolves to:

| where the session starts | `Instance.worktree` | scan cost |
|---|---|---|
| inside a git repo | the repo root | ~10ms |
| **outside any git repo** | **`/`** — `Project.fromDirectory` returns the global project (`project.ts:293`) | walks the whole filesystem |

`Glob.exists` uses `globIterate`, which yields lazily, so the walk is abandoned at the first match.

Same binary, same prompt, same account — only the implementation differs. Run from a directory outside a git repo:

| implementation | runs (s) |
|---|---|
| `scan(...).length > 0` (current) | 48.4, 52.9, 51.6 |
| **`Glob.exists` (this PR)** | **9.1, 6.8, 6.4** |
| scan removed entirely (floor) | 7.1, 6.8, 6.7 |

Inside a repo: **7.5, 7.3s — unchanged.** The fix lands within ~1s of the floor, so the remaining startup is not this scan.

Traces confirm where the time went: a session with 73.3s of startup returned its first generation in **0.08s**. The model was never the bottleneck.

Tests: four cases on `Glob.exists` pinning agreement with `scan`, `include: "file"` behaviour, `ignore` pruning, and a missing directory. Mutation-tested — dropping the options forwarding fails two of them.

- `packages/core`: 1072 pass, 26 fail — **the same 26 fail on unmodified `main`** (verified by stashing this change and re-running).
- `test/session` + `test/skill`: 1516 pass, 0 fail.
- `bun typecheck`: 13 tasks clean. Marker Guard: passes. No new Prettier violations.

### Rejected alternative

Passing `ignore: Glob.DEFAULT_IGNORE` here, mirroring what #1184 did for the MCP scans, **makes it slower**: 61–82s against a ~51s baseline. Outside a repo the tree is not dependency-dominated, so every candidate path pays 12 minimatch tests while almost nothing gets pruned. Early exit is the right lever for an existence check; pruning is the right lever when the tree really is mostly `node_modules`.

### Likely root cause

Two independent things combine; neither is a problem alone.

**1. `worktree` is `/` outside a git repo — present since the initial import.** `Project.fromDirectory` walks up looking for `.git`; finding none, it returns the global project with `worktree` and `sandbox` hardcoded to `/` (`project.ts:293`). That line is original to the repository's first commit (`f2cd5c124`, 2026-03-01) and carries no `altimate_change` markers, so it is inherited upstream code rather than something added here. As a sentinel meaning "no project", it was harmless — nothing walked it.

**2. A per-skill worktree scan — added 2026-05-29.** `anyMatchInWorktree` (#849, `a490bd45e`) walks `Instance.worktree` once per `applyPaths` skill, and the same change shipped two builtin skills carrying `applyPaths`. Inside a repo that root is bounded and each scan costs ~10ms, which is what the code was written against — the comment on the function reasons explicitly about catching `dbt_project.yml` "no matter how deep the user's cwd is". Outside a repo, the same call inherits `/`.

So the combination has existed since **2026-05-29**, and what determines whether anyone feels it is simply where the CLI is launched from: inside a git repo it is invisible, outside one it costs ~45s per session.

**No evidence of a recent regression.** The function is byte-identical since it landed, the two builtin skills gained `applyPaths` in that same commit, and worktree resolution has not changed since the import. Reports clustering recently are more consistent with where sessions are being started than with any code change — a same-machine, same-account A/B differed by 51s vs 8s on working directory alone.

This PR fixes the second half, which is the part that turns an inert sentinel into a filesystem walk. The first half is noted below.

### Not addressed here

`Project.fromDirectory` hardcodes `worktree: "/"` when no `.git` is found. Beyond performance, that means skills auto-load based on unrelated files elsewhere on the machine — with this PR applied, the scan returns `matched=true` in 11ms at `root=/` because some unrelated `dbt_project.yml` exists on disk. That is a correctness question for whoever owns project identity, and deliberately left out of this change.

### Checklist

https://claude.ai/code/session_018fJ3X7pcGT4R9yzjsJnqsV

---
## Summary by cubic
Fixes skill auto-load so `applyPaths` globs stop scanning the whole tree to answer "does any file match" and stop matching against the whole filesystem outside a git repo.

- Adds `Glob.exists` to `packages/core`, which abandons the walk at the first match; startup outside a git repo drops from ~51s to ~7s, and startup inside a repo is unchanged.
- Falls back to the session directory when `Instance.worktree` is `/` and there is no VCS — the no-project sentinel — so an empty non-git directory no longer auto-loads the dbt skills from an unrelated `dbt_project.yml` elsewhere on the machine; a git repo genuinely rooted at `/` keeps scanning from its root.
- Rejected `ignore: Glob.DEFAULT_IGNORE`; it was slower (61–82s) because paths outside a git repo rarely get pruned.
- Tests cover `Glob.exists` agreement with `scan`, `include`/`ignore` handling, missing directories, and the `autoLoadScanRoot` sentinel and VCS distinction.

<sup>Written for commit 1f1e23abc6b89eb7e04389d932aee34633d4cbde. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1213?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added a fast way to check whether files match a glob pattern.
  - File matching now supports efficient exclusions and stops once a match is found.

- **Performance**
  - Improved worktree checks by avoiding unnecessary full-directory scans.

- **Bug Fixes**
  - Glob-based checks now consistently honor inclusion and exclusion options.
  - Scans now use the session directory when no project worktree is available.
  - Scans correctly preserve the filesystem root when a Git repository is genuinely rooted there.


sahrizvi:
## Code Review

**Verdict**: Approve
**Major**: 0 · **Minor**: 3 · **Nit**: 4

Small, cleanly-scoped performance fix (66 insertions / 9 deletions, 3 files) with unusually
strong empirical backing: the PR description ships three-run before/after benchmarks in-repo
and out-of-repo, states and measures a rejected alternative, and traces the root cause
(`Instance.worktree` resolving to `/` outside a git repo) precisely.

---

### Minor

**m1. No test pins the actual behavioral guarantee — only the return value**
`packages/core/src/util/glob.ts:69-72`, `packages/core/test/util/glob.test.ts:126-161`

The whole point of `Glob.exists` is that `globIterate` **abandons the walk** after the first
match, not just that it *returns* `true` early. The four new tests only assert boolean
agreement with `scan()`, option-forwarding, and edge cases — none of them would fail if
`Glob.exists` were later reimplemented as `(await Glob.scan(...)).length > 0`, silently
reintroducing the exact startup regression this PR fixes.

The underlying behavior is correct on the current `glob@13.0.5` — confirmed by tracing the
`glob` source and by instrumenting `fs.readdir` calls before/after breaking out of the
iterator (reads stop immediately once the consumer stops pulling, consistent with the PR's
measured 51s → 6-9s improvement). But that guarantee currently isn't pinned by anything in
the test suite that would catch a future `glob` upgrade changing it.

**Fix**: add a regression test that counts `readdir` calls (via `glob`'s injectable `fs`
option) or measures wall-clock on a tree with an early match and a large, unvisited sibling
subtree, and assert the walk actually stops.

**m2. `Glob.exists` comment doesn't state the miss-path cost**
`packages/core/src/util/glob.ts:65-68`

"`globIterate` yields lazily, so this abandons the walk as soon as one path matches" is
accurate but incomplete: when *nothing* matches, `Glob.exists` still walks the entire tree,
identical cost to `scan`. The win is strictly on the match path. A reader skimming the
comment could reasonably assume `exists` is cheap unconditionally.

**Fix**: append "When nothing matches it still walks the full tree — the win is on the match
path, not the miss path."

**m3. Inline benchmark numbers in the caller comment will rot**
`packages/opencode/src/session/system.ts:249-253`

The comment cites specific measurements ("~45s of a ~51s startup"). These are
environment/machine/glob-version specific and will drift from reality; the durable part is
the rationale ("`scan` walks the whole tree before the caller can look"), and the numbers
already live permanently in the PR/commit history.

**Fix**: trim the comment to the rationale; leave exact figures to the commit message.

---

### Nit

**n1.** `packages/core/test/util/glob.test.ts:127-133` — the "agrees with scan()" test only
exercises `{ cwd, absolute }`; `include`, `ignore`, `dot`, and `symlink` are each pinned
individually elsewhere but never together. Low risk since `toGlobOptions` is shared between
`scan`/`exists`, but a combined-options case would be cheap insurance against the two
diverging.

**n2.** `packages/core/test/util/glob.test.ts:127-132` — the test name `"agrees with scan() on
whether anything matched"` verifies boolean agreement only; it doesn't check that `exists`'s
`absolute: true` paths are correct, since it never receives any paths. Fine as shorthand, but
a pedantic reader could misread it as asserting path-level equivalence.

**n3.** `packages/core/src/util/glob.ts:69,75,79` — the `Options` default is duplicated across
`exists`/`scan`/`scanSync`. Pre-existing pattern, not introduced by this PR; three one-line
call sites, not worth extracting.

**n4.** `packages/core/test/util/glob.test.ts:128` — the test pattern `**/*.ts` matches many
files in a typical project tree, making the test slower than necessary without adding
coverage over a narrower, unique pattern.

---

### Considered and not applicable

- **Adding `Glob.DEFAULT_IGNORE` to the `anyMatchInWorktree` caller for a further speedup** —
  already tried and measured by the PR author: the description's "Rejected alternative"
  section shows this makes things *slower* (61-82s vs. ~51s baseline). Outside a repo the tree
  isn't dependency-dominated, so almost nothing gets pruned and every candidate path pays 12
  extra `minimatch` tests.
- **Expanding the fix to other `Glob.scan(...).length > 0` call sites** (`skill.ts`,
  `truncation.ts`, `filesystem.ts`'s `globUp`) — checked each: `globUp` isn't a valid target at
  all, since it accumulates and returns the actual match list across directory levels, so
  callers need the paths, not a boolean; `exists()` there would add a redundant walk, not
  remove one. The `skill.ts`/`truncation.ts` sites still need the full match list on the
  common (non-empty) path, and unlike `anyMatchInWorktree` (measured: 45 of 51 startup
  seconds, worktree can resolve to `/`), there's no evidence these are hot. Good follow-up PR
  material, not a gap in this one — consistent with this PR's own scope discipline in leaving
  `mcp/discover.ts`'s `Glob.scan` (which genuinely needs the list) untouched.

### Positive observations

- PR description is exemplary: quantified benchmarks (in-repo and out-of-repo, 3 runs each), a
  stated-and-measured rejected alternative, and a root-cause trace to `Instance.worktree`
  resolving to `/` outside a git repo — while correctly keeping that root cause out of this
  diff's scope.
- `Glob.exists` reuses `toGlobOptions`, so its `ignore`/`include`/`dot`/`symlink` semantics
  cannot drift from `scan`'s — exactly what the new tests are built to catch.
- Correctly left alone every other `Glob.scan` call site that genuinely needs the full match
  list (`mcp/discover.ts`, `globUp`, config loaders) — no scope creep.
- Test fixtures are properly isolated: the "prunes with ignore" test uses its own `mkdtemp`
  rather than piggybacking on the shared `root` fixture, avoiding a false-positive pass.
- The `system.ts:249-253` comment is a good model of "comment the why," modulo m3.

### Missing tests

1. A regression test that pins walk-abandonment as *behavior*, not just return value (m1).
2. Combined-options agreement test for `exists` vs. `scan` (n1).
3. `dot` and `symlink` option parity for `Glob.exists` specifically — nice-to-have, not
   blocking; both are exercised indirectly via shared `toGlobOptions`.
4. An integration test proving `SystemPrompt.skills()` still auto-loads on match and skips on
   no-match — pre-existing gap, not introduced by this PR.

ralphstodomingo:
@codex review — scoped round. Review this head against the numbered claims below and report only a reproducible trace that violates one. Style and prose suggestions are not findings.

C1: `Glob.exists` returns exactly the same boolean as the previous `scan(...).length > 0` for the same pattern and options — same matches, same `include`/`ignore`/cwd semantics, no case where one is true and the other false.
C2: `Glob.exists` abandons the walk at the first match, so it cannot walk the whole tree when an early match exists.
C3: Options forwarding is complete: every option the old `scan` call relied on is passed through, and dropping any of them changes behaviour a test would catch.
C4: The autoload path answers "does any file match" only — no caller depends on the count, the list, or the ordering of matches that `scan` previously produced.
C5: Behaviour is unchanged when the pattern matches nothing, when the directory does not exist, and when the worktree resolves to a root path.

A counterexample is a concrete path — pattern, cwd, filesystem state — where this head differs from the previous implementation or from the claim. A round with no claim violation ends review.

packages/opencode/src/session/system.ts
```diff
@@ -245,21 +245,41 @@ export namespace SystemPrompt {
...
     // catches the file no matter how deep the user's cwd is.
     // Errors propagate to the caller's try/catch (collectAutoLoadedSkills)
     // so the warning log there actually fires.
-    const root = Instance.worktree
+    // `Glob.exists` rather than `scan(...).length > 0`: this only needs to know whether any
+    // file matches, and `scan` walks the whole tree before the caller can look. That cost is
+    // paid once per `applyPaths` skill — two ship builtin — and the root is the worktree, which
+    // is `/` for a directory outside any git repo. Measured from such a directory, the two
+    // scans were ~45s of a ~51s startup, all of it before the first token.
+    const root = autoLoadScanRoot(Instance.worktree, Instance.directory)
```
ralphstodomingo: **The comment three lines above this is now false for the case this line was added to handle.**

```
// Search from worktree root so a skill that wants `dbt_project.yml`
// catches the file no matter how deep the user's cwd is.
```

That is exactly what stops being true outside a git repo. `autoLoadScanRoot` falls back to `Instance.directory`, so a match now has to sit at or below the cwd — a `dbt_project.yml` one level up no longer auto-loads its skill when the session starts from a nested directory.

I think the change is right and the old behaviour was the bug: matching from `/` meant an empty directory picked up the dbt skills from any `dbt_project.yml` anywhere on the machine, which your `autoLoadScanRoot` docstring says plainly. But the two comments now contradict each other in the same function, and the older one is the one a reader hits first.

This is also the substance of cubic's P2. I would characterise it differently than cubic does, though: it reads the change as a regression ("silently stops auto-loading"), when the state it narrows from was itself broken. Neither `/` nor the raw cwd is *correct* for a project that has no root — the PR trades over-matching for under-matching, deliberately, and that is a defensible call. It just needs to be a stated one rather than one a future reader has to reconstruct from two disagreeing comments.

Cheapest fix is to rewrite the older comment to say what the root actually is now and why it differs for non-git projects. If you want cubic's suggestion (walk upward to a discovered project root) that is a separate change and a bigger one — worth its own issue rather than growing this PR.

ralphstodomingo (dismissed):
Reviewed at `fc4b78ce8`. This is a good fix and the evidence behind it is better than most — measurements against a floor, an explicitly rejected alternative with its own numbers, and mutation testing.

I verified the parts that carry the claims rather than reading them:

- **`Glob.exists` is a faithful swap.** It reuses `toGlobOptions`, so `include`, `ignore`, `dot`, `symlink` and cwd resolve exactly as they did for `scan`, and the early exit is the whole of the change in behaviour.
- **The mutation claim is exact.** Dropping the options forwarding fails precisely two tests — `prunes with ignore, like scan` and one sibling — which is what the description says.
- **Tests pass here:** `packages/core/test/util/glob.test.ts` 12/12, `packages/opencode/test/session/autoload-scan-root.test.ts` 3/3.

The rejected-alternative section is the part I would keep in future PRs. "Passing `ignore: Glob.DEFAULT_IGNORE` makes it slower, 61–82s against a ~51s baseline, because outside a repo the tree is not dependency-dominated" is a real finding in its own right, and it is the kind of thing that otherwise gets re-proposed in review by someone reasoning from the general case.

Approving. Two things I would still do before merge, neither of them a defect in the code:

**1. The description does not mention the scan-root change.** The PR reads as a pure performance fix — `scan(...).length > 0` becoming an early-exit existence check. It also changes *which files match*, via `autoLoadScanRoot` falling back to the session directory when the worktree resolves to `/`. That is arguably the more consequential half: it stops an empty directory auto-loading the dbt skills off any `dbt_project.yml` on the machine. The verification table measures only wall-clock, so nothing in the description covers the semantics that moved. Worth a paragraph — the change is a good one and currently has to be discovered from the code.

**2. A retained comment now contradicts the new behaviour.** Detail inline on `system.ts`. Same substance as cubic's P2, framed differently.

Process notes, not review items:

- No `Closes #N` in the description. Per the repo's issue-first policy this PR wants a linked issue, and it is the kind that gets closed unreviewed without one.
- `packages/opencode/src/provider/models-snapshot.ts` is a regenerated build artifact (`requesty` → `subconscious` in the provider list) unrelated to this fix. Harmless, but it makes the diff look like it touches the provider layer.
- coderabbit's nested `altimate_change` marker note is a style nit against the repo's marker convention; I would record it rather than fix it here.
- Codex was summoned scoped to the swap-equivalence claims and had not returned as of 07:00Z, so the above is my read plus cubic's.

---

### Appendix — complexity delta (altimate-code#1213)

`e6d3817146` → `fc4b78ce86`, functions touched by this diff only. Cognitive ≈ how hard it is to review (sonar rules); CCN ≈ branch count. Advisory — pre-existing complexity is not counted against this change.

✅ No touched function changed in complexity (2 touched, 1 new, all under 10).

saravmajestic (approved):

PR #1266, opened 2026-09-08
feat: expose Altimate Base registration over HTTP for non-TUI hosts

saravmajestic:
### Issue for this PR

Closes # _(no tracking issue — raised from the VS Code extension side, companion PR: AltimateAI/vscode-altimate-mcp-server#460)_

### Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation

### What does this PR do?

Lets a non-terminal host register Altimate Base.

`registerAfterConsent` only acts on a token armed through a module-private authority, and the armer is claimable once per process. Previously only the TUI worker claimed it, so the VS Code extension — which talks HTTP to `altimate serve` — could never mint a Base credential: the model appears in `GET /provider`'s `all` but never in `connected`, and selecting it failed the next prompt with `model altimate-base not found`.

Two routes:

- `GET /altimate/base/disclosure` — read-only. Returns the consent text, the picker hint, `registered`, and the SHA-256 the client must echo back. It arms nothing.
- `POST /altimate/base/register` — checks the echoed hash, then mints, arms and redeems the consent token in one operation, registers, and calls `Instance.disposeAll()`. Returns `staleProviders: true` if that disposal fails, since the credential is written but provider lists may be stale.

**What the hash does and does not prove.** It makes *"the caller holds the current disclosure text"* checkable — a client rendering superseded wording cannot register people against text they were never shown, so a copy change like #1268 actually takes effect. It does **not** prove a human read anything: any caller can `GET` the disclosure and echo the hash. Over HTTP, "the user saw this" remains an assertion by the caller. (An earlier version of this description claimed otherwise; that was wrong and is retracted in the comments below.)

`disposeAll()` rather than `dispose()`: the Base credential is one global file, but `Instance.dispose()` evicts only `cache.delete(Instance.directory)` — the directory the request happened to carry — so a multi-root workspace or a second window kept a provider list in which Base was still disconnected.

The gate is injected by the entrypoint (`cli/cmd/serve.ts`) rather than claimed in the route, because `issueArmer()` throws on a second claim and the TUI worker already holds it while serving HTTP from the same process. A host that injects nothing — the TUI worker — gets `501` rather than a second surface that could bypass its own dialog. `capability.ts` owns the canonical statement of who may claim it, and a test enforces that list against the source.

Registration is also refused when an `Origin` header is present and no server password is set: a page on a CORS-allowed origin can reach this port without being a local process, which is a different reachability class from "can already execute tools here". Native clients send no `Origin`. Note this does **not** stop a non-browser local caller; requiring a password unconditionally was rejected because the extension never sets `OPENCODE_SERVER_PASSWORD`, so it would disable the feature for its whole audience.

The disclosure text has one definition in `packages/core`, since `packages/tui` and `packages/opencode` cannot import each other. #1268's copy change is merged in and reaches both surfaces through it.

Three new/changed suites, **20 tests**:

- `test/server/altimate-base-registration.test.ts` (11) — both `501`-when-no-gate paths, `GET` arming nothing, mint/arm/redeem token identity, stale-hash refusal without touching the gate, case-insensitive hash, the `403` browser refusal, failure-taxonomy passthrough, validator rejection, `provide()` single-shot.
- `test/altimate/altimate-base-disclosure-claims.test.ts` (6) — the gate carries logging / model use / don't-send-secrets / rate limits whatever the wording; the docs remain a superset; the known "may be logged" vs "are logged" deviation is pinned so it cannot drift further silently.
- `test/altimate/altimate-base-armer-callsites.test.ts` (3) — the `issueArmer()` call-site set matches the canonical docstring, so adding a claimer fails here instead of quietly falsifying prose.

`packages/tui` Base dialog suite: 8, now including one that asserts **which** option is default by cursor position (it previously checked only that "(default)" appeared somewhere, so it passed unchanged through an inversion of the default and back) and one covering Return declining.

`bun turbo typecheck` clean across 13 packages.

End-to-end against a running `altimate serve` pointed at a local self-signed stub gateway: forged, malformed and replayed tokens all rejected; a valid request registers and `altimate-free` joins `connected` only after disposal. Also end-to-end in code-server against the production gateway via the companion PR, including a deliberate 45-second pause before accepting — which the first revision of this PR failed, because it armed the token at disclosure-fetch time against a 30s TTL.

### Screenshots / recordings

Server-side only. The UI is in the companion PR.

### Checklist

## Summary by CodeRabbit

- **User Experience**
  - Updated Altimate Base consent wording to clarify that usage can be rate limited.
  - Removed persistent per-installation identifier details from the consent prompt.
  - Standardized provider and model picker subtitles to “free · no signup · rate limited.”

- **Registration**
  - Improved registration handling with clearer responses when unavailable or when consent wording is outdated.
  - Registration cleanup is now more reliable after successful activation.

- **Bug Fixes**
  - Fixed onboarding behavior so declining Altimate Base consent does not trigger registration.

### DIFF
--- a/packages/opencode/src/altimate/free/host.ts
+++ b/packages/opencode/src/altimate/free/host.ts
@@ -22,11 +22,17 @@ let registration: Registration | undefined
 
 /**
  * Install the process's registration gate. Called once by the entrypoint, before the server starts
- * accepting requests. A second call replaces the gate rather than throwing: the capability itself is
- * already single-claim, so there is nothing to protect here, and throwing would only turn a
- * misordered startup into a crash.
+ * accepting requests.
+ *
+ * Single-shot, matching every other capability in this area: a second call throws rather than
+ * silently replacing the gate. The earlier last-write-wins behaviour let any in-process caller swap
+ * the gate out from under the routes after `serve` installed the real one. That was never a
+ * privilege escalation — such code is already trusted and still cannot forge a token the private
+ * authority accepts — but it was a weaker invariant than `issueArmer()`/`issueRedeemer()` next door,
+ * for no benefit.
  */
 export function provide(value: Registration): void {
+  if (registration) throw new Error("Altimate Base registration gate already provided for this process")
   registration = value
 }
 
--- a/packages/opencode/src/altimate/free/host.ts
+++ b/packages/opencode/src/altimate/free/host.ts
@@ -12,9 +12,26 @@
 //
 // `altimate serve` provides a gate. The TUI worker deliberately does NOT — the TUI owns its own
 // disclosure dialog, and a second registration surface inside that process would let a caller
-// register without the dialog ever being shown. Consumers must treat `undefined` as "this host
-// cannot register Altimate Base" and refuse, exactly as the TUI's provider picker does.
-import type { createRegistrationConsentGate } from "./consent"
+// register without the dialog ever being shown. Consumers must treat "cannot register" as final
+// and refuse, exactly as the TUI's provider picker does.
+//
+// The gate itself is NEVER handed back out. An earlier revision exposed `current()`, which returned
+// the whole gate — including `setToken` (closing over the real armer) and `register` (redeeming
+// against the real authority) — so any importer held a raw mint primitive and could register
+// without going near the disclosure, in any order it liked. The check now happens *inside* this
+// module, in the same call that mints, arms and redeems: there is no ordering for a caller to get
+// wrong and no primitive to borrow.
+//
+// What this is NOT: a trust boundary against in-process code. `registerWithAcceptedDisclosure` is
+// exported, and the hash it demands is a SHA-256 of public text that any caller can recompute via
+// `FreeTierConsent.disclosureHash()`. In-process code can therefore still cause a registration —
+// it simply cannot do so while bypassing the documented precondition, and there is now one
+// audited path instead of a capability handed to every importer. The real boundary is the process:
+// anything running here is already trusted to execute tools. What this closes is accidental
+// misuse and the drift that comes from re-implementing the check at each call site.
+import { randomBytes } from "node:crypto"
+import type { createRegistrationConsentGate, RegistrationResult } from "./consent"
+import { FreeTierConsent } from "./consent"
 
 export type Registration = ReturnType<typeof createRegistrationConsentGate>
 
@@ -36,9 +53,38 @@ export function provide(value: Registration): void {
   registration = value
 }
 
-/** The host-injected gate, or `undefined` when this host cannot register Altimate Base. */
-export function current(): Registration | undefined {
-  return registration
+/** Whether this host can register Altimate Base at all, i.e. whether an entrypoint provided a gate. */
+export function canRegister(): boolean {
+  return registration !== undefined
+}
+
+export type RegisterOutcome =
+  /** No gate was provided; this host cannot register Altimate Base. */
+  | { kind: "unavailable" }
+  /** The caller echoed a hash that is not the current disclosure's. */
+  | { kind: "staleDisclosure" }
+  /** The gate ran; `result` carries its success or its classified failure. */
+  | { kind: "done"; result: RegistrationResult }
+
+/**
+ * Verify the caller accepted the current disclosure text, then mint, arm and redeem in one step.
+ *
+ * The hash comparison lives here rather than in the caller so that holding the current disclosure
+ * text is a precondition of minting, not a convention the caller is trusted to follow. It is a
+ * **text-version agreement, not proof of consent**: it establishes that the caller holds the
+ * current wording, so a client still rendering superseded text cannot register people against text
+ * they were never shown. Any caller can fetch the disclosure and echo the hash, so "a human read
+ * this" remains an assertion by the caller.
+ */
+export async function registerWithAcceptedDisclosure(acceptedDisclosureSha256: string): Promise<RegisterOutcome> {
+  const gate = registration
+  if (!gate) return { kind: "unavailable" }
+  if (acceptedDisclosureSha256.toLowerCase() !== FreeTierConsent.disclosureHash()) {
+    return { kind: "staleDisclosure" }
+  }
+  const token = randomBytes(32).toString("hex")
+  gate.setToken({ token })
+  return { kind: "done", result: await gate.register({ token }) }
 }
 
 export * as FreeTierHost from "./host"
--- a/packages/opencode/src/altimate/free/host.ts
+++ b/packages/opencode/src/altimate/free/host.ts
@@ -1,0 +1,39 @@
+// altimate_change start — host-injected Altimate Base registration for non-TUI entrypoints.
+//
+// `FreeTierCapability.issueArmer()` is claimable exactly once per process and throws on a second
+// call, so an HTTP route cannot claim one for itself: the TUI worker already claims it at boot for
+// its own RPC gate, and that worker also serves HTTP from the same process. A route-level claim
+// would therefore break the TUI worker the moment the routes module loaded — and any test that
+// imported both the server and the Base test harness.
+//
+// Instead the entrypoint that owns the process claims the capability once and hands the resulting
+// gate here. This is the same shape the TUI already uses for the same operation
+// (`packages/tui/src/context/altimate-base-consent.tsx`): the host injects, the consumer checks.
+//
+// `altimate serve` provides a gate. The TUI worker deliberately does NOT — the TUI owns its own
+// disclosure dialog, and a second registration surface inside that process would let a caller
+// register without the dialog ever being shown. Consumers must treat `undefined` as "this host
+// cannot register Altimate Base" and refuse, exactly as the TUI's provider picker does.
+import type { createRegistrationConsentGate } from "./consent"
+
+export type Registration = ReturnType<typeof createRegistrationConsentGate>
+
+let registration: Registration | undefined
+
+/**
+ * Install the process's registration gate. Called once by the entrypoint, before the server starts
+ * accepting requests. A second call replaces the gate rather than throwing: the capability itself is
+ * already single-claim, so there is nothing to protect here, and throwing would only turn a
+ * misordered startup into a crash.
+ */
+export function provide(value: Registration): void {
+  registration = value
+}
+
+/** The host-injected gate, or `undefined` when this host cannot register Altimate Base. */
+export function current(): Registration | undefined {
+  return registration
+}
+
+export * as FreeTierHost from "./host"
+// altimate_change end
--- a/packages/opencode/src/server/server.ts
+++ b/packages/opencode/src/server/server.ts
@@ -40,9 +40,12 @@ import { readMcpEntryFromDisk } from "../mcp/config"
 import { resolveConfigPath } from "../mcp/config"
 import { enhancePrompt, isAutoEnhanceEnabled } from "../altimate/enhance-prompt"
 // altimate_change - Altimate Base disclosure + consent-gated registration for HTTP hosts
-import { randomBytes } from "node:crypto"
 import { FreeTier } from "../altimate/free/client"
 import { FreeTierConsent } from "../altimate/free/consent"
+// altimate_change start — Altimate Base registration must invalidate BOTH instance registries.
+import { InstanceStore } from "@/project/instance-store"
+import { AppRuntime } from "@/effect/app-runtime"
+// altimate_change end
 import { FreeTierHost } from "../altimate/free/host"
 // altimate_change end
 import { FileRoutes } from "./routes/file"
@@ -724,7 +727,7 @@ export namespace Server {
           },
         }),
         async (c) => {
-          if (!FreeTierHost.current()) {
+          if (!FreeTierHost.canRegister()) {
             return c.json({ error: "This host cannot register Altimate Base." }, 501)
           }
           const registered = await FreeTier.isRegistered().catch((error) => {
@@ -744,7 +747,7 @@ export namespace Server {
         describeRoute({
           summary: "Register Altimate Base after consent",
           description:
-            "Mints the managed Altimate Base credential. The caller must echo the SHA-256 of the disclosure it displayed, which is verified against the canonical text, so a caller that never fetched the current disclosure cannot register. Disposes the instance on success so the provider loader re-reads the new credential.",
+            "Mints the managed Altimate Base credential. The caller must echo the SHA-256 of the disclosure it displayed, which is verified against the canonical text, so a caller that never fetched the current disclosure cannot register. On success this disposes EVERY cached instance in the process — both registries — so provider loaders re-read the new credential. That is deliberately process-wide because the credential is a single global file, and it is disruptive: instance-scoped state elsewhere on this server (sessions, LSPs, PTYs, MCP connections, file watchers) is torn down and re-created, and `server.instance.disposed` is emitted for each. `staleProviders: true` in the response means the credential was written but at least one registry could not be invalidated, so provider lists may still show Altimate Base as disconnected.",
           operationId: "altimateBase.register",
           responses: {
             200: {
@@ -777,8 +780,7 @@ export namespace Server {
         }),
         validator("json", z.object({ acceptedDisclosureSha256: z.string() })),
         async (c) => {
-          const gate = FreeTierHost.current()
-          if (!gate) {
+          if (!FreeTierHost.canRegister()) {
             return c.json({ error: "This host cannot register Altimate Base." }, 501)
           }
 
@@ -801,12 +803,14 @@ export namespace Server {
           }
 
           const { acceptedDisclosureSha256 } = c.req.valid("json")
-          if (acceptedDisclosureSha256.toLowerCase() !== FreeTierConsent.disclosureHash()) {
-            // A text-version agreement, NOT proof of consent. It establishes that the caller holds
-            // the current disclosure, so a client rendering superseded wording cannot register
-            // people against text they were never shown. It does not establish that a human read
-            // anything — any caller can GET the disclosure and echo the hash back. "A human saw
-            // this" remains an assertion by the caller, exactly as it was.
+          // Hash verification, mint, arm and redeem all happen inside `FreeTierHost`, so this route
+          // cannot mint without checking and no other module can borrow the mint primitive. See
+          // `altimate/free/host.ts` for why the gate is never handed back out.
+          const attempt = await FreeTierHost.registerWithAcceptedDisclosure(acceptedDisclosureSha256)
+          if (attempt.kind === "unavailable") {
+            return c.json({ error: "This host cannot register Altimate Base." }, 501)
+          }
+          if (attempt.kind === "staleDisclosure") {
             return c.json(
               {
                 ok: false as const,
@@ -816,12 +820,7 @@ export namespace Server {
               200,
             )
           }
-
-          // Mint, arm and redeem in one operation, mirroring the TUI's accept handler. Nothing
-          // outside this handler can obtain a token that the private authority will accept.
-          const token = randomBytes(32).toString("hex")
-          gate.setToken({ token })
-          const outcome = await gate.register({ token })
+          const outcome = attempt.result
 
           // The provider loader caches its credential read, so a freshly registered Base stays out
           // of `/provider`'s `connected` list until the instance cache is dropped. Doing it here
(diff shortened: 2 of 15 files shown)


anandgupta42 (commented):
@saravmajestic — deep review below. Net: **sound design, safe to merge after two fixes** (one real TTL correctness risk, one test-coverage gap). The security narrowing is real but consciously made and correctly bounded.

I traced the full mechanism first (`ConsentCapabilityStore`: 64-hex tokens, 30s TTL, single-use `consume`, ≤16 pending, unforgeable once-per-process armer/redeemer).

## What's done right (non-trivial)
- **Unforgeability is preserved.** The route can't mint an accepted token out of thin air — it arms through the *same* private `productionAuthority` via the injected `issueArmer()` closure, and `registerAfterConsent` redeems via `issueRedeemer()`. A forged/self-armed token still fails `consume()`. The PR weakens *only* "a human saw the text," not "the token is authentic." That distinction is the crux and it's right.
- **Host-injection instead of a route-level claim** is correct — `issueArmer()` throws on a second claim and the TUI worker already holds it in-process. Claiming at the `serve` entrypoint + `501` from the TUI worker (rather than a second bypass surface) is the right shape.
- Single definition of the disclosure in `core`; honest, specific security note in the PR body.

## Issues to address

**[P1 — correctness/UX] The 30s TTL now spans human read-and-accept time.**
In the TUI, arm→redeem is instantaneous (armed *at* accept). Here, `GET /disclosure` arms the token at *disclosure-show* time and hands it to the extension; the human then reads and clicks accept, and only then does `POST /register` redeem. `DEFAULT_TTL_MS = 30_000`, so a user who reads the privacy text for >30s gets `consent expired` and registration fails. The E2E likely passed because the click was fast. Fix options: raise the TTL for this flow, arm at accept-time (a small arm step just before register), or have the extension re-`GET` + retry on expiry. This is the one I'd block on.

**[P1 — test coverage] The security-critical properties are only verified by manual curl, not automated tests.**
Please add server route tests for: `501` when no gate is injected (the TUI-worker bypass guard), forged/expired token → `ok:false`, single-use enforced (second redeem fails), and the outcome taxonomy. These are exactly the invariants a future refactor could silently break; the manual curl in the PR body isn't a regression guard.

**[P2] `GET` has a side effect (arms a token).**
A side-effecting GET is a REST smell — prefetch/CSRF can churn token slots (harmless alone, but burns 1 of the 16-token bound per call). Consider issuing the token on an explicit step. Minor: the route comment frames the token as "armed here and nowhere else, single-use" — accurate on single-use, but the store holds up to 16 pending (`maxPending`), evicting FIFO, not one slot.

**[P2] `FreeTierHost.provide()` is exported and last-writer-wins.**
Any importer can replace the gate. It can't inject a *working* bypass (no second armer available; redeem still enforces consent), so worst case is DoS (replace with `undefined`/broken → `501`/failures). Low severity, but a stray/duplicate call silently swaps the process's registration gate — worth a guard or a comment that only the entrypoint may call it.

**[P2] `console.error` in `serve.ts`** vs `log.warn` in the route — use the project logger for consistency.

## Heads-up: conflicts with open PR #1268 (disclosure-text change)
PR #1268 (per a product decision today) **removes the `Logs are linked to a persistent per-installation identifier.` line and softens `Usage is rate limited` → `Usage can be rate limited`** in `ALTIMATE_BASE_DISCLOSURE`. This PR moves that constant into `packages/core/.../altimate-base-disclosure.ts` with the *old* wording. They'll conflict:
- Whichever merges second must reconcile.
- If this PR lands first, that copy change must be re-applied to the new `core` file (the TUI constant becomes a re-export), and the core file's comment — which asserts it "discloses that requests are linkable across launches" — must be updated, since that becomes false once the per-install line is removed.

I'll handle the rebase on the other PR once we pick a merge order.

## Security narrowing — accept, with two conditions
Turning "disclosure shown" into a caller assertion is acceptable for a first-party extension, given the token is authentic-only and this isn't a new privilege boundary. Two conditions:
1. The companion extension must actually render the disclosure before POST — verify that in its PR; this PR's guarantee leans entirely on it.
2. Consider whether a consent-minting route should require server auth (refuse when `OPENCODE_SERVER_PASSWORD` is unset). Today an unsecured `serve` lets any local caller silently register + opt the user into logging — "can already run tools" covers exec, not the silent data-logging opt-in.

saravmajestic:
Thanks — this was a genuinely useful review. Both P1s and all four P2s are addressed, plus the #1268 reconciliation you flagged. Independently, a three-model panel (Gemini 3.1 Pro, GPT 5.2 Codex, Claude) hit your P1 TTL finding and your P1 test-coverage finding as its own top two, which was a good confirmation signal.

**P1 — TTL spanning human read time.** Fixed by your second option, arming at accept time. `GET /altimate/base/disclosure` is now read-only and returns no token at all; `POST /altimate/base/register` mints, arms and redeems in one operation, so no TTL can elapse mid-flow. I went a step further than a bare arm-before-register: the client echoes back the SHA-256 of the disclosure it displayed and the server verifies it against the canonical text. That converts "a human saw the text" from a pure caller assertion into something checkable — a client that never fetched the disclosure, or one showing superseded wording, cannot register. It also happens to close your P2 GET-side-effect point, and it removes the extension's need to know about instance caching, since the route now disposes itself.

You were right that the E2E passed on speed. Playwright clicked in about two seconds.

**P1 — test coverage.** `packages/opencode/test/server/altimate-base-registration.test.ts`, 11 tests, covering exactly the invariants you listed: `501` on both routes when no gate is injected (the TUI-worker bypass guard), stale/forged hash rejected without the gate being touched, mint/arm/redeem token identity, the failure taxonomy passing through intact, plus the `403` browser refusal, validator rejection, and `provide()` single-shot.

**P2 — `provide()` last-writer-wins.** Now single-shot, throws on reuse, matching `issueArmer`/`issueRedeemer`. Your DoS-only severity read was correct, and that is exactly why it wasn't worth leaving as-is for no benefit.

**P2 — the maxPending comment.** You were right that it misdescribed the bound as one slot when the store holds up to 16 and evicts FIFO. The comment is gone along with the token it described.

**P2 — `console.error` in `serve.ts`.** Now `Log.create({ service: "serve" })`, matching `serve-upgrade-check.ts`.

**#1268.** Merged. Kept this branch's structure (the TUI re-exports the shared constant) and adopted main's new wording in the `core` definition, so that copy change now lands on the HTTP route too — which is the argument for one definition. I also fixed the `core` comment you specifically predicted would become false: it no longer claims the gate discloses cross-launch linkage, and points at the docs "Data handling" note instead. The route tests reference `FreeTierConsent.DISCLOSURE` rather than a literal, so they picked the new text up unchanged.

**Your two conditions.**

1. *The extension must render the disclosure before POST.* It does — the card fetches, renders, and only then can accept fire. And this no longer rests on trust: the hash check means a POST without a preceding fetch of the current text fails. Worth reviewing in #460 all the same.

2. *Require server auth on a consent-minting route.* I could not adopt this as literally stated, and I want to be explicit rather than quietly skip it: **the extension never sets `OPENCODE_SERVER_PASSWORD`** (checked across `src/` — nothing sets it), so refusing when it is unset would disable the feature for every extension user, which is the entire audience. What I did instead is refuse when an `Origin` header is present and no password is set. That targets the reachability class that is genuinely new — a page on a CORS-allowed origin can hit this port without being a local process — while leaving native clients, which send no `Origin`, working. A local CLI on an unsecured server can still register, so your concern is narrowed rather than eliminated. If you want the stricter rule, it needs the extension to start passing a password, which is a larger change I'd rather do deliberately than smuggle in here.

Also picked up main's header-timeout fixes (#1260 and the parsing hardening) in the merge, which resolve the `headers timed out after 10000ms` failures seen while testing this.

Re-verified after all of the above: `bun turbo typecheck` clean across 13 packages, new suite 11/11, Base suites 58/58, TUI Base dialog 7/7.

saravmajestic:
Correcting two claims I made earlier on this PR. Both were wrong, and both are now fixed in code as well as in the record.

**1. "Server-side dispose also refreshes other windows attached to the same `serve` process."** False. `Instance.dispose()` only evicts `cache.delete(directory)` for the directory the request happened to carry (`packages/opencode/src/project/instance.ts:141-147`); `disposeAll()` is what iterates every cached instance. So a multi-root workspace, or a second window on one `serve`, kept a provider list in which Base was still disconnected. The route now calls `disposeAll()`, and returns `staleProviders: true` when disposal fails rather than reporting plain success over a stale list.

**2. "The hash check converts 'the user was shown the current disclosure' from an assertion by the caller into something checkable."** Overstated. It makes *"the caller holds the current disclosure text"* checkable — a client rendering superseded wording cannot register, which is worth having and does mean a copy change like #1268 takes effect. But any caller can `GET` the disclosure and echo the hash back; whether a human read it remains an assertion, exactly as @anandgupta42 originally characterised it. `consent.ts` and the route comment now say that plainly.

Relatedly, the claim about who may arm the consent authority had been paraphrased in three files and went stale in two of them when `serve` became a second claimer — including the `client.ts` docstring I had quoted as evidence. `capability.ts` now owns the canonical statement and lists both entrypoints; the others point at it. `test/altimate/altimate-base-armer-callsites.test.ts` enforces that list against the source, so adding a claimer fails there instead of quietly falsifying the prose.

Your two P1s and all four P2s from the first review are addressed. The remaining known deviation is the one you'd expect: the gate says "may be logged" where `providers.md` says "are logged" — pre-existing from #1268, kept per product decision, now pinned by a test so it cannot drift further silently.

packages/opencode/src/altimate/free/host.ts
```diff
@@ -0,0 +1,45 @@
...
+ * authority accepts — but it was a weaker invariant than `issueArmer()`/`issueRedeemer()` next door,
+ * for no benefit.
+ */
+export function provide(value: Registration): void {
+  if (registration) throw new Error("Altimate Base registration gate already provided for this process")
+  registration = value
+}
+
+/** The host-injected gate, or `undefined` when this host cannot register Altimate Base. */
+export function current(): Registration | undefined {
```
sahrizvi: **MAJOR — Security / Design**

`current()` returns the full `Registration` object — including `setToken()`, which closes over the real production armer, and `register()`, which redeems against the real production authority.

Any in-process code that imports `FreeTierHost` (not just this route, and not just code that has passed the disclosure-hash/origin checks in `server.ts`) can do:

```ts
const gate = FreeTierHost.current()
gate?.setToken({ token: "anything" })
await gate?.register({ token: "anything" })
```

...and mint a Base credential without ever going through `GET /altimate/base/disclosure` or the `acceptedDisclosureSha256` check. That defeats the invariant `capability.ts`'s own docstring claims for this armer/redeemer pair — that no in-process code other than the entrypoint that claimed `issueArmer()` can ever produce a token `registerAfterConsent` accepts. Here, the *entrypoint* still claims it correctly (once, in `serve.ts`), but then hands out an object that re-exposes the same power to everyone downstream.

Suggest narrowing what `current()` returns to something that can't be misused as a raw mint primitive — e.g. a single `registerWithDisclosure(acceptedDisclosureSha256: string): Promise<RegistrationResult>` that performs the hash check *inside* `host.ts` itself (mint token → verify hash → arm → redeem, all in one call), rather than handing the caller `setToken`/`register` directly and trusting every future caller to redo the hash check correctly. That would make the HTTP route's own hash check redundant-but-safe instead of load-bearing.

packages/opencode/src/server/server.ts
```diff
@@ -671,6 +676,179 @@ export namespace Server {
...
+          // `disposeAll()`, not `dispose()`: the Base credential is a single global file, but
+          // `Instance.dispose()` only evicts `cache.delete(Instance.directory)` — the directory this
+          // request happened to carry. A multi-root workspace, or several windows against one
+          // `serve`, would keep every other instance's provider list showing Base as disconnected.
+          // Invalidation has to be as wide as the state that changed.
+          //
+          // A failure here leaves the credential written but provider lists stale, so it is reported
+          // rather than swallowed: the client needs to know its picker may be out of date.
+          if (outcome.ok) {
+            const disposed = await Instance.disposeAll().then(
```
sahrizvi: **MAJOR — Logic Error / Design**

Two separate concerns with calling the legacy `Instance.disposeAll()` here:

**1. It may not actually fix the staleness it's meant to fix.** `/api/*` traffic is forwarded via `forwardHttpApiBridge` (`server.ts:204`) *before* the legacy `Instance.provide` middleware runs (`server.ts:~285`), and that typed API path is backed by a separate `InstanceStore` (`project/instance-store.ts`), not the legacy `Instance` registry this call disposes. A directory/window that was only ever touched through `/api/*` won't be in the legacy `Instance` cache, so this call can't invalidate its cached provider state — which is exactly the "other window keeps showing Base as disconnected" case the surrounding comment says this fixes.

**2. It's simultaneously broader than it needs to be.** `Instance.disposeAll()` tears down *every* `Instance.state` entry process-wide — session prompt state, LSPs, PTYs, plugins, schedulers, file watchers, MCP connections — and emits `server.instance.disposed`, closing event streams for every other legacy-backed window on this `serve` instance. So a user registering Base in one window can interrupt active work in every other window attached to the same server, purely to refresh a provider list.

If a full instance-wide reset really is required today, that's a legitimate trade-off worth stating explicitly in the route's `describeRoute` description (right now a client has no way to know a registration call might disrupt unrelated sessions) — but the fix for concern #1 is separate from accepting the blast radius in #2: even accepting the broad reset, it should be reaching `InstanceStore`'s directories too, or the `/api/*`-only case stays broken regardless of how disruptive the workaround is.

sahrizvi (changes requested):
## Multi-model consensus review

Reviewed by Claude + GPT-5.4-Codex + Kimi K2.5 + MiniMax M2.7 + GLM-5.1 + Qwen 3.6 + MiMo V2 Pro (Gemini/Antigravity sat this round out — quota-locked; Kimi crashed mid-review due to host disk space, not a code problem). Two independently-verified MAJOR issues below are requesting changes; everything else is minor polish that doesn't need to block merge.

### Requesting changes on

1. **`FreeTierHost.current()` leaks the raw mint capability to any in-process importer** (inline comment on `host.ts`). This defeats the "only one entrypoint can ever arm the authority" invariant that `capability.ts`'s own docstring claims — any code path that imports `FreeTierHost`, not just the disclosure-gated HTTP route, can mint a Base credential.
2. **`Instance.disposeAll()` targets the wrong/incomplete registry while being simultaneously too broad** (inline comment on `server.ts`). It won't refresh provider state for directories only touched through the typed `/api/*` bridge (a separate `InstanceStore`), while tearing down active sessions/PTYs/MCP connections in every other window on the same `serve` process.

### Minor findings (non-blocking, worth a follow-up)

**Testing**
- No test exercises the `staleProviders: true` branch (when `Instance.disposeAll()` rejects after a successful registration) — `server.ts:839-846`. Flagged independently by 4 of 6 reviewers; every *other* branch in this route has direct coverage, so this is the one gap in an otherwise thorough test file.
- `altimate-base-registration.test.ts` is order-dependent across the whole `bun test` process (the "no gate injected" block must run before the gate is provided). The comment explains why, but a `bun test` config change or file split could silently reorder suites and break it invisibly.
- No test for the "server password set + browser Origin present" combination on `POST /altimate/base/register` — the code's own comment says this combination should be *allowed* (basicAuth already covers it), but nothing asserts that.
- No test for concurrent `POST /altimate/base/register` requests exercising the `inflight` dedup in `client.ts`.

**Design**
- `GET /altimate/base/disclosure` doesn't carry the same Origin check `POST /altimate/base/register` has. Low impact (disclosure text is intentionally public, the hash is derived from public text, and the only other field is a `registered` boolean) — but worth a one-line comment on why it's intentionally omitted, or add the same guard for consistency, since a reviewer will keep re-noticing the asymmetry otherwise.
- `GET .../disclosure`'s `registered` field silently returns `false` on a gateway misconfiguration (`FreeTier.isRegistered().catch(() => false)`), which is indistinguishable from a genuinely-unregistered install. A caller can't tell "not registered" from "couldn't check."

**Nits**
- `disclosureHash()` recomputes SHA-256 on every call rather than memoizing the (constant) result.
- `acceptedDisclosureSha256` validator accepts an empty string, which just falls through to the normal "stale hash" response — harmless, but a `min(64)` on the schema would give a clearer 400 for a client bug.
- A couple of NIT-level polish items across models: a redundant type assertion in the disclosure route's `registered` field, an unsafe type assertion in one test helper, and the OpenAPI description for `GET /disclosure` doesn't mention the `registered` field.

### What's genuinely strong here

- The disclosure text now has a single source of truth (`packages/core/src/altimate-base-disclosure.ts`), closing a copy-drift bug this feature had hit repeatedly across earlier rounds.
- `altimate-base-armer-callsites.test.ts` turns "which entrypoints may claim `issueArmer()`" from a comment into an enforced invariant — genuinely good pattern, worth replicating elsewhere.
- GET is correctly read-only (mint+arm+redeem now happen atomically inside POST) — the PR's own comment explains the prior bug this fixes (TTL starting on disclosure-fetch, not on accept).
- The TUI test fix for "which option is default" (asserting the actual selected row, not just that the word "(default)" appears) plus the new Return-key regression test are good, specific catches.
- Error handling in `createRegistrationConsentGate.register()` is exhaustive — nothing escapes as an unhandled rejection into the new HTTP handler.

---
*Consensus review: Claude, GPT-5.4-Codex, MiniMax M2.7, GLM-5.1, Qwen 3.6, MiMo V2 Pro (6 participants; Kimi K2.5 and Gemini 3.1 Pro did not return usable output this round).*

saravmajestic:
Thanks — both blocking findings are fixed in 3692084, plus the minors that were cheap. Taking them in turn.

## 1. `FreeTierHost.current()` leaked the raw mint capability — fixed

You were right that the entrypoint claimed the armer correctly and then handed the same power to every importer. `current()` is gone. `host.ts` now exposes only:

- `canRegister(): boolean`
- `registerWithAcceptedDisclosure(acceptedDisclosureSha256): Promise<RegisterOutcome>`

which hash-verifies and *then* mints, arms and redeems, all inside the module — your suggested shape. The route can no longer mint without checking, and there is no `setToken`/`register` pair to borrow.

One thing I want to state rather than let the diff imply: **this is not a trust boundary against in-process code.** `registerWithAcceptedDisclosure` is exported and the hash is a SHA-256 of public text that anyone can recompute via `FreeTierConsent.disclosureHash()`, so in-process code can still cause a registration — it just cannot bypass the documented precondition. What actually improved is that there is one audited path instead of a capability handed to every importer, and no per-call-site re-implementation of the check to drift. The module comment now says exactly that, because an earlier draft of it overclaimed the closure and that is the kind of thing this feature keeps getting wrong.

Verified: `POST /altimate/base/register` with a wrong hash is refused **and no credential is written**.

## 2. `Instance.disposeAll()` targeted the wrong/incomplete registry — fixed, with the breadth called out

Your first concern was correct and I could reproduce the reasoning: `/api/*` is forwarded to the typed HttpApi bridge before the legacy `Instance.provide` middleware runs, and that path is backed by `InstanceStore`. A directory reached only through `/api/*` was therefore not in the registry being disposed. Both are now disposed, via the existing `AppRuntime.runPromise(InstanceStore.Service.use(...))` pattern already used by `routes/tui.ts`, and `staleProviders: true` is returned if **either** fails.

Your second concern — that this is simultaneously too broad — I have **not** fixed, deliberately, and I want to be straight about why. A later review round confirmed your point with a specific mechanism: `session/prompt.ts`'s state teardown calls `item.abort.abort()`, so registering in one window aborts in-flight generations and kills PTYs in every other window on that `serve`. The right fix is the targeted one you implied: invalidate `Provider.state` rather than tearing down the instance. `State.invalidate` already exists and `Provider`'s state has no disposer, so it would be safe — but `Provider`'s init is an anonymous arrow, so it needs naming and exporting in `provider.ts` plus a `State.invalidateAll` in `state.ts`. Both are shared hot-path files outside this feature, and the instruction on this PR is to keep it to the feature. So it is going to a separate ticket.

What I did do is take your suggestion to state the trade-off where a client can see it. `describeRoute` now says disposal is process-wide, names what it tears down (sessions, LSPs, PTYs, MCP connections, file watchers), and explains what `staleProviders: true` means.

## Minors

- **Docs/design asymmetry on `GET /disclosure` having no Origin check** — left as-is, and worth stating the reasoning since you predicted a reviewer would keep re-noticing it: the route is read-only, serialises public constants plus a boolean, and performs no I/O beyond one `.catch`-wrapped disk read. Adding the guard would imply it protects something it does not.
- **`client.ts` docstring** — this one was actually wrong and is fixed: it still located the disclosure-hash check "in the register route" after the check moved into `host.ts`.
- **`registered: false` on a gateway misconfiguration being indistinguishable from genuinely unregistered** — agreed, not addressed. The extension side is unaffected in practice because it fails closed either way now, but the ambiguity is real.
- **`disclosureHash()` recomputing, `min(64)` on the validator, the OpenAPI description omitting `registered`** — not addressed; low value against the churn, happy to take them if you would rather they landed here.

## Testing gaps you flagged

Still open, and I would rather say so than quietly leave them: no test covers the `staleProviders: true` branch, the order-dependence of `altimate-base-registration.test.ts` across the whole `bun test` process is unchanged, and there is no test for password-plus-Origin or for concurrent `POST /register` exercising the `inflight` dedup. The disposal branch is now harder to fake than before, since it needs both registries to fail.

Current state: 20 Altimate Base tests and 8 TUI dialog tests pass, no type errors in the changed files. Verified end-to-end against a freshly built binary — stale hash refused without minting, and a successful registration reports `altimate-free` as connected on the very next `/provider` call.

sahrizvi (dismissed):
## Re-review — both blocking findings verified fixed

Verified `3692084` against the tree myself (not just the description): read the diff, ran `bun test test/server/altimate-base-registration.test.ts test/altimate/altimate-base-armer-callsites.test.ts test/altimate/altimate-base-disclosure-claims.test.ts` (20/20 pass), ran the full `test/altimate/` suite (5115 pass, 2 unrelated pre-existing failures in `sample-setup.test.ts` that touch neither this PR's files nor its feature area), and typechecked `packages/opencode` (the only errors are pre-existing, in `dialog-move-session.tsx`, untouched by this PR).

**1. `FreeTierHost.current()` capability leak — fixed.** `current()` is gone. `host.ts` now exposes only `canRegister(): boolean` and `registerWithAcceptedDisclosure(hash): Promise<RegisterOutcome>`, which verifies the hash *and then* mints/arms/redeems, all inside the module. There's no `setToken`/`register` pair left to borrow, so a caller can no longer skip the check by reordering calls. The new module comment is appropriately honest about what this does and doesn't guarantee (not a trust boundary against in-process code — the real boundary is the process itself) rather than overclaiming, which is exactly right.

**2. `Instance.disposeAll()` wrong/incomplete registry — fixed.** The route now disposes both the legacy `Instance` registry and `InstanceStore` (via the existing `AppRuntime.runPromise(InstanceStore.Service.use(...))` pattern already used in `routes/tui.ts` — confirmed this isn't a novel construct), in parallel, and reports `staleProviders: true` if either fails. This closes the "directory only reached through `/api/*` stays stale" gap I flagged.

**On the blast-radius half of finding #2 — deliberately not fixed, and that's fine.** The reply traced the actual mechanism (`session/prompt.ts`'s teardown calls `item.abort.abort()`, so this does interrupt other windows' in-flight generations and PTYs) and correctly scoped the real fix (`Provider.state` invalidation, not full instance disposal) as touching shared hot-path files outside this feature — deferred to a separate ticket, with the trade-off now stated in the route's own `describeRoute` description so API consumers aren't surprised by it. That's a legitimate scope boundary, not a dodge.

**Remaining open items — all non-blocking, all disclosed rather than hidden:** the `staleProviders` test gap, test order-dependence, no password+Origin test, no concurrent-registration test, GET disclosure's missing Origin check (reasoned: read-only, no I/O, public data only), and the `registered: false`-on-misconfig ambiguity. None of these were silently dropped — the author's reply names each one explicitly as still open. Worth a follow-up PR for the test gaps given how much of the rest of this feature is unusually well-tested, but nothing here should block merge.

Approving.

---
*Re-review by Claude, verifying the fix commit against the prior consensus round's two MAJOR findings.*

sahrizvi (approved):

PR #1274, opened 2026-09-08
feat: make linked workspace names clickable in the sidebar, CLI, and post-scan dialog

### LINKED ISSUE #1275: workspace: no way to open a linked workspace in the browser from the CLI/TUI
Once a project is linked to a workspace, the workspace name shows up in several places — the TUI sidebar tile, the "already linked" prompt on re-scan, and the \`altimate-code link\` CLI picker — but it's always inert text. The user has to separately go find the workspace on the SaaS index page and match it up by name/id manually.

Workspace names (or an adjacent "open in browser" affordance) should be clickable/actionable wherever they're shown, using the manage URL that's already deterministically derivable from tenant + workspace id.

sahrizvi:
### Issue for this PR

Closes #1275

### Type of change

- [x] New feature

### What does this PR do?

Workspace names shown once a project is linked are now clickable / open the workspace in the browser, in three places:

- **TUI sidebar tile** (`workspace-sidebar.tsx`) — the workspace name and its manage URL render underlined in the accent color and open the browser on click. Mouse events land on the enclosing `<text>` block rather than the inline `<span>`, and the visible affordance is a styled span rather than a real `<a href>` — matching the existing pattern in the sidebar footer's docs/community links, since raw `<a href>` hyperlink nodes crash in the current `@opentui/solid` JSX layer.
- **`AlreadyLinkedDialog`** (shown when re-entering an already-linked project) — added an "Open in browser" menu option. The plugin-facing `DialogSelect` API only takes a plain string `title` (no `titleView`/JSX escape hatch like the native `packages/tui` component has), so the workspace name in the title itself can't be inline-clickable — a selectable option is the equivalent affordance within that API.
- **`altimate-code link` CLI** (`cli/cmd/link.ts`) — the "Currently linked to \"X\"" prompt and the matching picker row now wrap the workspace name in a real **OSC 8 terminal hyperlink** plus underline styling. This is plain stdout text (not the TUI's JSX renderer), so a real hyperlink escape sequence works directly — no click handler needed, the terminal itself makes it clickable.

Also extracted the guarded `open()` + toast-on-failure logic that `WorkspaceLinkedDialog` already had into a shared `openManageUrl()` helper, reused by the sidebar tile and the new `AlreadyLinkedDialog` option.

- `tsc --noEmit` clean on all three changed files
- `bun test test/altimate/plugin/workspace.test.ts` — 41/41 pass, unchanged
- Full monorepo typecheck (pre-push hook) — 13/13 tasks pass
- Built a local binary and manually verified in an isolated sandbox (fake credentials/state, no real network):
  - Sidebar tile renders the workspace name + manage URL underlined in accent color; raw ANSI capture confirms the `\x1b[4m` underline + accent RGB codes wrap only the link text
  - `AlreadyLinkedDialog`'s "Open in browser" option renders and is wired to the same guarded open helper (triggered via the TUI's own `/tui/publish` event API against a headless `serve` + `attach` pair, mirroring the real post-scan trigger)
  - CLI `link` picker: raw OSC 8 escape sequence (`\x1b]8;;<url>\x1b\\<name>\x1b]8;;\x1b\\`) confirmed wrapping the workspace name in both the message and the picker row

### Screenshots / recordings

Verified visually with terminal screenshots (VHS) during development for all three surfaces — sidebar tile (underlined accent link), `AlreadyLinkedDialog`'s new option, and the CLI's OSC-8-linked workspace name. Not embedded here (static images don't convey the click-to-open interaction well); happy to attach on request.

### Checklist

https://claude.ai/code/session_01JsVwcBQP6VUwqEej3vK2zG

## Summary by CodeRabbit

* **New Features**
  * Workspace names and management URLs are now clickable in the workspace sidebar and can open in a browser.
  * Management links correctly preserve URL paths while removing conflicting query and fragment components.
  * The link command now displays plain-text management URLs when clickable links aren’t supported.

* **Bug Fixes**
  * Improved terminal hyperlink compatibility across TTY and non-TTY environments.
  * Enhanced terminal text sanitization to remove control characters and prevent malformed links.
  * Prevented incorrectly placed workspace management URLs.

### DIFF
--- a/packages/opencode/src/altimate/workspace/browser-handoff.ts
+++ b/packages/opencode/src/altimate/workspace/browser-handoff.ts
@@ -184,6 +184,30 @@ export function resolveWorkspaceWebUrl(altimateUrl: string, tenant: string): URL
   }
 }
 
+/** Append ``/w/<id>`` to ``base``'s pathname using real URL semantics, rather
+ * than string-concatenating ``toString()``. The dev-only
+ * ``ALTIMATE_WORKSPACE_WEB_URL`` override (above) can carry its own path/
+ * query/fragment (e.g. a local dev server), and naive concatenation would
+ * land ``/w/<id>`` inside the query string instead of the path — clears
+ * search/hash for the same reason.
+ *
+ * Lives beside ``resolveWorkspaceWebUrl`` (this module is already imported
+ * by both the CLI (``cli/cmd/link.ts``) and the TUI plugin
+ * (``plugin/tui/altimate/workspace.tsx`` / ``workspace-sidebar.tsx``) for
+ * that function, so sharing this one too doesn't cross the deliberate CLI/
+ * TUI self-containment boundary the way importing FROM one side INTO the
+ * other would (see ``isSafeHttpUrl``'s split in ``link.ts`` for that
+ * reasoning). Previously duplicated as three near-identical private copies
+ * — one bug (missing this exact fix) needed three separate edits to close.
+ * (multi-model review, PR #1274 round 7.) */
+export function buildManageUrl(base: URL, workspaceId: number): string {
+  const u = new URL(base)
+  u.pathname = `${u.pathname.replace(/\/+$/, "")}/w/${workspaceId}`
+  u.search = ""
+  u.hash = ""
+  return u.toString()
+}
+
 interface HandoffPending {
   state: string
   expectedTenant: string
--- a/packages/opencode/src/cli/cmd/link.ts
+++ b/packages/opencode/src/cli/cmd/link.ts
@@ -130,8 +130,16 @@ export function buildManageUrl(base: URL, workspaceId: number): string {
  * SGR code than OSC 8, so emitting it unconditionally would make the name
  * look clickable in terminals where it isn't. (cubic, PR #1274, rounds 2 + 3.) */
 export function hyperlink(text: string, url: string | null): string {
-  if (!url || !text) return text
+  if (!text) return text
   const safeText = stripControlChars(text)
+  // Sanitize before checking `url` — the null-URL early return used to skip
+  // stripControlChars entirely, so a caller relying on hyperlink() as its
+  // sanitization boundary got the raw name whenever no manage URL existed
+  // (BYOK/unresolvable deployments). Every caller in this file now also
+  // sanitizes independently before calling this (defense in depth, not the
+  // sole boundary), but this fixes the function's own contract too.
+  // (CodeRabbit, PR #1274 round 4.)
+  if (!url) return safeText
   if (!process.stdout.isTTY) return safeText
   const OSC8 = "\x1b]8;;"
   const ST = "\x1b\\"
@@ -221,7 +229,13 @@ export const LinkCommand = cmd({
       ? projectNameFromRemote(identifier.repoRemote)
       : projectNameFromPath(identifier.projectPath)
     const currentId = existing?.datamate.id
-    const currentName = existing?.datamate.name
+    // Sanitized once here so every downstream display (the picker message,
+    // the "Kept" outro, hyperlink()'s own text) is covered — hyperlink()
+    // only sanitized its own `text` param, not the raw name reaching
+    // `prompts.outro`/`prompts.select`'s message directly. (CodeRabbit,
+    // PR #1274 round 4 — flagged one call site; the underlying gap was
+    // every raw name-interpolation in this file, not just that one.)
+    const currentName = existing ? stripControlChars(existing.datamate.name) : undefined
 
     // Only offer the browser-based handoff when the deployment supports it
     // (freemium only today). Enterprise / localhost / custom-domain callers
@@ -274,12 +288,17 @@ export const LinkCommand = cmd({
           ? "Creates a new workspace and repoints this project to it (no browser step)."
           : "No browser step; configure integrations later in the SaaS.",
       },
-      ...list.map((dm) => ({
-        value: String(dm.id),
-        label:
-          dm.id === currentId ? `● ${hyperlink(dm.name, currentManageUrl)}` : `  ${dm.name}`,
-        hint: dm.id === currentId ? "currently linked here" : undefined,
-      })),
+      ...list.map((dm) => {
+        // Every row's name is server-controlled (any workspace the account
+        // can see, not just ones this user created) — sanitize regardless
+        // of whether this row also goes through hyperlink() below.
+        const safeDmName = stripControlChars(dm.name)
+        return {
+          value: String(dm.id),
+          label: dm.id === currentId ? `● ${hyperlink(safeDmName, currentManageUrl)}` : `  ${safeDmName}`,
+          hint: dm.id === currentId ? "currently linked here" : undefined,
+        }
+      }),
     ]
 
     const pick = await prompts.select<string>({
@@ -376,7 +395,7 @@ async function runBrowserHandoff(
       projectPath: res.binding.project_path,
       linkedAt: Date.now(),
     }, { awaitBackfill: true })
-    bindSpin.stop(`Linked to "${res.binding.datamate_name}".`)
+    bindSpin.stop(`Linked to "${stripControlChars(res.binding.datamate_name)}".`)
     prompts.log.info("Saved memory blocks will sync to this workspace if memory is enabled for it.")
     const manageUrl = await manageUrlFor(res.binding.datamate_id)
     if (manageUrl) prompts.log.info(`Manage it at: ${manageUrl}`)
@@ -384,8 +403,11 @@ async function runBrowserHandoff(
   } catch (err) {
     bindSpin.stop("Link failed.", 1)
     if (err instanceof ConflictError) {
+      const existingName = err.detail.existing_datamate_name
+        ? stripControlChars(err.detail.existing_datamate_name)
+        : "another workspace"
       prompts.log.error(
-        `This project is already linked to "${err.detail.existing_datamate_name ?? "another workspace"}". Workspace "${projectName}" was created but is not linked — re-run \`altimate-code link\` and pick a different action to switch, or delete the new workspace in the SaaS.`,
+        `This project is already linked to "${existingName}". Workspace "${projectName}" was created but is not linked — re-run \`altimate-code link\` and pick a different action to switch, or delete the new workspace in the SaaS.`,
       )
     } else if (err instanceof NotFoundError) {
       prompts.log.error("Workspace not found — the tenant or workspace may have changed.")
@@ -462,8 +484,11 @@ async function createThenBindOrRebind(
     // can pick from the list; if the pre-check missed it, this is the
     // authoritative signal — surface it and hint the picker.
     if (err instanceof ConflictError) {
+      const existingName = err.detail.existing_datamate_name
+        ? stripControlChars(err.detail.existing_datamate_name)
+        : "another workspace"
       prompts.log.error(
-        `This project is already linked to "${err.detail.existing_datamate_name ?? "another workspace"}". Re-run \`altimate-code link\` to switch to a different workspace.`,
+        `This project is already linked to "${existingName}". Re-run \`altimate-code link\` to switch to a different workspace.`,
       )
     } else {
       prompts.log.error(err instanceof Error ? err.message : String(err))
@@ -471,7 +496,11 @@ async function createThenBindOrRebind(
     process.exitCode = 1
     return
   }
-  spin.stop(`Workspace "${created.datamate.name}" created.`)
+  // Sanitized once — echoed back from the create-workspace API response
+  // (not the locally-typed `name` param), so it's technically server data
+  // even though it usually just round-trips the caller's own auto-name.
+  const safeCreatedName = stripControlChars(created.datamate.name)
+  spin.stop(`Workspace "${safeCreatedName}" created.`)
 
   // If the project was already linked, the new workspace exists but the
   // binding still points at the OLD workspace — rebind so the project is
@@ -479,7 +508,7 @@ async function createThenBindOrRebind(
   // wrote the binding as part of the atomic create; we're done.
   if (existing) {
     const rebindSpin = prompts.spinner()
-    rebindSpin.start(`Repointing project at "${created.datamate.name}"...`)
+    rebindSpin.start(`Repointing project at "${safeCreatedName}"...`)
     try {
       await rebindByMatchedIdentifier({
         identifier,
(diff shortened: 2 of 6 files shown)


saravmajestic:
## Multi-model review (Claude + Gemini 3.1 Pro)

Reviewed at **6fb1c43e** — i.e. *after* your bot-findings commit, so the items that commit fixed are excluded below rather than re-reported. Verdict: **one MAJOR that no bot caught**, plus minors.

Credit where it's due first: 6fb1c43e already handles the control-char injection on the workspace name, gates the underline behind a terminal allowlist, and fixes the URL join via real `URL` semantics. That last one — the dev-only `ALTIMATE_WORKSPACE_WEB_URL` override carrying its own path/query, so naive concatenation landed `/w/<id>` in the query string — is a genuine bug **neither** of us found. Fixing all three occurrences rather than only the flagged one was the right call.

Both of us also independently verified 4 of your 5 technical claims as **accurate**, two of them from `@opentui/solid`/`@opentui/core` internals rather than by taking your word:

| Claim | Verdict | Evidence |
|---|---|---|
| `<a href>` crashes the JSX layer | **accurate** | `a` instantiates `LinkRenderable` but `createElement` never passes `options.href`; corroborated by `packages/tui/src/feature-plugins/sidebar/footer.tsx` documenting the same limitation |
| Mouse events land on block-level `<text>`, not inline `<span>` | **accurate** | `isTextNodeRenderable` extracts only `href`/`style`, dropping `onMouseUp`; `TextNodeRenderable` has no layout box or event dispatcher. Also confirmed by `packages/tui/src/ui/link.tsx`'s own docstring: *"Clicking anywhere on the link text opens the URL"* |
| Plugin `DialogSelect` has only `title: string`; native has `titleView` | **accurate, both halves** | `packages/plugin/src/tui.ts:174-186` vs `packages/tui/src/ui/dialog-select.tsx:25,57` (consumed at `:490`, used by `dialog-move-session.tsx:290`) |
| `openManageUrl()` extraction is faithful | **accurate** | Same `isSafeHttpUrl` guard, same two toast variants, same 15s durations, same catch-on-rejection |
| OSC 8 needs no capability check | **was overstated** | Now moot — 6fb1c43e added `terminalSupportsHyperlinks()`. See m1 for the half that remains |

Also worth noting: the `\x1b[24m`-instead-of-`\x1b[0m` choice is **load-bearing, not incidental**. clack wraps a submitted label in `picocolors.dim()`, so a full reset would kill the dim for the rest of the line. Its option window also slices by count rather than display width, so the invisible bytes can't misalign rows. Easy to get wrong; you didn't.

---

### MAJOR — `AlreadyLinkedDialog`'s async option insertion desyncs the keyboard selection

`workspace.tsx` (still current at 6fb1c43e):

```tsx
const [manageUrl, setManageUrl] = createSignal<string | null>(null)
onMount(async () => {
  setManageUrl(await buildManageUrl(props.workspaceId))
})
```

The dialog paints three options, then becomes four when the promise resolves — inserting **"Open in browser" at index 2** and pushing "Skip for now" to 3.

The problem is what the selection does. `packages/tui/src/ui/dialog-select.tsx:212`:

```ts
createEffect(on([() => store.filter, () => props.current], ([filter, current]) => { ... }))
```

`flat()` (derived from `props.options`) is *read* inside but is **not** a tracked dependency, so this never re-runs when the options array changes — and `store.selected` is a **numeric index**. `packages/tui/src/plugin/adapters.tsx:15` confirms the plugin-facing `api.ui.DialogSelect` is this exact component.

So a user who presses Down twice to reach "Skip for now" before the promise resolves keeps a highlight pinned at index 2, which silently becomes "Open in browser". **Pressing Enter to skip launches a browser instead.**

Worth stressing this is a race with a real trigger: this dialog fires on entering an already-linked project, so it competes with a user who already knows which option they want and is reaching for it immediately.

**Suggested fix** — the one your own code already uses elsewhere: `showPostScanGate` is already `async` and already pre-resolves gating state (`const browserAvailable = await isBrowserHandoffAvailable()`), and `WorkspaceLinkedDialog` already receives `manageUrl` as a prop. Add `manageUrl?: string | null` to `AlreadyLinkedProps`, resolve it in the caller before `dialog.replace()`, and keep `AlreadyLinkedDialog` fully synchronous. That removes the moving target rather than trying to resync around it.

---

### Minor

**m1. The CLI never prints the manage URL, so outside the allowlist there's no way to reach it.** `terminalSupportsHyperlinks()` deliberately errs toward false negatives, which is the right instinct — but it means *more* users now get a plain, unstyled workspace name, and `currentManageUrl` is never rendered as text anywhere in `link.ts`. The TUI side degrades properly (`Could not open browser. Copy this URL: ${url}`); the CLI has no equivalent. A dim hint line when `terminalSupportsHyperlinks()` is false would close the gap.

Related: OSC 8 is still emitted unconditionally — the allowlist gates only the underline. That's a defensible trade (the misleading part is the underline, and the bytes are invisible where they're swallowed), but on a genuinely non-ECMA-48 parser they will still print as junk. Your call.

**m2. `hyperlink()` sanitises `text` but interpolates `url` raw.** `stripControlChars` is applied to the name, and `isSafeHttpUrl` exists in this file (`:413`) and is used at `:397` — but not on the value that goes into the escape sequence. Post-6fb1c43e the URL is built through real `URL` manipulation, so control characters are no longer reachable and this is now protocol-validation only: the residual question is whether `resolveWorkspaceWebUrl` can ever return a non-http scheme via the `ALTIMATE_WORKSPACE_WEB_URL` override. If it can't, this is purely defence-in-depth for a generic-looking helper.

**m3. Three new pure functions, no tests.** `hyperlink`, `stripControlChars` and `terminalSupportsHyperlinks` are all deterministic and need no TUI harness — and `workspace.test.ts:67` says the file "focuses on the deterministic layer", so they'd sit squarely in its remit. Worth pinning: `stripControlChars` actually removing `\x1b]8;;`, `hyperlink(text, null)` returning bare text, and the allowlist's TERM/TERM_PROGRAM branches. Your manual ANSI-capture verification is stronger evidence for the *rendering* than a unit test would be, so this is only about the pure layer.

### Nits

- **The sidebar name line's click target includes text that isn't a link.** That first `<text>` block carries the name *plus* the `pinned via --workspace` and drift/unverified hints, so clicking a purely informational badge navigates to the browser. Your comment is admirably honest about the block-level constraint, and claim 2 means it can't be scoped to the span — but the URL line below shows the cleaner shape, because it contains only the link. Splitting the hints into their own `<text>` would make the hit area match the affordance. Also, when `manageUrl()` is null the block still carries `onMouseUp`, leaving a click target that does nothing.
- **`hyperlink("")`** guards a null `url` but not empty `text`, so an empty name would emit the OSC 8 wrapper and underline around zero visible characters — an invisible clickable region. Probably unreachable if names are non-empty by contract.
- **Same action, different wording** across the two dialogs: "Continue editing in browser" / *"Open the workspace in your browser."* vs "Open in browser" / *"View this workspace on the web."*

### Pre-existing, separate ticket

`packages/tui/src/ui/link.tsx` calls `open(props.href)` with **no protocol validation** and swallows every failure (`.catch(() => {})`). Your `openManageUrl` is strictly better on both counts — flagging only because your care about `isSafeHttpUrl` makes the shared component's lack of it conspicuous.

---

*Reviewed by Claude + Gemini 3.1 Pro against 6fb1c43e. One correction to our own output: Gemini quoted the `WorkspaceLinkedDialog` description as "View and configure this workspace in the Altimate web app" — the actual string is "Open the workspace in your browser." The nit stands, the quote was wrong.*

saravmajestic:
## Multi-model re-review (Claude + Gemini 3.1 Pro) — at `e8cc4378dd`

**Verdict: NEEDS REVISION** — one three-line fix. Everything from the previous round is fixed, and fixed at the root.

### Regression check: clean

Both models traced all eight commits independently and agree: **none of the seven fix commits broke or weakened an earlier one.** The sequence is monotonic — round 2 added C1 bytes and dropped the `Apple_Terminal` false positive, round 3 closed the non-TTY leak and a second URL-join duplicate, round 4 moved `stripControlChars` ahead of `hyperlink()`'s early return, rounds 5-6 fixed the test-runner `isTTY` leak. Given seven consecutive fix-on-fix commits, that was the main thing worth checking, and it held.

---

### MAJOR — round 4's sanitisation sweep missed its third sibling site

`link.ts:669`, in `bindOrRebind`'s `ConflictError` handler:

```ts
`Already linked to "${err.detail.existing_datamate_name ?? "another workspace"}". Re-run \`altimate-code link\` to switch.`
```

There are three `existing_datamate_name` interpolations. `:406-407` and `:487-488` both wrap it in `stripControlChars`. This one doesn't — a server-controlled name reaches stdout unescaped, which is precisely the threat `stripControlChars` was added for in the first place.

It also makes a comment false. `link.ts:232-237` claims:

> *"Sanitized once here so every downstream display … is covered … the underlying gap was every raw name-interpolation in this file, not just that one."*

Two of three isn't every, so that comment now overstates the guarantee.

**Fix** — mirror the other two blocks:

```ts
if (err instanceof ConflictError) {
  const existingName = err.detail.existing_datamate_name
    ? stripControlChars(err.detail.existing_datamate_name)
    : "another workspace"
  prompts.log.error(
    `Already linked to "${existingName}". Re-run \`altimate-code link\` to switch.`,
  )
```

A test alongside the existing adversarial-name case would stop it recurring. This is the one item worth doing before merge.

---

### Confirmed fixed from the last round — all of it, at the root

- **The dialog selection race.** `manageUrl` is now a required synchronous prop; `AlreadyLinkedDialog` has no `onMount`, no signal for options, no async effect, so the options list is stable from the first frame. Fixed at **both** call sites (`workspace.tsx:1159` and `:1199`), not just one. Both models verified this independently and agree it's a root-cause fix rather than a timing mitigation — the async-insertion shape is no longer expressible.
- **The CLI URL fallback.** `link.ts:258-260` prints `Manage it at: <url>` when `terminalSupportsHyperlinks()` is false, so the URL is reachable outside the allowlist.
- **Sanitisation ahead of the early return.** `hyperlink()` now strips before `if (!url)`, closing a hole where the null-URL path (BYOK/unresolvable) returned the raw name — a gap I did not spot last round and CodeRabbit did.
- **The non-TTY leak.** `if (!process.stdout.isTTY) return safeText` means piped/redirected output never carries OSC 8 bytes.
- **Both nits.** `if (!text) return text` closes the invisible-clickable-region case, the `pinned via --workspace` hint moved to its own line so the click region no longer spans non-link text, and `onMouseUp` is now omitted entirely rather than a no-op when there's no URL.

### The declined Kilo finding was correctly declined

Verifying in a real pty (via tmux, because the sandboxed shell always pipes stdout) rather than trusting the bot *or* your own earlier comment was the right call. `isTTY` genuinely is an own property when stdout is a TTY, so the suggested unconditional `delete` would have destroyed it whenever those tests run interactively. Round 5's finding was also real and neatly self-referential — the leak was in the helper written to fix round 4's finding.

### The test file is now the strongest part of this PR

`test/cli/cmd/link.test.ts` carries a regression test for **each** of the six rounds, not just the last: the embedded-OSC-8 and adversarial-name cases, C1/DEL bytes, the `Apple_Terminal` false positive, the non-TTY leak, the URL join with a query/fragment, sanitisation-when-url-is-null, the `restoreTTY` helper itself, and the `VTE_VERSION >= 5000` boundary in both directions. These assert behaviour on pure functions rather than echoing the implementation back.

Reasonably left untested: `openManageUrl` (needs a `TuiPluginApi` mock for its two toast branches) and the sidebar rendering.

---

### Minor

**m1. `hyperlink()` sanitises `text` but interpolates `url` raw.** Sharper than last round, because `hyperlink` is now **exported** for tests: its contract is wider than its two in-file callers, so an external caller passing something like `created.manage_url` straight through would defeat the escaping that the `text` path is careful about. `isSafeHttpUrl` already exists in this file — `if (!url || !isSafeHttpUrl(url)) return safeText` closes it.

**m2. The manage-URL join now lives in three places** — `link.ts`'s `buildManageUrl`, `workspace.tsx`'s own copy, and the sidebar's `resolveManageBase` plus inline join — and `isSafeHttpUrl` is duplicated across two. The CLI/TUI split is deliberate and we agree with it; the CLI shouldn't import from the TUI plugin path. But the cost is now demonstrated rather than theoretical: round 3's single URL-join bug needed three separate edits, and landed in all three only because you went looking past the one site that was reported. `resolveWorkspaceWebUrl` in `altimate/workspace/browser-handoff.ts` is already the shared origin of all three, so a `buildManageUrl(base, id)` beside it would be reachable from both sides without crossing the boundary that matters — and would delete the class instead of the instance.

### Nits

- The same action is worded differently across the two dialogs: "Continue editing in browser" / *"Open the workspace in your browser."* vs "Open in browser" / *"View this workspace on the web."* Possibly deliberate given one is a post-creation success card and the other a you're-already-linked prompt — ignore if so.
- `workspace.tsx:595-597` says `openManageUrl` has "the same … behavior as `WorkspaceLinkedDialog`'s 'open' action below" — that action no longer implements the behaviour, it calls `openManageUrl`. Phrasing only.

### On the accumulated complexity

`link.ts` gained four helpers for what began as "underline a name", and we looked at whether that's over-built. It isn't: each exists because a specific real failure was found, all four are pure and now tested, and the alternative for `terminalSupportsHyperlinks` was a new dependency for one affordance. The comments citing reviewer and round are unusual, but they earn their keep — they're what stopped round 6 from re-litigating round 5 incorrectly.

One asymmetry that is *correct* despite looking inconsistent: the sidebar still resolves its URL asynchronously, unlike the dialog. That's fine — a sidebar tile has no selection indices to shift, so the failure mode that made the dialog's async resolution unsafe doesn't exist there.

---

*Reviewed by Claude + Gemini 3.1 Pro against `e8cc4378dd`. Correction to our own prior output: in this round Claude audited the name interpolations and reported round 4's "every raw name-interpolation" claim as verified-accurate. That was wrong — the grep matched variable-style interpolations and missed the inline `${err.detail.… ?? …}` expression at `:669`. Gemini caught it; the MAJOR above is its find, confirmed against the code.*

sahrizvi:
Replying to each point from the round-7 re-review, all addressed in `e016cc0258`:

**MAJOR — the third `existing_datamate_name` site (`link.ts:669`)**: Confirmed and fixed. Round 4's sweep fixed the same pattern in `runBrowserHandoff` and `createThenBindOrRebind` but missed this one in `bindOrRebind`. Re-swept the whole file after fixing it — every display site now goes through `stripControlChars`; the three `recordApprovedBinding` cache-storage calls are deliberately left raw (sanitization belongs at display time, not storage time). The "every raw name-interpolation" claim in the round-4 comment is now actually true.

**m1 — `hyperlink()` validates `text` but not `url`**: Fixed. Added the existing `isSafeHttpUrl` guard. Agreed this became a real gap rather than a theoretical one once the function was exported for tests — its contract is wider than its two trusted in-file callers now.

**m2 — three-way URL-join duplication**: Fixed by moving `buildManageUrl` to `browser-handoff.ts`, beside `resolveWorkspaceWebUrl` — a module all three call sites already import for that function, so it doesn't cross the CLI/TUI boundary `isSafeHttpUrl`'s split exists to respect. Good point that the cost was demonstrated rather than theoretical (round 3's bug needed three edits); this deletes the class. Tests moved alongside it in `browser-handoff.test.ts`.

**Nit — stale `openManageUrl` comment**: Fixed, reworded to describe what the function does instead of a now-circular cross-reference.

**Nit — differing "open in browser" wording**: Left as-is, per your own "possibly deliberate — ignore if so." It is: one's a post-creation success card, the other a you're-already-linked prompt.

Thanks for the thorough pass across all seven rounds, and for the correction on your own prior output about `:669` — appreciated.

saravmajestic (dismissed):

saravmajestic (approved):

PR #1294, opened 2026-09-09
feat(telemetry): first-run health — startup_ready, event_loop_stall, registration timing

anandgupta42:
## Why

The 0.11.0 first-run freeze (#1292: an in-process `@npmcli/arborist` install blocking Bun's event loop for 2.5 to 5 minutes) shipped in six releases without a trace in telemetry. Three reasons, all structural:

1. Nothing timed startup or registration. The first event a session reports is `session_start`; the next is a `generation` after the model answers. A freeze between them produced no number anywhere.
2. Events flush on a 5 s `setInterval` on the same loop that was blocked. A frozen-then-killed process died with its buffer and left only `session_start` + `task_classified` (a "dead session": 0 to 2 s span, no generation, no error, no end).
3. The only place it was visible was first-generation latency split by brand-new versus returning machines, which no default view does. Fresh machines on 0.9.5 and 0.9.7 had a p90 of 195 to 244 s versus 17 to 21 s for returning ones.

## What

| Event | Emitted | Fields |
|---|---|---|
| `startup_ready` | once per process: `serve` listening, TUI transport resolved, `run` about to send its first prompt | `command`, `duration_ms` (process uptime), `fresh_install` |
| `event_loop_stall` | a 250 ms monitor tick that fires more than 1 s late; capped at 20 per process; main thread and TUI server worker | `blocked_ms`, `since_start_ms`, `thread`, `command` |
| `altimate_base_registration` | every `registerAfterConsent` outcome, so the TUI dialog and the HTTP consent route both count | `result`, `duration_ms`, `status` |

Plus: anchor events (`first_launch`, `startup_ready`, `event_loop_stall`, `altimate_base_registration`, `session_start`) flush immediately instead of waiting for the interval, and the CLI middleware records the top-level command via `Telemetry.setCommand`.

The stall monitor reports when the loop resumes, so a stall that ends in a killed process is still lost, but every stall a user waited through is now a number, and it is flushed before anything else can go wrong.

`docs/internal/first-run-telemetry.md` carries the rationale, the event table, and KQL for startup time, stalls, registration, and the fresh-versus-returning dead-session rate (the last one works on historical data).

## Verification

- `bun test`: 8 new tests (startup_ready once per process; pure lag check; monitor detects a synchronous 250 ms block and stays quiet otherwise; idempotent start/stop; registration success, HTTP 429 with status, misconfigured gateway URL, network failure) plus the existing telemetry, Altimate Base, TUI and provider-API suites, all green. `tsgo --noEmit` clean. Upstream marker guard clean.
- Live: `serve` in a fresh isolated HOME with `APPLICATIONINSIGHTS_CONNECTION_STRING` pointed at a local sink received `startup_ready command=serve` within the anchor flush, from both the dev build (2048 ms) and a `bun build --compile` single-target binary (1561 ms), so the monitor and `isMainThread` behave in the compiled form.
- Second-model review folded in: `startup_ready` now fires for `run --command` as well as plain prompts, and a `ConfigurationError` from the gateway URL is reported as `result: configuration` instead of escaping before the timer starts.

## Known limit

Telemetry is initialised and shut down per prompt by `session/prompt.ts`, so in a long-lived `serve` process the stall monitor, like every other event, is dark between prompts. That is pre-existing lifecycle behaviour, noted here so nobody reads an absence of stalls outside a prompt window as proof of health.

https://claude.ai/code/session_01WrT7MEUL5CYvpjf9cJbeQR

---

> [!NOTE]
> **Medium Risk**
> Touches telemetry flush timing and runs instrumentation on CLI main thread and TUI worker at startup; low user-facing risk but changes when events ship and adds process-lifetime monitors tied to telemetry init/shutdown.
> 
> **Overview**
> Adds **first-run health telemetry** so long startup freezes and silent “dead sessions” show up in App Insights instead of only `session_start` with no follow-up.
> 
> **New events:** `startup_ready` (once per process when `tui` / `serve` / `run` can work, with uptime and `fresh_install`), `event_loop_stall` (250 ms monitor, >1 s late tick, capped at 20, main vs worker), and `altimate_base_registration` (every `registerAfterConsent` outcome with duration and HTTP status when relevant).
> 
> **Behavior:** Selected **anchor** events (`first_launch`, `startup_ready`, `event_loop_stall`, `altimate_base_registration`, `session_start`) **flush immediately** instead of waiting on the 5 s interval. CLI middleware records the top-level command via `Telemetry.setCommand`; the TUI worker starts the stall monitor and calls `Telemetry.init()` at boot so pre-prompt blocking is observable.
> 
> **Altimate Base:** Registration paths report telemetry for expired consent, bad gateway URL (`configuration`), success/failure; caller aborts are classified as `cancelled` rather than `network`.
> 
> **Docs/tests:** `docs/internal/first-run-telemetry.md` with KQL for startup, stalls, registration, and dead-session rate; new unit tests plus an existing flush-retry test adjusted to avoid anchor auto-flush.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f6343e2085e259f27014da0990a93f74df51cdbe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Adds first-run health telemetry so the 2.5–5 minute freeze from the 0.11.0 first-run `@npmcli/arborist` install shows up in telemetry instead of a "dead session" that reports only `session_start`.

- Emits `startup_ready` once per process when `serve`, `tui`, or `run` can start work, with command, duration, and fresh-install flag.
- Emits `event_loop_stall` when a 250 ms monitor tick is over 1 s late, capped at 20 per process, on main and worker threads; the TUI worker starts the monitor at boot.
- Emits `altimate_base_registration` for every registration outcome on both TUI and HTTP consent paths; a caller abort reports `cancelled` and a bad gateway URL reports `configuration`.
- Anchor events (`first_launch`, `startup_ready`, `event_loop_stall`, `altimate_base_registration`, `session_start`) flush immediately instead of waiting for the 5 s interval.
- CLI middleware records the top-level command via `Telemetry.setCommand`, and documentation with KQL queries lives in `docs/internal/first-run-telemetry.md`.

<sup>Written for commit f6343e2085e259f27014da0990a93f74df51cdbe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1294?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added first-run health telemetry for startup readiness and event-loop stalls across CLI, server, and TUI workflows.
  * Added telemetry for Altimate Base registration outcomes, including duration, result type, cancellation status, and relevant HTTP status details.
  * Startup health events now flush immediately for more timely reporting.

* **Documentation**
  * Added guidance for interpreting first-run health telemetry and querying startup, event-loop, registration, and session metrics.

### DIFF
--- a/docs/internal/first-run-telemetry.md
+++ b/docs/internal/first-run-telemetry.md
@@ -1,0 +1,79 @@
+# First-run health telemetry
+
+CLI telemetry lands in Azure Application Insights (`altimate-code-os`). Event names are the
+`Telemetry.Event` `type`; string properties are in `customDimensions`, numbers in
+`customMeasurements`, and `session_id` is lifted into the `session_Id` column.
+
+## Why these events exist
+
+On 2026-09-09 a fresh 0.11.0 install froze for 2.5 to 5 minutes on first use (an in-process
+`@npmcli/arborist` install blocked Bun's event loop). Nothing in telemetry showed it:
+
+- no event timed startup or registration, so a silent freeze produced no number anywhere;
+- events flush on a 5 s interval on the same loop, so a frozen-then-killed process died with
+  its buffer and left only `session_start` + `task_classified` (a "dead session");
+- the only place it was visible was first-generation latency split by brand-new versus
+  returning machines, which no default view does.
+
+Three events and one behaviour change close the gap:
+
+| Event | Emitted | Key fields |
+|---|---|---|
+| `startup_ready` | once per process when `tui`, `serve` or `run` can do work | `command`, `duration_ms` (process uptime), `fresh_install` |
+| `event_loop_stall` | when a 250 ms monitor tick fires more than 1 s late, capped at 20 per process | `blocked_ms`, `since_start_ms`, `thread`, `command` |
+| `altimate_base_registration` | on every `registerAfterConsent` outcome (TUI and HTTP consent paths) | `result`, `duration_ms`, `status` |
+
+Anchor events (`first_launch`, `startup_ready`, `event_loop_stall`, `altimate_base_registration`,
+`session_start`) flush immediately instead of waiting for the interval.
+
+## Queries (KQL)
+
+Startup time by command, fresh versus returning machines:
+
+```kusto
+customEvents
+| where timestamp > ago(7d) and name == "startup_ready"
+| extend v=tostring(customDimensions.cli_version), cmd=tostring(customDimensions.command), fresh=tostring(customDimensions.fresh_install)
+| summarize n=count(), p50_s=percentile(todouble(customMeasurements.duration_ms)/1000,50), p90_s=percentile(todouble(customMeasurements.duration_ms)/1000,90) by v, cmd, fresh
+| order by v desc
+```
+
+Event-loop stalls (the freeze, measured directly):
+
+```kusto
+customEvents
+| where timestamp > ago(7d) and name == "event_loop_stall"
+| extend v=tostring(customDimensions.cli_version), cmd=tostring(customDimensions.command), thread=tostring(customDimensions.thread)
+| summarize stalls=count(), machines=dcount(user_Id), p50_blocked_s=percentile(todouble(customMeasurements.blocked_ms)/1000,50), max_blocked_s=max(todouble(customMeasurements.blocked_ms))/1000 by v, cmd, thread
+| order by machines desc
+```
+
+Registration outcome and latency:
+
+```kusto
+customEvents
+| where timestamp > ago(7d) and name == "altimate_base_registration"
+| extend v=tostring(customDimensions.cli_version), result=tostring(customDimensions.result)
+| summarize n=count(), p50_s=percentile(todouble(customMeasurements.duration_ms)/1000,50), p90_s=percentile(todouble(customMeasurements.duration_ms)/1000,90) by v, result
+```
+
+Dead-session rate, fresh versus returning (works on historical data too):
+
+```kusto
+let fresh = customEvents
+| where timestamp > ago(30d) and name == "first_launch" and tostring(customDimensions.is_upgrade) == "false"
+| summarize launch=min(timestamp) by user_Id;
+customEvents
+| where timestamp > ago(30d) and isnotempty(session_Id)
+| extend v=tostring(customDimensions.cli_version)
+| summarize started=countif(name=="session_start"), gens=countif(name=="generation"), errors=countif(name in ("error","provider_error")), start=minif(timestamp, name=="session_start"), last_ts=max(timestamp), ver=any(v) by session_Id, user_Id
+| where started > 0
+| join kind=leftouter fresh on user_Id
+| extend fresh_machine = isnotnull(launch) and start between (launch .. (launch + 24h))
+| extend dead = gens == 0 and errors == 0 and datetime_diff("second", last_ts, start) <= 2
+| summarize sessions=count(), dead_pct=round(100.0*countif(dead)/count(),1), generated_pct=round(100.0*countif(gens > 0)/count(),1) by ver, fresh_machine
+| order by ver desc, fresh_machine desc
+```
+
+Gotchas: the `az monitor app-insights query --analytics-query` argument must be a single line;
+`last` is a reserved word; `percentileif` does not exist (filter first, then `percentile`).
--- a/packages/opencode/src/altimate/free/client.ts
+++ b/packages/opencode/src/altimate/free/client.ts
@@ -5,6 +5,8 @@ import { Installation } from "../../installation"
 import { Log } from "../util/log"
 import { FreeTierStore } from "./store"
 import { FreeTierUrl } from "./url"
+// altimate_change — first-run health: time every Altimate Base registration
+import { Telemetry } from "../telemetry"
 
 const log = Log.create({ service: "altimate-base" })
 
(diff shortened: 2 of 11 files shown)


anandgupta42:
Review round addressed in the amended commit (CI was red on one test; now green locally: 574 tests across telemetry, Altimate Base, TUI and provider suites, typecheck and marker guard clean).

**CI**: `telemetry.flush > flush re-adds events to buffer on network error` tracked a `session_start`, which is now an anchor event and flushes the moment it is tracked, consuming the test's first fetch. The test now uses a non-anchor `error` event so it keeps exercising the retry path alone.

**Cursor / Kilo / cubic**
- Registration events dropped without init (high): `reportRegistration` now awaits `Telemetry.init()` (idempotent) before tracking, same pattern as `OnboardingTelemetry.emit`, so the anchor flush fires instead of the event sitting in a pre-init buffer.
- Pre-init anchors never flush (medium): `doInit` now flushes immediately after enabling when the buffer already holds an anchor event (`first_launch` always does).
- Worker stalls report `command: "unknown"` (medium): `setCommand` publishes the command via `ALTIMATE_CLI_COMMAND`, which the TUI server worker inherits and reads on module load.
- Default `$0 [project]` stores the project path as the command: the middleware now only accepts a registered command name from `_[0]` and otherwise records `tui`.
- `startup_ready("tui")` before `validateSession`: moved after successful validation; the comment notes that on the internal transport this measures time-to-interactive, not worker boot.
- `stopLoopMonitor()` before init settled: moved into `doShutdown` after `await initPromise`.
- Expired consent never reported: timing starts before `redeemConsent`, and the `cancelled` outcome is reported. The pending-dedupe path deliberately stays one event per real attempt.
- Tests: `isolateAltimateBaseHome` at module scope, gateway env restored in `afterEach`, `Telemetry.resetFirstRunStateForTest()` seam used in `afterEach`, the "stays quiet" assertion removed (a scheduler pause on a loaded runner is a real stall), new `cancelled` case.
- Docs: the fresh-cohort subquery now filters `is_upgrade == "false"`.

Not changed: the per-prompt telemetry lifecycle (dark between prompts in `serve`), which predates this PR and is called out under "Known limit".

anandgupta42:
Round 3 addressed in the amended commit:

- **TUI worker monitor starts too late** (Cursor, medium): the worker now calls `Telemetry.startLoopMonitor()` at boot, before config and plugin loading, instead of waiting for the first prompt's `init()`. Its stall events buffer until init enables telemetry, which now drains anchor events immediately.
- **Command set omits `link` / `workspace-serve`** (Cursor): added.
- **Caller aborts reported as `error`/`network`** (CodeRabbit): `registerOnce` now throws `RegistrationError(..., "cancelled")` when the caller's signal is aborted during the request, and `registrationResult` maps the pre-request `AbortError` to `cancelled`; the registration timeout stays `network`. New test with a pre-aborted signal.
- **`ALTIMATE_CLI_COMMAND` leaks from the first-run test** (CodeRabbit major, cubic): saved in `beforeEach` and restored or deleted in `afterEach` before `resetFirstRunStateForTest()`.
- **Sleep-based `settled()` poll** (cubic, `test/AGENTS.md`): replaced by `await Telemetry.init()`. `reportRegistration` attaches its track callback to that same promise before the caller's promise settles, so awaiting it queues the assertion behind the track deterministically. No sleep, no poll.

anandgupta42:
Cursor `worker.ts:35` (pre-init worker stalls never flush): the TUI server worker now calls `Telemetry.init()` at boot right after starting the stall monitor, the same early, idempotent init the CLI middleware does on the main thread. init drains buffered anchor events as soon as it enables, so a stall the user waits through before the first prompt is reported even if they then quit. The Kilo "failure" status on this PR is Kilo's own provider rate limit (its summary says the review did not run); its previous run at 4dc0c10 reported no issues.

anandgupta42:
Kilo `worker.ts:38` (worker boot init cannot see a config-file opt-out): fixed in the latest commit. `doInit()` now records when `Config.get()` could not be read (no Instance context yet) instead of treating that as the final answer. The next `Telemetry.init()` call, which the prompt loop makes inside Instance context, re-reads the config once; if `telemetry.disabled` is set it stops the loop monitor and flush timer and disables telemetry for the rest of that generation, and `track()` drops events from then on. Once a re-check has read the config, `init()` falls back to the plain idempotent path, so there is no per-call cost.

What this does not change: the env var is still the only opt-out honoured before config becomes readable, so anchor events tracked on the worker between boot and the first prompt can still ship for a config-file-only opt-out. That matches the existing main-thread behaviour, where the CLI middleware initialises telemetry before `Instance.provide()` for every command. Holding flushes until the config is verified would re-open the frozen-then-killed blind spot this PR exists to close, so I kept the immediate anchor flush.

Tests: unreadable-then-disabled (no further fetch, loop monitor stopped), unreadable-then-enabled (stays enabled, config not re-read a third time), readable-and-disabled (existing behaviour).

anandgupta42:
This PR merged before the config opt-out fix above landed on its branch, so commit 612bf5bb38 here is not in main. The fix is re-opened as #1295 off main.

PR #1295, opened 2026-09-10
fix(telemetry): re-check the config-file opt-out once config is readable

anandgupta42:
Follow-up to #1294, which merged before this fix landed on its branch. Addresses Kilo's `worker.ts:38` finding there: "Worker boot init cannot see a config-file telemetry opt-out".

## Problem

#1294 made the TUI server worker call `Telemetry.init()` at module load, before Instance context exists on that thread. Inside `doInit()`, `Config.get()` throws there and was treated as "not disabled", so telemetry came up enabled. Because `init()` is idempotent, the prompt loop's later call (inside Instance context, where the config is readable) joined the settled promise and never re-read it. A user with `telemetry.disabled: true` in `opencode.json` and no env var shipped telemetry for the whole first session.

## Fix

- `doInit()` records when the config check could not run (`configOptOutUnverified`).
- The next `Telemetry.init()` call re-reads the config once it is readable. If the file opts out, it stops the loop monitor and flush timer and disables telemetry for the rest of that generation; `track()` then drops events. Once a re-check has read the config, `init()` is back on the plain idempotent path.
- The env var remains the only opt-out honoured before the config is readable, which matches the main thread, where the CLI middleware initialises telemetry before `Instance.provide()` for every command. Holding flushes until the config is verified would re-open the frozen-then-killed blind spot #1294 exists to close, so the immediate anchor flush is kept.

## Tests

`test/altimate/telemetry/first-run-health.test.ts`: unreadable-then-disabled (no further fetch, loop monitor stopped), unreadable-then-enabled (stays enabled, config not re-read a third time), readable-and-disabled (existing behaviour).

https://claude.ai/code/session_01WrT7MEUL5CYvpjf9cJbeQR

---

> [!NOTE]
> **Medium Risk**
> Changes telemetry enable/disable timing and init concurrency guards; incorrect behavior would either leak events or drop legitimate telemetry, but scope is isolated to the altimate telemetry module with strong test coverage.
> 
> **Overview**
> Fixes a privacy bug where **early `Telemetry.init()`** (TUI worker at module load, before Instance context) could not read `opencode.json`, started telemetry anyway, and **never re-evaluated** `telemetry.disabled` because later `init()` calls only joined the settled promise.
> 
> **`configOptOutUnverified`** is set when `Config.get()` fails during `doInit()`; the next **`init()`** runs **`recheckConfigOptOut()`** (memoized via **`recheckPromise`**) and, if the file opts out, tears down the flush timer, loop monitor, and App Insights state so **`track()`** drops events. A **generation token** prevents a stale recheck from disabling a new lifecycle after **shutdown + re-init**; shutdown also clears recheck flags. Failed flushes no longer **retry into the buffer** once telemetry has been disabled mid-flight.
> 
> Adds **`first-run-health.test.ts`** coverage for late disable, stay-enabled after verified config, unchanged first-read disable, repeated recheck until readable, and the shutdown race.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7c7ca1e765f4493551e0069784fc7092a7e7cfa1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---
## Summary by cubic
Fixes the telemetry config-file opt-out being ignored when the first `Telemetry.init()` runs before the config is readable (the TUI server worker). Users with `telemetry.disabled: true` in `opencode.json` no longer ship telemetry for the whole first session.

**Bug Fixes**
- `doInit()` now flags when the config check could not run; the next `init()` call re-reads the config once readable and, if it opts out, disables telemetry for the rest of the session.
- A pending recheck that finishes after shutdown and re-init is ignored, so it can't clear the new session's telemetry state.
- Failed flushes don't re-buffer events once the config recheck has disabled telemetry.
- The env var remains the only opt-out honored before the config is readable.
- The anchor flush is kept so the frozen-then-killed blind spot from #1294 stays closed.

<sup>Written for commit 7c7ca1e765f4493551e0069784fc7092a7e7cfa1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1295?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Telemetry now correctly honors configuration-based opt-outs when configuration is initially unavailable during startup.
  * Telemetry state is reset cleanly across initialization and shutdown cycles.
  * Prevented stale configuration checks from affecting later telemetry sessions.
  * Events are no longer restored after telemetry is disabled during a failed flush.
* **Tests**
  * Added coverage for delayed configuration checks, retry behavior, session restarts, and disabling telemetry while monitoring is active.

### DIFF
--- a/packages/opencode/src/altimate/telemetry/index.ts
+++ b/packages/opencode/src/altimate/telemetry/index.ts
@@ -1844,6 +1844,24 @@ export namespace Telemetry {
   let reinitPromise: Promise<void> | undefined
   // altimate_change — the currently running flush, so shutdown waits rather than racing it.
   let inFlightFlush: Promise<void> | undefined
+  // altimate_change start — first-run privacy: doInit() can run outside Instance context (the TUI
+  // server worker calls Telemetry.init() at module load, before Instance.provide() has run on that
+  // thread), where Config.get() throws and is treated as "not disabled" so telemetry doesn't hang
+  // the worker forever waiting on config it cannot read yet. That is correct for the FIRST init —
+  // the env var checked above is the only opt-out this thread can honor before config is readable —
+  // but it must not be the LAST word: a user with `telemetry.disabled: true` in opencode.json and no
+  // env var would otherwise ship telemetry for the rest of the process, because init() is idempotent
+  // and every later call (notably session/prompt.ts's init() inside Instance context, where config IS
+  // readable) just joins the already-settled promise instead of re-evaluating config.
+  //
+  // `configOptOutUnverified` tracks whether THIS generation's config check actually ran; when it
+  // didn't, the next init() call (regardless of caller) re-checks config once config becomes
+  // readable — see recheckConfigOptOut() and init() below.
+  let configOptOutUnverified = false
+  // altimate_change — memoises the in-flight recheck so concurrent init() callers racing a pending
+  // recheckConfigOptOut() share one Config.get() instead of each starting their own.
+  let recheckPromise: Promise<void> | undefined
+  // altimate_change end
 
   // altimate_change start — per-launch correlation id, shared across threads via the environment.
   // The TUI worker is spawned after the CLI middleware has already initialised telemetry on the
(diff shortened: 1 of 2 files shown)


anandgupta42:
All findings on cdd0c7ca20 are addressed in the latest commit:

- **Generation race** (CodeRabbit `index.ts:2029`, cubic `index.ts:2040`): the re-check captures its generation (`initPromise`) before awaiting `Config.get()` and ignores a completion whose generation has since been shut down or re-initialised, so it can no longer clear the next generation's buffer, timer, `appInsights` or flag. `doShutdown()` also resets the re-check memo, and the memo's `finally` only clears itself. Regression test: a re-check left pending across `shutdown()` + `init()` resolves with `disabled: true` and the new generation stays enabled.
- **Flush write-back** (Kilo `index.ts:2062`): `doFlush()`'s retry path now returns on `shuttingDown || !enabled`, so a fetch failing after the re-check disabled telemetry cannot refill the cleared buffer.
- **Vacuous loop-monitor assertion** (Kilo `first-run-health.test.ts:179`, CodeRabbit `:185`, cubic): the test restarts the monitor with a 10 ms interval and 100 ms threshold, proves a stall is detected while enabled, then asserts none after the disabling re-check. Removing `stopLoopMonitor()` from the re-check now fails it.
- **Unreadable re-check path** (Kilo `first-run-health.test.ts:193`): new test where `Config.get()` rejects twice; the third `init()` still re-reads and then disables.
- **TypeScript job**: its only failure was the repo's marker-integrity test. The new `altimate_change start` opened inside an already-marked block in `init()` is now a single-line marker; start/end counts balance again.

PR #1314, opened 2026-09-16
fix: [AI-9171] create a quick workspace from an already-linked project

saravmajestic:
## Problem

In a project that is already linked, the `altimate link` picker's **"＋ Create a quick workspace … here"** row always failed — and the error told the user to re-run the command they were already inside.

```
◇  Currently linked to "jaffle_shop-altimate". Pick a workspace (or create a new one):
│  ＋ Create a quick workspace "jaffle_shop-altimate" here
│
■  Failed to create workspace.
■  This project is already linked to "jaffle_shop-altimate". Re-run `altimate-code link`
   to switch to a different workspace.
```

The row's own hint promises the opposite: *"Creates a new workspace and repoints this project to it (no browser step)."*

## Cause

`createThenBindOrRebind` already had a correct rebind branch. It was simply unreachable.

Step one called `createAndBind` → `POST /datamate-project-bindings/`, whose server handler pre-checks both identifiers and returns 409 **before creating anything** — deliberately, so a binding conflict cannot strand a half-created workspace. On an already-linked project that refuses the entire call, so nothing was created and the rebind below it was dead code.

The handler's own comment shows the assumption that broke — *"the new workspace exists but the binding still points at the OLD workspace"*. It does not exist: create and bind are atomic server-side, so the binding conflict takes the creation down with it.

Worth noting the neighbouring **"＋ Set up in browser"** row is correctly hidden when already linked, with a comment explaining this exact 409. One of the two guards was applied; the other was missed.

## Fix

Split the two cases, because the server offers two different things:

- **Unlinked** — unchanged. `createAndBind` still creates and binds in one transaction, which is what makes a stranded workspace impossible there.
- **Already linked** — create the workspace unbound via `POST /datamates/`, then repoint through the rebind path that already existed.

### The trap this had to avoid

`createWorkspaceUnbound` sends `memory_enabled` and `knowledge_engine_enabled` **explicitly**. `POST /datamates/` is the SaaS/extension creation path and defaults both to `false`, while the create-and-bind path (`_create_datamate_flush_only`) sets both `true`.

Without those two lines the same menu row would hand back a differently-configured workspace depending only on whether the project happened to be linked — memory and the knowledge engine silently off, with nothing surfacing the difference. The comment names the coupling so the two move together if backend defaults change.

### Error message

After this change a 409 can only mean another workspace claimed the project mid-selection, so it says that, instead of directing the user back into the command they are already running.

### Against a real backend

Local backend on `:5001`, tenant `hackdev`, project bound to workspace 63.

**Before** — what the old code did:

```
POST /datamate-project-bindings/  {"name":"jaffle_shop-altimate", <same identifiers>}
  HTTP 409
  detail: This project is already linked to a different workspace
```

Nothing created, rebind unreachable.

**After** — what the new code does:

```
POST /datamates/  {..., "memory_enabled":true, "knowledge_engine_enabled":true}
  -> {"id": 66}

PUT /datamate-project-bindings/by-remote
    {"target_datamate_id":66, "expected_current_datamate_id":63}
  -> {"binding":{"id":44,"datamate_id":66, ...}}

GET /datamate-project-bindings/by-remote
  -> bound to workspace 66 - 'jaffle_shop-altimate'
```

And the new workspace carries the workspace defaults, not the SaaS ones:

```
memory_enabled           = True
knowledge_engine_enabled = True
```

All seeded rows removed afterwards; `hackdev` restored to its prior counts.

### Tests

New `test/altimate/workspace/create-workspace-unbound.test.ts` — 7 tests stubbing `globalThis.fetch`, following the `skill-sync.test.ts` harness (real credentials file, assertions on the request that actually goes out). They pin the two things that would regress silently: that this does **not** reach the binding router, and that both feature flags are sent.

Mutation-tested rather than assumed — each of these fails the suite:

| Mutation | Result |
|---|---|
| drop `memory_enabled` | 1 fail |
| drop `knowledge_engine_enabled` | 1 fail |
| post to the bindings router instead | 1 fail |
| remove the id guard | 2 fail |

Existing suites: `test/cli/cmd/link.test.ts` + all of `test/altimate/workspace/` → **503 pass, 0 fail**. Typecheck clean on both changed files. `oxlint` contributes no new errors (the 4 reported are pre-existing, in unrelated `tsconfig.json` files).

## Note for review

The alternative fix is backend-side: a `replace_existing` flag on `POST /datamate-project-bindings/` would make this atomic and remove the duplicated defaults entirely. That is the cleaner long-term shape, but it is a two-repo change; this keeps the fix in the CLI and reuses the rebind path that already exists.

Ticket: [AI-9171](https://altimateai.atlassian.net/browse/AI-9171)

---
## Summary by cubic
Fixes the “＋ Create a quick workspace … here” row in both the CLI and TUI on already-linked projects, where it always failed and pointed the user back into the command they were already running. The row now creates the workspace unbound, then repoints the existing binding to it.

- Unlinked projects keep the atomic create-and-bind behavior; only the already-linked path takes the new two-step route.
- The unbound create sends `memory_enabled` and `knowledge_engine_enabled` explicitly, matching the configureation the create-and-bind path produces.
- The account is pinned across the create-and-rebind pair so an account switch mid-flow cannot rebind on the wrong tenant.
- A strict id guard rejects malformed responses instead of coercing values (a boolean would have rebound the project to workspace 1).
- The rebind's server binding row is what gets cached, replacing locally-derived identifiers the server never stored.
- 409 messages are split by which create ran: a conflict on the unbound path is reported as another workspace claiming the project.

<sup>Written for commit 7b21606fcc22f7ee8381670460c2feb7765a2aba. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1314?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Added support for creating a standalone private workspace during `altimate link`.
  - Existing workspace links can now be safely repointed without creating an unintended project binding.
  - Workspace creation applies memory and knowledge engine defaults automatically.
  - Management links are displayed and opened when available.

- **Bug Fixes**
  - Improved handling of workspace creation failures and invalid responses.
  - Prevented rebinding when the active account changes during workspace creation.
  - Clarified conflict messaging when no workspace is created.

[AI-9171]: https://altimateai.atlassian.net/browse/AI-9171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/opencode/src/cli/cmd/link.ts
```diff
@@ -484,19 +485,36 @@ async function createThenBindOrRebind(
...
+    // conflicting binding can never strand a half-created workspace.
+    //
+    // Already linked: that same atomicity makes it unusable. ``create_and_bind``
+    // pre-checks the identifiers and 409s *before* creating anything, so the
+    // rebind below never got a target and this row simply always failed — with
+    // an error telling the user to re-run the command they were already inside
+    // (AI-9171). Create unbound first, then repoint, which is what the row's
+    // own hint promises.
+    if (existing) {
+      const ws = await WorkspaceApi.createWorkspaceUnbound({ name })
```
sahrizvi, line 519: **Major**: This path is now three sequential requests (`createWorkspaceUnbound` → `rebindByMatchedIdentifier` → `recordApprovedBinding`), and `req()` reloads credentials independently on every call. If the active account changes mid-flow, each step could run against a different tenant — and since workspace IDs are tenant-local, a credential change between create and rebind could repoint this project to an unrelated workspace that happens to share the newly-created numeric ID in another tenant.

`runBrowserHandoff` elsewhere in this file already re-verifies credentials before binding for exactly this class of risk; this new flow doesn't have an equivalent guard.

Suggest pinning one credential snapshot across the whole create/rebind/cache sequence (or having `req()` accept an explicit snapshot and verify it hasn't changed before the rebind step).

packages/opencode/src/cli/cmd/link.ts
```diff
@@ -537,23 +554,31 @@ async function createThenBindOrRebind(
...
   // ``--directory`` argument so ``altimate-code link -d ./myproj`` and its
   // symlink-resolved twin both write under the same cache key (Kilo cycle 6).
+  // The unbound path has no server binding row to echo, so fall back to the
+  // identifier we just rebound on — the same values the rebind wrote.
   await recordApprovedBinding(identifier.projectPath ?? directory, {
     datamateId: created.datamate.id,
     datamateName: created.datamate.name,
-    repoRemote: created.binding.repo_remote,
-    projectPath: created.binding.project_path,
+    repoRemote: created.binding?.repo_remote ?? identifier.repoRemote ?? null,
```
sahrizvi: **Major**: The rebind's response is discarded above — `rebindByMatchedIdentifier(...)` returns the server's authoritative binding row (`Promise<BindingResponse>`), but nothing captures it. Because `created.binding` is `undefined` on this path, the cache here falls back to the local `identifier` object instead:

```ts
repoRemote: created.binding?.repo_remote ?? identifier.repoRemote ?? null,
projectPath: created.binding?.project_path ?? identifier.projectPath ?? null,
```

This is the only place in the file that caches a binding from a client-side guess rather than the server's response — every other path here uses `res.binding.*` directly. In the common case this coincidentally matches, but if a path-matched rebind ever leaves the remote null/stale server-side (or the server canonicalizes an identifier), the local cache will silently diverge from the actual binding row.

Suggest capturing the rebind response and using its `binding` fields directly, e.g. `created.binding = (await rebindByMatchedIdentifier({...})).binding`.

sahrizvi (commented):
## Review

Overall this is a well-reasoned fix for AI-9171 — splitting the atomic `createAndBind` (unlinked) from `createWorkspaceUnbound` + rebind (already-linked) matches the server's actual API contract, and explicitly sending `memory_enabled`/`knowledge_engine_enabled` closes a real silent-drift hazard between the two creation paths. Two Major issues are left as inline comments on `link.ts`. One more Major issue below can't be anchored to this diff since it lives in a file this PR doesn't touch.

### Major — The TUI plugin has the identical, still-unfixed bug
`packages/opencode/src/plugin/tui/altimate/workspace.tsx:496-546` (`createAndBindInline`)

This PR fixes the "＋ Create a quick workspace" row in the CLI (`link.ts`). The TUI picker has a structurally identical row wired to `createAndBindInline`, which unconditionally calls `WorkspaceApi.createAndBind` at line 509 regardless of whether `rebindFrom` (i.e. the project is already linked) is set. When the project is already linked, `createAndBind` 409s before creating anything — the catch block at line 511 shows a warning toast and returns, and the rebind branch at line 525 never executes. This is the exact bug this PR fixes, still present in the TUI's copy of the same flow. Worth applying the same unbound-create-then-rebind split to `createAndBindInline` before merging, or tracking as an immediate follow-up.

### Minor — Shared 409 message misattributes the cause on the unbound-create branch
`link.ts:507-524`

The `ConflictError` message ("Another workspace... claimed this project while you were choosing") is shared between both branches, but `createWorkspaceUnbound`'s request carries no `repo_remote`/`project_path` at all, so a genuine identity-conflict 409 isn't something that call can produce. If `/datamates/` ever 409s for an unrelated reason (e.g. a duplicate name), this message would misattribute it as a binding race. Worth branching the message by which path was actually taken.

### Minor — `created`'s inline type is a workaround for two different shapes
`link.ts:488`

`let created: { datamate: DatamateRef; binding?: Binding; manage_url?: string }` exists only because the two branches return different shapes. A discriminated union would make it type-safe to access `binding`/`manage_url` only where they're actually present, rather than relying on optional-chaining everywhere downstream.

### Minor — generic `Error` instead of a typed error
`api-client.ts:432-434`

Every other failure mode in this file throws a typed error (`ConflictError`, `PreconditionFailedError`, `WorkspaceApiError`), but the "server returned no usable id" guard throws a bare `Error`, making it harder for callers to distinguish this failure programmatically. Consider a typed error for consistency.

### Missing tests
- No test exercises `createThenBindOrRebind`'s control flow end-to-end — only the raw `createWorkspaceUnbound` request shape is tested.
- No test for the create-succeeds-but-rebind-fails orphan path.
- No test for the `manage_url` fallback via `manageUrlFor` on the unbound path (including the BYOK/`null` case).
- No test for the TUI's `createAndBindInline` already-linked path (would have caught the Major issue above).

saravmajestic:
Thanks — all addressed in `7b21606`, including a round of self-review to try to avoid a second pass.

## The Major one: the TUI had the same bug

You were right, and this was the important find. `createAndBindInline` called `createAndBind` unconditionally, so on an already-linked project the server 409'd before creating anything and the rebind at the bottom was unreachable — the identical defect, in the surface this PR wasn't touching. Fixing one and shipping the other would have left the CLI and TUI disagreeing about what the same row does.

Its `rebindFrom` doc also asserted the assumption the bug rested on ("createAndBind succeeds but leaves the binding pointing at the OLD workspace"), so that's corrected rather than left contradicting the code below it.

## Everything else

| Finding | Done |
|---|---|
| 409 message misattributes on the unbound branch | Split by which call ran. The unbound create sends no identifiers, so it cannot produce an identity conflict |
| `created`'s inline type | Discriminated union — `binding` / `manage_url` only reachable on the path that has them |
| Bare `Error` in `api-client.ts` | Now `WorkspaceApiError`, matching the rest of the module |
| **Pin the account across the two-step** (kilo/cubic) | Captured before the create, re-checked before the rebind, on both surfaces. The API key is deliberately excluded — rotating a key for the same user on the same tenant isn't an identity change and would abort a legitimate flow |
| **Cache the authoritative binding** (kilo/cubic) | The rebind response is kept and used. A path-keyed row rebound through `/by-path` was being cached with a `repo_remote` the server never stored |
| **Reject non-number ids** (coderabbit/cubic) | Real, and worse than it looks — see below |
| Test global state (kilo/coderabbit) | Documented why the env assignment must be module-scope (the module resolves `Global.Path` at import), sandbox keyed by pid+clock, original restored in `afterAll`, fetch restored per-test |

### On the id guard

Confirmed by running it: `Number.isSafeInteger(Number(x))` accepts `true` → 1, `"7"` → 7 and `[5]` → 5. The boolean case is the dangerous one — a malformed body would have rebound the project to **workspace 1** rather than failing. `typeof` now runs before the arithmetic.

### One I'm not taking — cubic's nested `{datamate:{id}}`

`POST /datamates/` declares `response_model=CreateDatamateResponse`, which is `{id: int}`, and FastAPI enforces it — the nested shape cannot come back from this endpoint. `AltimateApi.createDatamate`'s `?? data.datamate?.id` is defensive legacy for a different call. Adding an unreachable branch would just be somewhere for a real contract break to hide; the strict guard turns that into a loud failure instead. Happy to add it if you'd rather.

## Tests

New `create-then-rebind.test.ts` asserts the **sequence of endpoints**, for both surfaces — the bug was never in a payload, it was in which request got sent, and a payload-shaped test can't see that:

- unlinked → `POST /datamate-project-bindings/` only
- already-linked → `POST /datamates/` then `PUT …/by-remote`
- rebind fails → non-zero exit (CLI) / error toast (TUI), never a claimed success
- 409 on the unbound create → rebind must not run

Plus the coercion cases and the account fingerprint. Mutation-tested rather than assumed — reverting the CLI fix fails 3, reverting the TUI fix fails 2, dropping the feature flags fails 1, and reverting the id guard fails 3.

`520 pass / 0 fail` across `test/cli/cmd/link.test.ts` + `test/altimate/workspace/`. Typecheck clean; lint adds no new errors.

Both flow functions are now exported purely so those tests can reach them — flagged in case you'd rather I found another way.

## From my own re-read

Two things I caught and fixed before pushing: a duplicated JSDoc block left the original comment orphaned above `createThenBindOrRebind`, and I'd left internal scratch labels ("B3", "B4") in source comments that mean nothing to a reader — normalised to the `(review, PR #1314)` style the file already uses.

saravmajestic:
Superseded by #1318 — same change, clean branch.

`Tracker Leaks` was failing on this PR, and not for something fixable in place: it rejects internal tracker keys from the **branch name** as well as the diff and commit messages, and GitHub does not allow a PR's head branch to be renamed. Re-raising was the only way to get it green. My oversight — I did not check CI when I opened this.

@sahrizvi your review is fully addressed in #1318, including the Major one: the TUI's `createAndBindInline` had the identical unreachable-rebind bug, and now makes the same split as the CLI. The four bot findings are handled there too, with one declined and the reasoning given.

Sorry for the thread split, and for the three duplicate review comments I left earlier — those were a tooling mistake on my side.

PR #1319, opened 2026-09-18
fix(session): flatten tool history when a request declares no tools

### LINKED ISSUE #1315: UnknownError: "Could not get a response from the agent. Please try again later."
### Description

An agent working through the Altimate LLM Gateway for some extended time just stop working with an error I don't understand:

```bash
UnknownError: "Could not get a response from the agent. Please try again later."
```

I'm using Altimate code through VSCode and installed it a week or so ago, so on a pretty recent version of altimate.

### Plugins

n/a

### OpenCode version

n/a

### Steps to reproduce

No idea?

### Screenshot and/or share link

Image

### Operating System

Windows Server 10

### Terminal

Powershell

sahrizvi:
### Issue for this PR

Closes #1315

### Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation

### What does this PR do?

**The bug.** `compaction`, `title` and `summary` all run with `"*": "deny"`, so they send no
`tools` array — while summarizing a session's own history, which is full of `tool-call` and
`tool-result` parts. The Altimate gateway rejects that shape: its Responses-API namespace
conversion only runs when `tools` is present, so the history references functions the request
never declares, every provider in its fallback chain fails, and the client sees only:

```
UnknownError: "Could not get a response from the agent. Please try again later."
```

It reads like a gateway outage while ordinary turns keep working — because ordinary turns
declare their tools. Reported by several users at the point where a long session compacts.

**Scope of the defect.** Measured directly against each provider, with an otherwise identical
request:

| Provider | tool history, no `tools` declared |
|---|---|
| `openai/gpt-4o-mini` | accepts |
| `anthropic/claude-sonnet-4.5` | accepts |
| Altimate gateway | **rejects** |

So the root defect is **gateway-side**, and is tracked separately — the gateway should
synthesize declarations from the `tool_calls` already present in the history, which would also
fix the VS Code extension and every other client. This PR is the client-side mitigation.

**The fix.** `ProviderTransform.flattenToolParts` rewrites `tool-call` parts into
`[tool call: name(args)]` and `tool-result` parts into `[tool result: name]` plus the output,
merging results into the preceding assistant turn so role alternation is preserved. It is
applied at the single `streamText` call in `session/llm.ts`, and only when the request declares
no tools — every tool-declaring request keeps its structured parts untouched.

Flattening runs for all providers rather than branching on one: a request that declares no
tools has no use for structured tool parts either way, and a single code path avoids a
provider-specific branch that would rot as the gateway changes. Happy to narrow it to
`altimate-backend` if reviewers prefer.

### How was this tested?

- 6 unit tests in `test/provider/flatten-tool-parts.test.ts`, covering removal, content
  preservation, an assistant turn holding only tool calls, an orphaned tool result, and a
  no-op on tool-free history.
- Mutation-checked: replacing the function body with a passthrough fails 2 of the 6.
- **End to end against the live gateway** — the real function output was posted to
  `/agents/v1/chat/completions`:

```
BEFORE fix (tool messages, no tools declared): FAIL -> generic gateway error
AFTER  fix (actual flattenToolParts output)  : PASS (63 chunks)
```

- `bun run typecheck` clean (13/13); full `bun test` 13,659 pass / 0 fail.

### Verification

Drive any session to auto-compaction against the Altimate LLM Gateway — it previously failed at
the Compaction step and now produces a summary. `title` and `summary` share the same chokepoint.

https://claude.ai/code/session_01V4pzHPsqfbTdLMKzMaSYcF

---
## Summary by cubic
Fixes compaction and title requests failing with a generic `UnknownError` on the Altimate gateway: those agents declare no tools but were sending tool-call history, which the gateway rejects (OpenAI and Anthropic both accept that shape). Requests that declare no tools now flatten their tool history to text before sending.

**Bug Fixes**
- `tool-call` parts become `[tool call: name(args)]`; `tool-result` parts become `[tool result: name]` plus output text, including results embedded in assistant messages.
- Results merge into the preceding assistant turn so role alternation is preserved; the blank-text filter only drops parts the transform itself produced.
- Title generation now sets `toolChoice: "none"`, preventing tool-stub injection from repopulating `tools` and bypassing the flattening.
- Tool-declaring requests keep structured parts untouched; flattening runs for all providers to avoid a provider-specific branch.
- Covered by 13 unit tests and verified against the live gateway (fails before, passes after).
- The root defect is gateway-side and tracked separately; this PR is the client-side mitigation.

<sup>Written for commit f25caf4418c7e6e471182850c3e74a92c313ca29. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/AltimateAI/altimate-code/pull/1319?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility with AI providers that do not support tools by converting historical tool interactions into readable conversation text when needed.
  - Preserved tool names, arguments, results, media indicators, and conversation order while avoiding unintended empty assistant messages.
  - Requests with tool support continue to retain their original tool interactions.
  - Improved handling of embedded and structured tool results, including readable text and media indicators.
  - Title generation remains tool-free, preventing historical tool interactions from being reintroduced.


saravmajestic (dismissed):
## Approve — sound fix, well-targeted. Two minors and a nit, none blocking.

Reviewed against the code rather than the description, in a clean worktree at `3a94ff0`.

### Load-bearing claims I verified independently

- **The `"invalid"` filter is correct.** It is the always-present AI-SDK fallback tool (`llm.ts:238-241` never deletes it), and the filter matches the existing `activeTools` filter at line 311. Consistent with how the rest of the file counts real tools.
- **The ordering works.** `addHistoricalToolStubs` (line 220) mutates `tools` *before* `declaresNoTools` is computed (line 271). So a request whose stubs repopulated `tools` correctly skips flattening — and it also now declares tools, so the gateway defect does not apply to it. The two mechanisms compose rather than collide.
- **`toolChoice: "none"` on title generation is load-bearing, not cosmetic.** `addHistoricalToolStubs` skips only when `toolChoice === "none"` **and** `tools` is empty, so without the flag the stub injection would repopulate `tools` and silently suppress the flattening this PR adds. The description is accurate on this point.
- **No mutation of caller input.** Verified with a probe that snapshots the input and compares after the call.
- **The other four `tools: {}` call sites** this PR does not touch — `altimate/skill-selector.ts:181`, `altimate/enhance-prompt.ts:122`, `altimate/review/ai-review.ts:116`, `server/routes/instance/httpapi/handlers/project-copy.ts:47` — do not pass `toolChoice: "none"`, but they are safe either way: either stub injection populates `tools` (so tools are declared and the gateway is satisfied), or the names are dropped as malformed and the flattening applies. No gap, but worth knowing they exist if the gateway fix lands later.
- **The 12 failures in `test/session/prompt.test.ts` are pre-existing.** Identical count and identical 5000ms-timeout signature on `origin/main`. Not caused by this PR.

### MINOR — coalescing breaks for a string-content assistant message

`appendAssistant` merges only when `Array.isArray(previous.content)`. A `ModelMessage` assistant may legitimately carry a plain string, and then the tool result becomes a separate assistant message:

```
assistant(content: "text") -> tool   ==>   assistant -> assistant
```

which is the exact shape the docblock says it prevents — and this PR exists because the gateway is stricter about message shape than upstream.

**Not currently reachable** on the flattening paths: the toolless agents get their history through `MessageV2.toModelMessages`, which always produces array content. But the function is exported and written generally, and this codebase does construct string-content assistant messages (`session/prompt.ts:1535`, `content: MAX_STEPS`).

One caveat if you fix it: the `Array.isArray` guard is *accidentally* what prevents mutating caller-owned messages, since those reach `result` via `result.push(msg)`. A naive fix reintroduces that. Replacing the entry avoids it:

```ts
result[result.length - 1] = { ...previous, content: [...prior, ...parts] }
```

### MINOR — the token estimate and the request now see different messages

`llm.ts:262` estimates from `input.messages` (unflattened) while the request sends the flattened messages. Flattened text is smaller than the structured parts, so the estimate over-counts — conservative, and therefore safe today. But they are now two different inputs and can drift, which is the sort of thing that is much cheaper to note now than to debug later.

### NIT — `toolCallId` is dropped by `renderToolResult`

Correct for the only consumer (summarization), but worth a line in the docblock so it reads as a decision rather than an oversight.

### Things done well

- The root-cause analysis is specific and falsifiable — the per-provider table showing OpenAI and Anthropic accept the shape while the gateway rejects it is what makes "the real defect is gateway-side, this is the client-side mitigation" credible rather than an assumption.
- Flattening for all providers instead of branching on `altimate-backend` is the right call, and the stated reason is the correct one: a request that declares no tools has no use for structured tool parts either way. A provider-specific branch would rot silently as the gateway changes.
- The blank-text filter only dropping parts the transform itself produced is a careful detail — it is what keeps the intentional empty separator between Anthropic signed-reasoning blocks alive.
- 13 unit tests plus an end-to-end check against the live gateway (failing before, passing after) is the right evidence for a fix whose trigger is a specific server's strictness.

saravmajestic (approved):

